PR #6, opened 2017-10-30
Couple of copy contructor and/or assignment operator missing.

Carmot:
The rule of three:
https://en.wikipedia.org/wiki/Rule_of_three_%28C%2B%2B_programming%29


qinzuoyan:
Thanks for you propose. Because the RocksDB project is forked from [facebook/rocksdb](https://github.com/facebook/rocksdb), so I think you'd better send pull request to facebook/rocksdb firstly.

PR #12, opened 2018-03-09
remove rocksdb filefolder, and reference it as a submodule

acelyc111:
- remove rocksdb filefolder, and reference it as a submodule
- add some test for upgrade and downgrade


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "rdsn"]
-	path = rdsn
-	url = https://github.com/xiaomi/rdsn.git
+  path = rdsn
+  url = https://github.com/XiaoMi/rdsn.git
+[submodule "rocksdb"]
+	path = rocksdb
```
qinzuoyan: 统一用空格吧

PR #51, opened 2018-05-22
shell: add get_app_envs; ls apps do not print envs

qinzuoyan:


src/shell/main.cpp
```diff
@@ -334,12 +334,21 @@ command_executor commands[] = {
         "<restore_app_id> [-d|--detailed]",
         query_restore_status,
     },
+    {
+        "get_app_envs", "set current app envs", "", get_app_envs,
```
acelyc111: "get current app envs"

qinzuoyan: done

PR #163, opened 2018-08-08
build: fix rocksdb build

shengofsun:
1. should add "-CMAKE_BUILD_TYPE=debug" to "CMAKE_OPTIONS" too.
2. when there is an existing makefile, we shoudn't run cmake again. otherwise the build process may fail.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -206,6 +206,13 @@ function run_build()
     else
         echo "ENABLE_GCOV=NO"
     fi
+    if [ "$BUILD_TYPE" == "debug" ]
+    then
+        echo "BUILD_TYPE=debug"
+        CMAKE_OPTIONS="$CMAKE_OPTIONS -DCMAKE_BUILD_TYPE=Debug"
+    else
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -217,37 +224,34 @@ function run_build()
         fi
     fi
 
-    if [ "$CLEAR" == "YES" -a -d "$ROCKSDB_BUILD_DIR" ]
+    if ( [ "$CLEAR" == "YES" ] || [ "$PART_CLEAR" == "YES" ] ) && [ -d "$ROCKSDB_BUILD_DIR" ]
```
acelyc111: part clear的时候不需要clear rocksdb吧？
“-cc|--half-clear      clear pegasus before building, not clear thirdparty/rdsn/rocksdb”

shengofsun: 这个没仔细看，可能是之前我记错了。

shengofsun: fixed

PR #167, opened 2018-08-20
 scripts: improve pegasus_rolling_update.sh; fix shell

qinzuoyan:


src/shell/commands.h
```diff
@@ -3654,10 +3654,9 @@ inline bool app_stat(command_executor *e, shell_context *sc, arguments args)
         << std::setw(w) << std::right << "INCR" << std::setw(w) << std::right << "CAS"
         << std::setw(w) << std::right << "CAM" << std::setw(w) << std::right << "SCAN";
     if (!only_qps) {
-        out << std::setw(w) << std::right << "storage_mb" << std::setw(w) << std::right
```
neverchanje: 为啥不打印 storage_mb ？

qinzuoyan: 因为storage_mb改成了file_mb，已经打印了

PR #210, opened 2018-11-05
scripts: fix get_xx_lib funcs in pack_common.sh

vagetablechicken:
add 32/64-bit check when getting stdcpp or system lib
only allows one lib to be found


qinzuoyan:
更新下rdsn

neverchanje:
screenshot from 2018-11-07 11 02 44
你不应该把 file mode 改了

PR #214, opened 2018-11-13
docker: support ubuntu 16.04

neverchanje:


wangdehu:
Should you execute the command apt-get update before install ?

neverchanje:
> Should you execute the command apt-get update before install ?

thanks for your review, this is a mistake.

PR #242, opened 2018-12-24
refactor as uri resolver is removed

shengofsun:


src/server/info_collector.cpp
```diff
@@ -34,7 +35,8 @@ info_collector::info_collector()
         _meta_servers.group_address()->add(ms);
     }
 
-    _cluster_name = dsn_config_get_value_string("pegasus.collector", "cluster", "", "cluster name");
+    _cluster_name = dsn_config_get_value_string(
+        "pegasus.collector", PEGASUS_CLUSTER_SECTION_NAME.c_str(), "", "cluster name");
```
qinzuoyan: 这里要用"cluster"

PR #271, opened 2019-01-30
shell: copy_data command supports filter

qinzuoyan:
improve shell's copy_data command:
* support HashKey/SortKey/Value filters
* support only copy one partition by `--partition` option
* support not overwrite data (only set if not exist) by `--no_overwrite` option


src/shell/command_helper.h
```diff
@@ -101,6 +102,9 @@ struct scan_data_context
     int split_id;
     int max_batch_count;
     int timeout_ms;
+    bool no_overwrite; // for copy_data
```
neverchanje: 应该加一行注释 "uses check_and_set instead of using set for insertion"，单看这个选项不容易看明白

qinzuoyan: improved

PR #275, opened 2019-02-15
scripts: fix pack tools to copy lib with the right name

qinzuoyan:
to fix the `lib file name not match` problem like:
- ldd pegasus_server: libaio.so.1 => /lib64/libaio.so.1
- pack_server: DSN_ROOT/lib/libaio.so.1.0.0


neverchanje:
没看懂这个是为啥，用 libaio.so.1.0.0 这个名字有什么问题吗

PR #300, opened 2019-03-06
*: add whitelist config & optimize pegasus_offline_node.sh

vagetablechicken:
### Whitelist
Ref #238 

When enable_white_list is true, replicas which are not in the white list will be rejected, and print logs like
```
W2019-03-06 15:18:11.622 (1551856691622858222 25b7f) replica.     fd1.0609000000000003: failure_detector.cpp:380:end_ping_internal(): worker rejected, stop sending beacon message, remote_master[x.x.x.x:xx], local_worker[x.x.x.x:xx]
```

So, if you want scale-out, you should update `replica_white_list` in the config of meta-server, and restart the meta-server.
```
[meta_server]   
  enable_white_list = true  
  replica_white_list = %{replica.server.list}
```

After that, restart the replica-server, then it will not be rejected by meta-server.

### Live Percentage
Control node_live_percentage_threshold_for_update when offline nodes to prevent getting meta-server  freezed.

waiting for rdsn pr


.gitignore
```diff
@@ -88,5 +88,6 @@ t
 LOG
 
 .vscode/
+.history/
```
neverchanje: 这是啥

vagetablechicken:
缩容脚本还要再测试一下，先不忙过

qinzuoyan:
offline脚本测试过了吗？

vagetablechicken:
@qinzuoyan 测试过了
offline_node_list.sh的逻辑为，live percentage设为0之后再开始批量下线节点。不需要再设回DEFAULT，原因有两个：
1. 完成下线后meta没有重启，这时候设置为DEFAULT，还是会触发freezed
2. 下线操作完成后需要重启meta-server，meta-server又会重新从配置文件中读取live_percentage参数

.gitignore
vagetablechicken: vscode的文件历史

PR #345, opened 2019-06-27
scripts: check balance state by total balance count on meta & add checking methods

neverchanje:
 
1. We used to check the imbalanced state by identifying if the primaries on each node are equal.
That's not accurate since in some cases the meta server may still see the cluster a balanced state while
the nodes are not equally serving replicas.
Now we use "balance_operation_count"  from "cluster_info" command to check the imbalance.

2. Add more cluster checking approaches, like "server version inconsistency", and "unsteady meta level".

### Check List 

Tests 

- Manual test

### DIFF
--- a/scripts/clusters_show.sh
+++ b/scripts/clusters_show.sh
@@ -50,7 +50,7 @@ do
   tmp_file="/tmp/$UID.$PID.pegasus.clusters_status.cluster_info"
   echo "cluster_info" | ./run.sh shell -n $cluster &>$tmp_file
   cluster_info_ok=`grep "succeed" $tmp_file | wc -l`
-  if [ $cluster_info_ok -ne 1 ]; then
+  if [ $cluster_info_ok -ne 0 ]; then
     echo "ERROR: get cluster info failed, refer error to $tmp_file"
     exit 1
   fi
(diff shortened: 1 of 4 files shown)


scripts/clusters_show.sh
```diff
@@ -50,7 +50,7 @@ do
   tmp_file="/tmp/$UID.$PID.pegasus.clusters_status.cluster_info"
   echo "cluster_info" | ./run.sh shell -n $cluster &>$tmp_file
   cluster_info_ok=`grep "succeed" $tmp_file | wc -l`
-  if [ $cluster_info_ok -ne 1 ]; then
+  if [ $cluster_info_ok -ne 0 ]; then
```
qinzuoyan: 为什么是 -ne 0 ？

neverchanje: 我这边正常返回是 0，不正常返回是 1

qinzuoyan: 这里的根本原因是以前会在最后打印一行“get cluster info succeed”，后来这行调整没了，所以这里的检查就不能这么写了，得采用别的方式。

qinzuoyan:
@neverchanje 解决冲突，解决上面comments的问题。

Issue #293, opened 2019-02-27
refactor shell to reduce duplicated code


qinzuoyan:
like:
* option parsing of copy_data/clear_data/count_data

PR #384, opened 2019-08-26
fix: unit test failed and update rdsn

hycdong:
- fix some unit tests
- update rdsn and update some cmake files

- fix unit test `multi_remove` failed
    - [rdsn #293](https://github.com/XiaoMi/rdsn/pull/293) update fail point usage, but not apply new usage to `multi_remove`, this pr fix it
- fix function test `restore` and `recovery` failed
    - pegasus #370 remove `config-server.ini` and create `config.min.ini`, however, `restore` and `recovery` function test directly use `config-server.ini` file name and dynamically modify config in those tests,  delete the file will lead to test failed
    - Moreover, `config.min.ini` format file adding two `space` characters on the beginning of the line, which leads to modifying config failed, so add `\s*` to identify `space` characters in regex for `sed` command
    - Besides, `config.min.ini` remove some configs used in those tests, add them to make tests pass
        - add thread pool `THREAD_POOL_LOCAL_SERVICE` for restore test
        - update worker count to 4 for 'threadpool..default', because recovery test require thread pool THREAD_POOL_META_SERVER has more than one worker
        - add recover_from_replica_server config for recovery test
- update rdsn
    - [rdsn #291](https://github.com/XiaoMi/rdsn/pull/291), add boost::regex in some cmake files

### Check List 

Tests 

- Unit test

### Tips
- current travis script could not notify tests error well, @vagetablechicken will fix it in further pull request


hycdong:
TODO:
- travis run time will exceed 50 minute currently [travis-build-timeout](https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts)
- build pegasus use 35mins, try to shorten the time

PR #489, opened 2020-03-03
fix: fix bug in del function of redis_proxy

levy5307:
In redis_proxy，"OK" will be returned using del function to delete a key, If the function executes successfully. But interface definition of redis shows that it should return the number of keys which are deleted.
While, there are some difference between interface definition of redis and redis proxy of pegasus. Because whether the key is exist or not, pegasus will return "OK" to redis proxy, so we cann't distinguish whether the key is deleted or not. so in this case , we unity returns 1.

Redis documentation about del: https://redis.io/commands/del

> **DEL key [key ...]**
> Removes the specified keys. A key is ignored if it does not exist.
> **Return value**
> Integer reply: The number of keys that were removed.
> Examples
> 
```
redis> SET key1 "Hello"
"OK"
redis> SET key2 "World"
"OK"
redis> DEL key1 key2 key3
(integer) 2
redis> 
```

Change the return value from "OK" to the number of keys.

### Check List 

Tests 

1.start a redis proxy in c4tst-injection
2.clone redis code, and build it.
3.use command of ./redis_cli -h host -p port to connect with the redis proxy
4.insert a k-v pair into redis proxy:
```
redis> SET key "value"
"OK"
```
5. in pervious versionm, regardless of the key is exist or not, del the key, it returns "OK"
```
redis> DEL key "value"
"OK"
```
6.del the key which was inserted before in current version. It returns 1, not "OK".
```
redis> DEL key "value"
(integer) 1
```
7.del a key which is not exist in current version. It returns 1, also.  
```
redis> DEL not_exist_key "value"
(integer) 1
```
Related changes


neverchanje:
Can you include the manual test in the PR description? 

Like this:

Previous version
```
> redis del a
> ok
```

Current version
```
> redis del a a
> -1
> redis del a
> 1
```

acelyc111:
> Can you include the manual test in the PR description? This PR does not show any kind of test for us.
> 
> Like this:
> 
> Previous version
> 
> ```
> > redis del a
> > ok
> ```
> 
> Current version
> 
> ```
> > redis del a a
> > -1
> > redis del a
> > 1
> ```

neverchanje:
I tried these commands on http://try.redis.io/. Deleting a non-existed key will return 0, but your version returns 1. This should be noted in our code.

```
> set a a
OK
> del a
(integer) 1
> del a
(integer) 0
```

src/redis_protocol/proxy_lib/redis_parser.cpp
```diff
@@ -681,7 +681,7 @@ void redis_parser::del_internal(message_entry &entry)
                     simple_error_reply(entry,
                                        "internal error " + std::to_string(rrdb_response.error));
                 } else {
-                    simple_integer_reply(entry, -1);
+                    simple_integer_reply(entry, 1);
```
neverchanje:
```suggestion
                    // NOTE: Deleting a non-existed key returns 1 too. But standard Redis returns 0. Pegasus behaves 
                    // differently in this case intentionally for performance.
                    // Because if we need to check the existence, we should use check_and_mutate instead.
                    simple_integer_reply(entry, 1);
```

Issue #486, opened 2020-02-27
Limit long-time running Scan/MultiGet operators


qinzuoyan:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Sometimes Scan/MultiGet operators may run for long time because of too many expired or filtered data, which may cause CPU/IO busy and make server unavailable. We should limit these long-time running Scan/MultiGet operators, for example:
* limit max rows or bytes to scan (including expired and filtered data) in one request
* limit max running time to process in one request

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

hycdong:
Pegasus server:
- muliget
  1. add max_iterator_count check, including expire_count and filter_count
  2. add max_iterator_size check, client may set max_kv_size as -1 means no-limit, server should set a max threshold
  3. add max_execute_time check to avoid long time scan
- scan
   1. add max_iterator_count_for_one_scanner check, including expire_count and filter_count
   2. add max_execute_time check
- sortkey_count
   1. add max_execute_time check
Options above are cluster level, and can also set by app_envs for each table. (consider it)
If iterator exceed count or size or time limit, server should return rocksdb Incomplete error (just like multiget)

Pegasus java client:
- multiGet (no need to update)
- sortKeyCount if got Incomplete error, return -1, not throw exception
- scan: not throw exception if got Incomplete error, and stop asyncNext if failed several times
- delRange: stop if scan got Incomplete error

Pegasus C++ client:
- interface update same as java client
- copy_data
- clear_data
- count_data

PR #500, opened 2020-03-20
feat: limit long time rocksdb iteration operation

hycdong:
As issue #486 shows, this pull request aims to limit long time rocksdb iteration.

In multiget, scan operator, pegasus might scan rocksdb by `rocksdb::Iterator`, the iteration count for is defined by the client option which is totally controlled by user.

**For multiget**, client can define `max_kv_count` and `max_kv_size` in option, if count or size exceed limit, server will stop iteration and return kv-pairs have already been scanned and error kIncomplete to client.  If client set `max_kv_count` or `max_kv_size` as -1, it means that server won't limit count or size, which will be dangerous. Besides, `max_kv_count` doesn't include expire data count. As result, we update iteration check for multiget below:
1. update `max_kv_count` to `max_iteration_count`, count expire_count and filter_count.
2. `max_iteration_count` = min(`max_kv_count`[from client], `rocksdb_iteration_count`[from config])
3. `max_iteration_data_size` = min(`max_kv_size`[from client], `multi_get_iteration_size`[from config])
4. add time threshold for iteration, if time exceed limit, server will also stop  iteration and return kv-pairs have already been scanned and error kIncomplete to client. If time threshold = 0 means no time limit. Besides, this option is defined in config file, but can also re-config from app_env.

**For scan**, client can define `batch_size` in option, server will stop iteration if count exceed and return kv-pairs to client, and this count is also not included expire_count. We add iteration check for scan below:
1. update `batch_size` to `max_iteration_count`, count expire_count and filter_count.
2. `max_iteration_count` = min(`batch_size`[from client], `rocksdb_iteration_count`[from config])
3. add time threshold for iteration, if time exceed limit, server will also stop  iteration and return kv-pairs have already been scanned and error kIncomplete to client. Time threshold meaning is as same as mulitget operator.

**For sortkey_count**, we only add time threshold, and if time exceed limit, server will return -1 to client.

### Config update
```
# cluster level restriction {3000, 30MB, 1000, 30s}
rocksdb_multi_get_max_iteration_count = 3000
rocksdb_multi_get_max_iteration_size = 31457280
rocksdb_max_iteration_count = 1000
rocksdb_iteration_threshold_time_ms = 30000
```

### Check List 

Tests 
- Unit test
- Integration test
- Manual test


src/server/pegasus_server_impl.cpp
```diff
@@ -755,11 +784,21 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
 
         std::unique_ptr<rocksdb::Iterator> it;
         bool complete = false;
+        uint64_t iteration_time = dsn_now_ns();
```
neverchanje: iteration_start_time. 

iteration_time confuses with the time duration rather than the timestamp.

src/server/pegasus_server_impl.cpp
```diff
@@ -675,12 +697,19 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
...
 
-    int32_t max_kv_count = request.max_kv_count > 0 ? request.max_kv_count : INT_MAX;
+    int32_t max_iteration_count =
+        (request.max_kv_count > 0 && request.max_kv_count < _rocksdb_max_iteration_count)
+            ? request.max_kv_count
+            : _rocksdb_max_iteration_count;
     int32_t max_kv_size = request.max_kv_size > 0 ? request.max_kv_size : INT_MAX;
+    int32_t max_iteration_size_config =
+        _multi_get_max_iteration_size > 0 ? _multi_get_max_iteration_size : INT_MAX;
+    int32_t max_iteration_size = std::min(max_kv_size, max_iteration_size_config);
```
acelyc111: Different style with L700~L703, I'd prefer the style here.

src/server/pegasus_server_impl.cpp
```diff
@@ -755,11 +784,21 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
...
+        uint64_t iteration_time = dsn_now_ns();
+        bool exceed_limit = false;
         if (!request.reverse) {
             it.reset(_db->NewIterator(_data_cf_rd_opts));
             it->Seek(start);
             bool first_exclusive = !start_inclusive;
-            while (count < max_kv_count && size < max_kv_size && it->Valid()) {
+            while (iteration_count < max_iteration_count && size < max_iteration_size &&
+                   it->Valid()) {
+                iteration_time = dsn_now_ns();
```
acelyc111: Call `dsn_now_ns` for each iteration may be to costly?

src/server/pegasus_server_impl.cpp
```diff
@@ -90,6 +90,28 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+        "rocksdb_max_iteration_count",
+        1000,
+        "max iteration count for each rocksdb iterator operation, if exceed this threshold,"
+        "iterator will be stopped");
+
+    _rocksdb_iteration_threshold_ns_in_config =
+        dsn_config_get_value_uint64("pegasus.server",
+                                    "rocksdb_iteration_threshold_ns",
+                                    30000000000,
+                                    "max duration for rocksdb iterator operation if exceed "
```
neverchanje: > "max duration for rocksdb iterator operation"

What does "iterator operation" mean? As far as I understand, this threshold is for one scan, not for one iterator operation.

```suggestion
                                    "max duration for handling one pegasus scan request (sortkey_count/multiget/scan)"
```

neverchanje:
I suggest on the server side we can set a validation before handling the request.

```
void pegasus_server_impl::on_multiget(...) {
  if(multiget_validate(req.max_kv_count, req.max_kv_size)) {
    return ERR_INVALID_ARGUMENT;
  }
}
```

hycdong:
> > For sortkey_count, we only add time threshold, and if time exceed limit, server will return -1 to client.
> 
> `sortkey_count` doesn't consume less resource than multiget or scan during iterations. Why it has less restriction?

`sortkey_count` will get sortkey count for a specific hashkey, its iteration_count will be equal to sortkey. If we restrict iteration count = 100, and the sortkey_count is 1000, client will never get the correct count. As a result, we only restrict iteration time.

hycdong:
> I suggest on the server side we can set a validation before handling the request.
> 
> ```
> void pegasus_server_impl::on_multiget(...) {
>   if(multiget_validate(req.max_kv_count, req.max_kv_size)) {
>     return ERR_INVALID_ARGUMENT;
>   }
> }
> ```

In client side, we don't restrict max_kv_count and max_kv_size, client even allow it to be set as -1, which means no limit. As a result, we add restriction in server side. Besides, max_kv_count default value is 100, max_kv_size default value is 100000, most user will not modify them.

src/server/pegasus_server_impl.cpp
hycdong: I consider your comment and I agree that it is too costly. Most normal iteration won't cost such long time, we needn't check every iteration, so I update it, check at most 10 times during iteration. For example, if max_iteration_count is 100, I only check if time exceed when iteration_count % 10 == 0.

src/server/pegasus_server_impl.cpp
hycdong: iteration_time is duration time, not iteration_start_time, it will be updated when checking time exceed limit.

src/server/pegasus_server_impl.cpp
```diff
@@ -675,12 +697,19 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
         return;
     }
 
-    int32_t max_kv_count = request.max_kv_count > 0 ? request.max_kv_count : INT_MAX;
+    int32_t max_iteration_count =
```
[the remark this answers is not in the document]
hycdong: I update it like the code below, like L703~706:
```
uint32_t max_kv_count = request.max_kv_count > 0 ? request.max_kv_count : INT_MAX;
uint32_t max_iteration_count = std::min(max_kv_count, _rocksdb_max_iteration_count);
```

src/server/pegasus_server_impl.cpp
hycdong: I update L700~L701 code style:
```
uint32_t max_kv_count = request.max_kv_count > 0 ? request.max_kv_count : INT_MAX;
uint32_t max_iteration_count = std::min(max_kv_count, _rocksdb_max_iteration_count);
```

src/server/pegasus_server_impl.cpp
```diff
@@ -675,12 +697,22 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
         return;
     }
 
-    int32_t max_kv_count = request.max_kv_count > 0 ? request.max_kv_count : INT_MAX;
+    uint32_t max_kv_count = request.max_kv_count > 0 ? request.max_kv_count : INT_MAX;
+    uint32_t max_iteration_count = std::min(max_kv_count, _rocksdb_max_iteration_count);
```
acelyc111: After the PR, we should update client doc, the returned kv count/size maybe less than user requested.

src/server/pegasus_server_impl.cpp
```diff
@@ -822,7 +867,7 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
             it->SeekForPrev(stop);
             bool first_exclusive = !stop_inclusive;
             std::vector<::dsn::apps::key_value> reverse_kvs;
-            while (count < max_kv_count && size < max_kv_size && it->Valid()) {
+            while (iteration_count < max_iteration_count && size < max_kv_size && it->Valid()) {
```
acelyc111: Should check `size < max_iteration_size` too ?

src/server/pegasus_server_impl.cpp
```diff
@@ -778,7 +816,14 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
                     }
                 }
 
-                iterate_count++;
+                iteration_count++;
+                if (time_check && iteration_count % module_num == 0) {
+                    iteration_time = dsn_now_ns();
```
levy5307: Maybe you can move `dsn_now_ns()` out of `while`

src/server/config.ini
```diff
@@ -280,6 +280,11 @@
   # Bloom filter type, should be either 'common' or 'prefix'
   rocksdb_filter_type = prefix
 
+  # 10MB, 1000, 30s
+  rocksdb_multi_get_max_iteration_size = 10000000
```
levy5307: 10 << 20

src/server/pegasus_server_impl.cpp
hycdong: Sure, I will update related document~

src/server/pegasus_server_impl.cpp
hycdong: Sure~ I miss to update condition here before

src/server/config.ini
hycdong: I update to `rocksdb_multi_get_max_iteration_size = 10485760` here. It seems that `10 << 20` is an invalid value while parsing options from config.

levy5307: What is invalid mean? parse failure?

hycdong: We read the value as a string_view, and use a function called `buf2unsigned` to convert it into a uint_64 value. `10 << 20` is not a valid string which can be parsed into `uint64_t`.

src/server/pegasus_server_impl.cpp
```diff
@@ -793,7 +831,14 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
                     }
                 }
 
-                iterate_count++;
+                iteration_count++;
```
neverchanje: Can we separate all these stuff about scan limitation into a new class? I think we don't to keep all these in pegasus_server_impl. Also, you can reuse the limitation logic in scan/multiget/sortkey_count, even they may be different in some behaviors. Mostly the codes are repeated. 

```cpp
class scan_limiter {
public:
  scan_limiter(int time, int size, int count) {}
  void start() { dsn_now_ns(); }
  bool valid() const {
    if(_max_time != 0 && duration() > _max_time) { return false; } 
    if(_max_count != 0 && _iterated_count > _max_count) { return false; } 
    if(_max_size != 0 && _iterated_size() > _max_size) { return false; }
    return true;
  }
  int duration() const { /*optimization on getting time*/ } 
  void add() { _iterated_size++; _iterated_count++; }

private:
  bool _exeed_limit{false};
  int _start_time, _end_time;
  int _iterated_count;
  int _iterated_size;

  const int _max_time{0}; // zero means no limit
  const int _max_size{0};
  const int _max_count{0};
};
```

src/server/pegasus_server_impl.cpp
```diff
@@ -856,7 +900,10 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
                     }
                 }
 
-                iterate_count++;
+                limiter->add_count();
+                if (!limiter->time_check()) {
```
neverchanje, line 904: You can include `time_check()` in `valid()`.

I guess why you separate the two calls is that `time_check` depends on `add_count`. Because in the first iteration, the condition always meets `_iteration_count % _module_num == 0`, so it checks time duration futilely.

If you want every "module_num" times, the `dsn_now` actually calls, it needs a slight change:

```cpp
    // during rocksdb iteration, if iteration_count % module_num == 0, we will check if iteration
    // exceed time threshold, which means we at most check ten times during iteration
    bool time_check()
    {
        if (_max_duration_time > 0 &&  (_iteration_count+1) % _module_num == 0 &&
            dsn_now_ns() - _iteration_start_time_ns > _max_duration_time) {
            _exceed_limit = true;
            _iteration_duration_time_ns = dsn_now_ns() - _iteration_start_time_ns;
            return false;
        }
        return true;
    }
```

So that only after (module_num-1) will there be the first call of dsn_now.

```cpp
    bool valid()
    {
        if (_iteration_count >= _max_count) {
            return false;
        }
        if (_max_size > 0 && _iteration_size >= _max_size) {
            return false;
        }
        return time_check();
    }
```

src/server/pegasus_server_impl.cpp
```diff
@@ -971,7 +1027,7 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
             // extract value
             if (status.ok()) {
                 // check if exceed limit
-                if (count >= max_kv_count || size >= max_kv_size) {
+                if (iteration_count > max_iteration_count || size > max_iteration_size) {
```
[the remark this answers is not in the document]
hycdong: I'm sorry, but I just don't know what is your suggestion. Firstly, abnormal log will not always print `start_sort_key="", stop_sort_key=""`, client is allowed to set start and stop key. Secondly, we don't distinguish range read and point-read in old implementation, this abnormal log is not printed if exceeding limit, but multiget slow query log print.

src/server/pegasus_server_impl.cpp
hycdong: Your conjecture is correct and just is part of the reason that I used to separate adding iteration count. 
Iteration_count will not always be auto-increment during iteration.

src/server/pegasus_server_impl.cpp
```diff
@@ -869,7 +916,7 @@ void pegasus_server_impl::on_multi_get(const ::dsn::apps::multi_get_request &req
                 if (r == 1) {
                     count++;
                     auto &kv = reverse_kvs.back();
-                    size += kv.key.length() + kv.value.length();
+                    limiter->add_size(kv.key.length() + kv.value.length());
```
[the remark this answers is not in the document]
hycdong: In current implementation, we calculate size by `blob` structure, not `Slice` structure provided by `iterator`. As a result, if we would like to add size, we can not use `it->key()` and `it->value`. You can see function `append_key_value_for_multi_get`, in this function, if  value is expired, return directly, will not parse sortkey and value from `Slice` to `blob`. I think it is not necessary to parse expired value for calculate exact iteration_size.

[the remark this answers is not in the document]
neverchanje: >  if value is expired, return directly, will not parse sortkey and value from Slice to blob. 

But reading such a row still costs at least disk resources. What if a scan reads only 100 rows but with 100000 expired rows?

src/server/pegasus_server_impl.cpp
hycdong: I misunderstand your suggestion before, the code is updated.

src/server/pegasus_server_impl.cpp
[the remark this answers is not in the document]
hycdong: We discuss it offline. I update code, don't update `iteration_count` if sortkey array is not empty.

src/server/pegasus_server_impl.cpp
hycdong: I misunderstand your suggestion before, the code is updated.

src/server/pegasus_server_impl.cpp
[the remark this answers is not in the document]
hycdong: We discuss it offline, expired will also add iteration_count.

PR #515, opened 2020-04-13
fix: fix the bug in restore_test

levy5307:
The restore_test will fail when we run function test now. This pr is used for fix this bug. 

In our implement, backup interval must be greatger than `cold_backup_checkpoint_reserve_minutes`, which is 10 minutes in default.
And in restore_test, the backup interval is 1 seconds, while cold_backup_checkpoint_reserve_minutes is 10 minutes. So it will produces some error in running restore_test.

### Check List 

Related changes

yes

### DIFF
--- a/src/server/config.min.ini
+++ b/src/server/config.min.ini
@@ -101,7 +101,7 @@
   cold_backup_root = onebox
   duplication_disabled = true
   cluster_name = onebox
-  cold_backup_checkpoint_reserve_minutes = 0
+  cold_backup_checkpoint_reserve_minutes = 10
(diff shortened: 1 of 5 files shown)


hycdong:
If `cold_backup_checkpoint_reserve_minutes` will only affect restore function test, I suggest update it dynamically like config cold backup interval.

levy5307:
> If `cold_backup_checkpoint_reserve_minutes` will only affect restore function test, I suggest update it dynamically like config cold backup interval for restore function test.

Done.

PR #523, opened 2020-04-24
refactor(collector): delete table_stats which is unnecessary

levy5307:
Delete table_stats which is unnecessary.  use row_stat instead.

### Check List 

Tests 


src/shell/command_helper.h
```diff
@@ -547,6 +565,42 @@ struct row_data
 
     double get_total_cu() const { return recent_read_cu + recent_write_cu; }
 
+    void aggregate(const row_data &row)
```
acelyc111, line 568: How about define a map that kv pair is: string metric_name -> double value?
- construct/init: register all metric names with value of 0 
- aggregate: we can iterate the map and aggregate values by the same key
- update_app_pegasus_perf_counter: find metric name in the map and increase the value
Then we only need to register new metric in constructor.

PR #524, opened 2020-04-26
refactor: Separate constructor of pegasus_server_impl to an independent file

acelyc111:
`pegasus_server_impl.cpp` is too large and too complex, we have to do some refactor on it.
This patch has no functional changes, just move pegasus_server_impl's constructor to an independent file. I'll going to do more refactors after this.


neverchanje:
It's not only the fault that `pegasus_server_impl` has a long constructor, but the definitions of configuration and perf-counter take too many lines of code.

```
    _abnormal_multi_get_iterate_count_threshold = dsn_config_get_value_uint64(
        "pegasus.server",
        "rocksdb_abnormal_multi_get_iterate_count_threshold",
        1000,
        "multi-get operation iterate count exceed this threshold will be logged, 0 means no check");
```

```
    snprintf(name, 255, "rdb.bf_seek_negatives@%s", str_gpid.c_str());
    _pfc_rdb_bf_seek_negatives.init_app_counter("app.pegasus",
                                                name,
                                                COUNTER_TYPE_NUMBER,
                                                "statistics the number of times bloom filter was "
                                                "checked before creating iterator on a file and "
                                                "useful in avoiding iterator creation (and thus "
                                                "likely IOPs)");
```

We can fix it by introducing some utilities, for example: https://github.com/XiaoMi/rdsn/pull/275

PR #525, opened 2020-04-27
fix: bind prometheus exposer to 0.0.0.0

neverchanje:

The prometheus exposer formerly bound to an external IP address, for example "10.231.52.xxx:9091", and served on the HTTP URL '10.231.52.xxx:9091/metrics'. But it's unable to serve on '0.0.0.0:9091' and '127.0.0.1:9091'. It is therefore inconvenient to set up an onebox environment with prometheus.

This limitation can be resolved by binding to '0.0.0.0:9091' so that '0.0.0.0:9091', '127.0.0.1:9091', '10.231.52.xxx:9091' are all accessible.

### Check List 

Tests 

Related changes

### DIFF
--- a/scripts/pack_server.sh
+++ b/scripts/pack_server.sh
@@ -106,6 +106,17 @@ copy_file `get_system_lib server aio` ${pack}/bin/`get_system_libname server aio
 copy_file `get_system_lib server zstd` ${pack}/bin/`get_system_libname server zstd`
 copy_file `get_system_lib server lz4` ${pack}/bin/`get_system_libname server lz4`
 
+DISTRIB_ID=$(cat /etc/*-release | grep DISTRIB_ID | awk -F'=' '{print $2}')
+DISTRIB_RELEASE=$(cat /etc/*-release | grep DISTRIB_RELEASE | awk -F'=' '{print $2}')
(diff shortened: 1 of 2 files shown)


scripts/pack_server.sh
```diff
@@ -106,6 +106,17 @@ copy_file `get_system_lib server aio` ${pack}/bin/`get_system_libname server aio
 copy_file `get_system_lib server zstd` ${pack}/bin/`get_system_libname server zstd`
 copy_file `get_system_lib server lz4` ${pack}/bin/`get_system_libname server lz4`
 
+DISTRIB_ID=$(cat /etc/*-release | grep DISTRIB_ID | awk -F'=' '{print $2}')
```
levy5307, line 109: what's the meaning of these code?

neverchanje: Because binary on linux distribution based on debian requires dependencies of icudata, icuuc, icui18n. If I build a binary on Ubuntu18.04, and want to distribute it to other platforms, like linux in docker, I need to include the libraries in my package.

See this:  https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2242

PR #526, opened 2020-04-28
feat(dup): add metric for time lag between master&slave

neverchanje:

We need to know the time lag between master and slave.  So in this PR, I added two perf-counters:

### New perf-counters

- `replica*app.pegasus*dup.time_lag_ms@<app_name>`

The time lag between master and slave. A write to master will first record a timestamp (`ts1`) before running 2PC. Then when it duplicates to the slave and applies to the slave's rocksdb (`ts2`), this counter will calculate the duration between the two timestamps.

We can set an alert for this metric in order to detect failures on duplication. 

- `replica*app.pegasus*dup.lagging_writes@<app_name>`

A write with time lag larger than `dup_lagging_write_threshold_ms` is defined as a "lagging write".
This counter calculates the number of such lagging writes.

### New configuration

- [pegasus.server] dup_lagging_write_threshold_ms

### Check List 

Tests 

Increases the duplication workload at 11:30~11:31, the lagging number bursts. While the workload going down, the time-lag decreases to 1s.

image

image

image

Related changes

### DIFF
--- a/src/server/pegasus_write_service.cpp
+++ b/src/server/pegasus_write_service.cpp
@@ -307,11 +307,11 @@ int pegasus_write_service::duplicate(int64_t decree,
 
     _pfc_duplicate_qps->increment();
     auto cleanup = dsn::defer([this, &request]() {
-        uint64_t latency_us = dsn_now_us() - request.timestamp;
-        if (latency_us > _dup_lagging_write_threshold_ms * 1000) {
+        uint64_t latency_ms = (dsn_now_us() - request.timestamp) / 1000;
+        if (latency_ms > _dup_lagging_write_threshold_ms) {
             _pfc_dup_lagging_writes->increment();
         }
-        _pfc_dup_time_lag->set(latency_us / 1000);
+        _pfc_dup_time_lag->set(latency_ms);
     });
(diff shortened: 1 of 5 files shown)


src/server/pegasus_write_service.cpp
```diff
@@ -286,6 +306,13 @@ int pegasus_write_service::duplicate(int64_t decree,
     }
 
     _pfc_duplicate_qps->increment();
+    auto cleanup = dsn::defer([this, &request]() {
+        uint64_t latency_us = dsn_now_us() - request.timestamp;
```
levy5307, line 310: I think you can get latency_ms here
```
uint64_t latency_ms = dsn_now_ms() - request.timestamp / 1000;
if (latency_ms > _dup_lagging_write_threshold_ms) {
            _pfc_dup_lagging_writes->increment();
        }
        _pfc_dup_time_lag->set(latency_ms);

```

acelyc111: > I think you can get latency_ms here
> 
> ```
> uint64_t latency_ms = dsn_now_ms() - request.timestamp / 1000;
> if (latency_ms > _dup_lagging_write_threshold_ms) {
>       _pfc_dup_lagging_writes->increment();
> }
> _pfc_dup_time_lag->set(latency_ms);
> ```

+1

PR #527, opened 2020-04-30
refactor: use flags API to define pegasus configs [WIP]

neverchanje:

This PR refactored the constructor of `pegasus_server_impl`, reduced the LOC from 472 to 312, and separate the config definitions to a single module. I believe this will make it more friendly for newbee of Pegasus.


acelyc111:
implemented by other PRs, tracked by https://github.com/apache/incubator-pegasus/issues/1323

PR #528, opened 2020-05-01
improvement(lb): add add_node_list.sh to add nodes with copy_pri after all copy_sec done

Shuo-Jia:
When we add new node for cluster, the copying operation will let the new node has very high load. And then the new node almost is unavailable.

This pr change the default order of `copying primary preferentially` to `copying secondary preferentially` by setting `only_move_primary = true` when we add new node and recover it after completed.

Before the total replica count is balanced, the new node won't have any primary until copying secondary completes. Because the new node won't serve user-requests, reads will not be influenced by the data-migration of replicas.

### Check List 

Tests (for the pr refactor the `offline_node_list.sh`, so need 2 manual test) 

- Manual test for `pegasus_online_node_list.sh` (add detailed scripts or steps below)
  * deploy cluster with minos
  * add new node for the cluster using the script
  * check the nodes using `nodes -d` and the falcon
  * the result is: the cluster copy secondary preferentially and the falcon show the new node don't serve for `read`:
  * default version:
选区_003
  * new version:
选区_002

- Manual test for `pegasus_offline_node_list.sh`(add detailed scripts or steps below)
  *  deploy cluster with minos and show the `running` replica server node = 5
  *  execute the `pegasus_offline_node_list.sh` and show the  `running` replica server node = 4

Related changes


scripts/minos_common.sh
```diff
@@ -123,3 +123,21 @@ function minos_restart()
...
+  local pwd=`pwd`
+  local options="--job $2"
+  if [ -n "$3" ]; then
+    options="$options --task $3"
+  fi
+  cd $minos_client_dir
+  echo "./deploy bootstrap pegasus $1 $options"
+  ./deploy bootstrap pegasus $1 $options
+  if [ $? -ne 0 ]; then
+    echo "ERROR: minos restart failed"
```
acelyc111, line 139: minos bootstrap failed

scripts/pegasus_online_node_list.sh
```diff
@@ -0,0 +1,183 @@
+#!/bin/bash
```
acelyc111, line 1: There are many duplicate code with `pegasus_offline_node_list.sh`, could you please combine the two scripts into a single one?

scripts/pegasus_online_node_list.sh
```diff
@@ -0,0 +1,118 @@
...
+  echo "ERROR: set meta level to steady failed"
+  exit 1
+fi
+
+for id in $task_id_list
+do
+  echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
+  minos_bootstrap $cluster replica $id
+  if [ $? -ne 0 ]; then
+    echo "ERROR: online replica task $id failed"
```
neverchanje, line 46:
```suggestion
    echo "ERROR: bootstrap replica task $id failed"
```

scripts/pegasus_online_node_list.sh
```diff
@@ -0,0 +1,118 @@
...
+    exit 1
+  fi
+  echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
+  echo "sleep for 10 seconds"
+  sleep 10
```
neverchanje, line 51: bootstrapping won't influence the cluster, no need to sleep.

scripts/pegasus_online_node_list.sh
```diff
@@ -0,0 +1,118 @@
...
+  echo "ERROR: meta.lb.only_move_primary true"
+  exit 1
+fi
+echo
+
+echo "Set meta level to lively..."
+echo "set_meta_level lively" | ./run.sh shell --cluster $meta_list &>/tmp/$UID.$PID.pegasus.online_node.set_meta_level
+set_ok=`grep 'control meta level ok' /tmp/$UID.$PID.pegasus.online_node.set_meta_level | wc -l`
+if [ $set_ok -ne 1 ]; then
+  echo "ERROR: set meta level to steady failed"
```
neverchanje, line 67:
```suggestion
  echo "ERROR: set meta level to lively failed"
```

scripts/pegasus_online_node_list.sh
```diff
@@ -0,0 +1,118 @@
...
+      op_count=`echo "cluster_info" | ./run.sh shell --cluster $meta_list | grep balance_operation_count | grep -o 'total=[0-9][0-9]*' | cut -d= -f2`
+      if [ $op_count -eq 0 ]; then
+        echo "Cluster becomes balanced."
+        break
+      fi
+    else
+      echo "Still $op_count balance operations to do..."
+      sleep 1
+    fi
+  done
```
neverchanje, line 92:
```suggestion
sleep 180
while true; do
    op_count=$(echo "cluster_info" | ./run.sh shell --cluster $meta_list | grep balance_operation_count | grep -o 'total=[0-9][0-9]*' | cut -d= -f2)
    if [ -z "op_count" ]; then
        break
    fi
    if [ $op_count -eq 0 ]; then
        echo "Cluster may be balanced, try wait 30 seconds..."
        sleep 30
        op_count=$(echo "cluster_info" | ./run.sh shell --cluster $meta_list | grep balance_operation_count | grep -o 'total=[0-9][0-9]*' | cut -d= -f2)
        if [ $op_count -eq 0 ]; then
            echo "Cluster becomes balanced."
            break
        fi
    else
        echo "Still $op_count balance operations to do..."
        sleep 1
    fi
done
```

why make indentions here?

scripts/pegasus_online_node_list.sh
```diff
@@ -0,0 +1,118 @@
...
+
+echo "Set meta level to steady..."
+echo "set_meta_level steady" | ./run.sh shell --cluster $meta_list &>/tmp/$UID.$PID.pegasus.online_node.set_meta_level
+set_ok=`grep 'control meta level ok' /tmp/$UID.$PID.pegasus.online_node.set_meta_level | wc -l`
+if [ $set_ok -ne 1 ]; then
+  echo "ERROR: set meta level to steady failed"
+  exit 1
+fi
+
+echo "set meta.lb.only_move_primary false"
```
neverchanje, line 105: So the procedure is:

1. meta set only_move_primary = true.
2. rebalance nodes
3. meta set only_move_primary = false.
4. rebalance nodes

I'm confused about why we need step 1 and step 2? As far as I can see, step1+2 rebalance the primaries among the old nodes, but they touch nothing on the new node.

scripts/pegasus_online_node_list.sh
```diff
@@ -0,0 +1,118 @@
...
+  if [ $? -ne 0 ]; then
+    echo "ERROR: online replica task $id failed"
+    exit 1
+  fi
+  echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
+  echo "sleep for 10 seconds"
+  sleep 10
+done
+
+echo "set meta.lb.only_move_primary true"
```
neverchanje, line 54: echo "This remote-command tells the meta-server to ignore copying primaries during rebalancing."
echo "So the following steps only include move_primary and copy_secondary."

neverchanje, line 54:
```suggestion
echo "set meta.lb.only_move_primary true"
echo "This remote-command tells the meta-server to ignore copying primaries during rebalancing."
echo "So the following steps only include move_primary and copy_secondary."
```

scripts/pegasus_online_node_list.sh
neverchanje, line 106:
```suggestion
echo "set meta.lb.only_move_primary false"
echo "This remote-command tells the meta-server to rebalance with copying primaries."
```

PR #529, opened 2020-05-06
improvement: add logging on rocksdb write stalls

neverchanje:

image

Pegasus has metrics for rocksdb write-stalls, but no logging for which replica that was stalled.
So even we have been alerted for this event, we are not able to improve the situation. 

I add logs for each rocksdb write stall, including delayed and stopped.

### Check List 

Tests 

Related changes

### DIFF
--- a/src/server/pegasus_event_listener.cpp
+++ b/src/server/pegasus_event_listener.cpp
@@ -3,11 +3,14 @@
 // can be found in the LICENSE file in the root directory of this source tree.
 
 #include "pegasus_event_listener.h"
+#include "logging_utils.h"
+
+#include <dsn/c/api_utilities.h>
 
 namespace pegasus {
 namespace server {
 
-pegasus_event_listener::pegasus_event_listener()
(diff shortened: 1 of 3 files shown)


src/server/pegasus_event_listener.cpp
```diff
@@ -61,10 +62,13 @@ void pegasus_event_listener::OnCompactionCompleted(rocksdb::DB *db,
 
 void pegasus_event_listener::OnStallConditionsChanged(const rocksdb::WriteStallInfo &info)
 {
-    if (info.condition.cur == rocksdb::WriteStallCondition::kDelayed)
+    if (info.condition.cur == rocksdb::WriteStallCondition::kDelayed) {
+        derror_replica("rocksdb write delayed");
```
hycdong, line 66: Printing log every stalling, will it produce too many error logs?

neverchanje: Yes. But this is a serious error. We print error logs on each serious error like 2PC write failure or read failure. Mostly write-stalls are followed by many write failures.

PR #535, opened 2020-05-14
feat(rocksdb): Use original CreateCheckpoint and read meta data from meta CF

acelyc111:
Ref: https://github.com/XiaoMi/pegasus/issues/468
We have store meta data(Pegasus data version, last manual compact timestamp, decree) both in RocksDB's manifest and meta column family in version [2.0.0](https://github.com/XiaoMi/pegasus/releases/tag/v2.0.0), now we are going to read meta data from meta CF instead of manifest.
Main changes of this PR:
- Validate meta CF existence when DB exist (means only allowed to upgrade from 2.0)
- Read meta data from meta CF
- Do checkpoint by rocksdb's original APIs.

**NOTE:** Since meta data is still wrote into manifest, so there is no compatibility problems with 2.0.0.

No user faced changes.

### Check List 

Tests 

1. Upgrade from 1.x.x
  Will fail, only allowed from version 2.0.x
2. Upgrade from 2.0.0
  - Deploy a cluster in version 2.0.0
  - Write some data
  - Upgrade to this commit's version
  - Read these data successfully

Code changes

No
No
No
- Has persistent data change
No

Side effects

No
- Increased code complexity
No
No

Related changes

Yes
Yes
Yes

### DIFF
--- a/src/server/meta_store.cpp
+++ b/src/server/meta_store.cpp
@@ -12,7 +12,7 @@ namespace server {
 
 DSN_DEFINE_string("pegasus.server",
                   get_meta_store_type,
-                  "manifest",
+                  "metacf",
                   "Where to get meta data, now support 'manifest' and 'metacf'");
 DSN_DEFINE_validator(get_meta_store_type, [](const char *type) {
     return strcmp(type, "manifest") == 0 || strcmp(type, "metacf") == 0;
@@ -84,9 +84,22 @@ uint64_t meta_store::get_last_manual_compact_finish_time() const
     }
 }
 
+uint64_t meta_store::get_last_flushed_decree_from_checkpoint(rocksdb::DB *db, rocksdb::ColumnFamilyHandle* meta_cf) const
+{
+    uint64_t last_flushed_decree = 0;
+    auto ec = get_value_from_meta_cf(db, meta_cf, true, LAST_FLUSHED_DECREE, &last_flushed_decree);
+    dcheck_eq_replica(::dsn::ERR_OK, ec);
+    return last_flushed_decree;
+}
+
 ::dsn::error_code meta_store::get_value_from_meta_cf(bool read_flushed_data,
                                                      const std::string &key,
                                                      uint64_t *value) const
+{
+    return get_value_from_meta_cf(_db, _meta_cf, read_flushed_data, key, value);
+}
+
+::dsn::error_code meta_store::get_value_from_meta_cf(rocksdb::DB *db, rocksdb::ColumnFamilyHandle *cf, bool read_flushed_data, const std::string &key, uint64_t *value)
 {
     std::string data;
     rocksdb::ReadOptions rd_opts;
@@ -94,14 +107,14 @@ uint64_t meta_store::get_last_manual_compact_finish_time() const
         // only read 'flushed' data, mainly to read 'last_flushed_decree'
         rd_opts.read_tier = rocksdb::kPersistedTier;
     }
-    auto status = _db->Get(rd_opts, _meta_cf, key, &data);
+    auto status = db->Get(rd_opts, cf, key, &data);
     if (status.ok()) {
         bool ok = dsn::buf2uint64(data, *value);
-        dassert_replica(ok,
-                        "rocksdb {} get {} from meta column family got error value {}",
-                        _db->GetName(),
-                        key,
-                        data);
+        dassert(ok,
+                "rocksdb {} get {} from meta column family got error value {}",
+                db->GetName(),
+                key,
+                data);
         return ::dsn::ERR_OK;
     }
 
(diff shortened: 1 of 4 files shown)


src/server/pegasus_server_impl.cpp
```diff
@@ -1326,16 +1326,16 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
...
-        // column family), or create new db, we have to create a new column family to store meta
-        // data (meta column family).
+    if (db_exist) {
+        // When DB exist, meta CF must be present.
+        bool missing_meta_cf = true;
+        if (check_meta_cf(path, &missing_meta_cf) != ::dsn::ERR_OK) {
+            derror_replica("check meta column family failed");
+            return ::dsn::ERR_LOCAL_APP_FAILURE;
+        }
+        dassert_replica(!missing_meta_cf, "You must upgrade Pegasus server from 2.0");
```
Shuo-Jia, line 1336: “NOTE: Since meta data is still wrote into manifest, so there is no compatibility problems with 2.0.0.”

I understand the pr resolve the `compatibility problems`, so is here still dassert because `compatibility`?

acelyc111: This pr just keep compatibility with 2.0.0, not compatible with 1.x, so if upgrade from 1.x, it will dassert false here.

src/server/meta_store.cpp
```diff
@@ -84,24 +84,42 @@ uint64_t meta_store::get_last_manual_compact_finish_time() const
...
-    auto status = _db->Get(rd_opts, _meta_cf, key, &data);
+    auto status = db->Get(rd_opts, cf, key, &data);
     if (status.ok()) {
-        bool ok = dsn::buf2uint64(data, *value);
-        dassert_replica(ok,
-                        "rocksdb {} get {} from meta column family got error value {}",
-                        _db->GetName(),
-                        key,
-                        data);
+        dassert(dsn::buf2uint64(data, *value),
```
hycdong, line 118: dassert_replica may be better.

src/server/pegasus_server_impl.cpp
```diff
@@ -1739,19 +1742,47 @@ ::dsn::error_code pegasus_server_impl::copy_checkpoint_to_dir_unsafe(const char
...
+        rocksdb::DB *snapshot_db = nullptr;
+        std::vector<rocksdb::ColumnFamilyHandle *> handles_opened;
+        auto cleanup = [&](bool remove_checkpoint) {
+            if (remove_checkpoint && !::dsn::utils::filesystem::remove_path(checkpoint_dir)) {
+                derror_replica("remove checkpoint directory {} failed", checkpoint_dir);
+            }
+            release_db(snapshot_db, handles_opened);
+        };
+
+        // always need to open default column family even though not use it
```
hycdong, line 1766: Why should always open default column family?

src/server/pegasus_server_impl.cpp
```diff
@@ -1739,19 +1742,47 @@ ::dsn::error_code pegasus_server_impl::copy_checkpoint_to_dir_unsafe(const char
...
+            cleanup(true);
+            return ::dsn::ERR_LOCAL_APP_FAILURE;
+        }
+        dcheck_eq_replica(handles_opened.size(), 2);
+        dcheck_eq_replica(handles_opened[1]->GetName(), META_COLUMN_FAMILY_NAME);
+        uint64_t last_flushed_decree =
+            _meta_store->get_last_flushed_decree_from_checkpoint(snapshot_db, handles_opened[1]);
+        *checkpoint_decree = last_flushed_decree;
+
+        cleanup(false);
```
hycdong, line 1785: We used to use pegasus-defined function `CreateCheckpointQuick` to create a checkpoint without flushing memtable and get last_flushed_decree. This pr uses official interface CreateCheckpoint instead, we should parse last_flushed_decree from this checkpoint, which L1757 - L1785 does.

If my view is correct, I suggest rename function `get_last_flushed_decree_from_checkpoint`, this function is not get last_flushed_decree from a checkpoint, but from a snapshot db readonly opened by checkpoint.

src/server/meta_store.cpp
acelyc111: `get_value_from_meta_cf` is a static function now, it can't access member function `replica_name()` which `dassert_replica` need.

src/server/pegasus_server_impl.cpp
acelyc111: If not open default column family, an error will return.
https://github.com/XiaoMi/pegasus-rocksdb/blob/v6.6.4-base/include/rocksdb/db.h#L155

src/server/pegasus_server_impl.cpp
acelyc111: I've renamed it to `get_decree_from_readonly_db`

src/server/pegasus_server_impl.cpp
levy5307, line 1766: Could you add some explain to show why always need to open default column family even though not use it?

PR #537, opened 2020-05-14
fix: fix bug in table_hotspot_policy

Smityz:
if you do like this, there will be a core dump.
```bash
create aaa -p 8 -r 3
# wait a moment
drop aaa
create aaa -p 16 -r 3
```
Because the previous design cannot change the number of partitions of table after the Collector starts.
And now I store a hotspot_calculator named `app_name+partition_num`

### DIFF
--- a/src/server/info_collector.cpp
+++ b/src/server/info_collector.cpp
@@ -306,7 +306,7 @@ void info_collector::on_storage_size_stat(int remaining_retry_count)
 hotspot_calculator *info_collector::get_hotspot_calculator(const std::string &app_name,
                                                            const int partition_num)
 {
-    std::string app_name_count = app_name + std::string(partition_num);
+    std::string app_name_count = app_name + std::to_string(partition_num);
     auto iter = _hotspot_calculator_store.find(app_name_count);
     if (iter != _hotspot_calculator_store.end()) {
         return iter->second;
--- a/src/server/info_collector.cpp
+++ b/src/server/info_collector.cpp
@@ -307,7 +307,7 @@ hotspot_calculator *info_collector::get_hotspot_calculator(const std::string &ap
 {
     // use appname+partition_num as a key can prevent the impact of dynamic partition changes
     std::string app_name_pcount = fmt::format("{}.{}", app_name, partition_num);
-    auto iter = _hotspot_calculator_store.find(app_name_count);
+    auto iter = _hotspot_calculator_store.find(app_name_pcount);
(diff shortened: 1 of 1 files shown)


src/server/info_collector.cpp
```diff
@@ -304,7 +304,8 @@ void info_collector::on_storage_size_stat(int remaining_retry_count)
 hotspot_calculator *info_collector::get_hotspot_calculator(const std::string &app_name,
                                                            const int partition_num)
 {
-    auto iter = _hotspot_calculator_store.find(app_name);
+    std::string app_name_count = app_name + std::to_string(partition_num);
+    auto iter = _hotspot_calculator_store.find(app_name_count);
```
Shuo-Jia, line 308: You may need add comment, otherwise the `key` is set `app_name + partition_num` may confuse others when read the code.

src/server/info_collector.cpp
```diff
@@ -304,7 +304,9 @@ void info_collector::on_storage_size_stat(int remaining_retry_count)
 hotspot_calculator *info_collector::get_hotspot_calculator(const std::string &app_name,
                                                            const int partition_num)
 {
-    auto iter = _hotspot_calculator_store.find(app_name);
+    // use appname+partition_num as a key can prevent the impact of dynamic partition changes
```
acelyc111, line 307: There must be problem when:
```
create test -p 64
```
and then
```
create test6 -p 4
```

src/server/info_collector.cpp
```diff
@@ -304,7 +304,9 @@ void info_collector::on_storage_size_stat(int remaining_retry_count)
...
+    std::string app_name_count = app_name + std::to_string(partition_num);
```
levy5307, line 308: fmt::format({}.{}, app_name, partition_num)

src/server/info_collector.cpp
```diff
@@ -304,7 +305,9 @@ void info_collector::on_storage_size_stat(int remaining_retry_count)
 hotspot_calculator *info_collector::get_hotspot_calculator(const std::string &app_name,
                                                            const int partition_num)
 {
-    auto iter = _hotspot_calculator_store.find(app_name);
+    // use appname+partition_num as a key can prevent the impact of dynamic partition changes
+    std::string app_name_count = fmt::format("{}.{}", app_name, partition_num);
```
neverchanje, line 309:
```suggestion
    std::string app_name_pcount = fmt::format("{}.{}", app_name, partition_num);
```

PR #538, opened 2020-05-18
fix(shell/sds): NULL check before memset

l2dy:
Potential use of NULL pointer.

Check for NULL before `memset()`.


l2dy:
See also https://github.com/antirez/sds/commit/6fa721cd8bc7ae6b2753f33573c1cd8778b9a8cb.

Alternative solution is to sync with upstream.

acelyc111:
> See also [antirez/sds@6fa721c](https://github.com/antirez/sds/commit/6fa721cd8bc7ae6b2753f33573c1cd8778b9a8cb).
> 
> Alternative solution is to sync with upstream.

acelyc111:
@l2dy Thanks for your contribute!

PR #542, opened 2020-05-29
fix: fix bug in function_test/test_restore

levy5307:
In restore test, it looks for the second backup timestamp to see if the first backup is complete.
But it doesn't take the `args`, which means the directory of block service, into accout.
So if we set `args`(like https://github.com/XiaoMi/pegasus/pull/534 shows), it can't find the second backup timestamp, and result in a timeout.

### Check List 

Tests 

- Unit test

Code changes

### DIFF
--- a/src/test/function_test/test_restore.cpp
+++ b/src/test/function_test/test_restore.cpp
@@ -37,6 +37,11 @@ public:
         system("./run.sh start_onebox --config_path config-server-test-restore.ini");
         std::this_thread::sleep_for(std::chrono::seconds(3));
 
+        // First of all, we are in the path of pegasus root, for example: /home/mi/pegasus.
+        // And we can get the provider_dir which actually is `block_service/local_service`,
+        // from config-server-test-restore.ini.
+        // With cluster_name = mycluster and policy_name = policy_1, we can get the absolute
+        // path of policy: /home/mi/pegasus/onebox/block_service/local_service/mycluster/policy_1
         cmd = "grep -A 5 block_service." + backup_provider_name +
               " config-server-test-restore.ini | grep args | cut -f2,3 -d'/'";
         std::stringstream ss;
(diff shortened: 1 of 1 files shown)


src/test/function_test/test_restore.cpp
```diff
@@ -39,9 +37,17 @@ class restore_test : public testing::Test
         system("./run.sh start_onebox --config_path config-server-test-restore.ini");
         std::this_thread::sleep_for(std::chrono::seconds(3));
 
+        cmd = "grep -A 5 block_service." + backup_provider_name +
+              " config-server-test-restore.ini | grep args | cut -f2,3 -d'/'";
+        std::stringstream ss;
+        assert(dsn::utils::pipe_execute(cmd.c_str(), ss) == 0);
+        std::string provider_dir = ss.str().substr(0, ss.str().length() - 1);
+        policy_dir = "onebox/" + provider_dir + '/' +
+                     dsn::utils::filesystem::path_combine(cluster_name, policy_name);
```
hycdong, line 46: Could you please add some comments here to show the example path of policy dir?

src/test/function_test/test_restore.cpp
```diff
@@ -19,10 +19,8 @@ class restore_test : public testing::Test
     virtual void SetUp() override
     {
         pegasus_root_dir = global_env::instance()._pegasus_root;
-        working_root_dir = global_env::instance()._working_dir;
 
         chdir(pegasus_root_dir.c_str());
-        cluster_name = dsn::utils::filesystem::path_combine(pegasus_root_dir, backup_data_dir);
```
hycdong, line 25: It seems that `backup_data_dir` is useless, I think you can remove this variety.

src/test/function_test/test_restore.cpp
```diff
@@ -256,19 +267,20 @@ class restore_test : public testing::Test
...
     {
-        std::string policy_dir = dsn::utils::filesystem::path_combine(cluster_name, policy_name);
         std::vector<std::string> dirs;
         ::dsn::utils::filesystem::get_subdirectories(policy_dir, dirs, false);
         return (dirs.size() >= 2);
     }
 
 public:
     pegasus_client *pg_client;
+    pegasus_client *new_pg_client;
```
hycdong, line 277: In my view, `new_pg_client` is used to operate restore table?

levy5307: Yes, you are right.

PR #546, opened 2020-06-05
feat(bulk-load): bulk load ingestion part3 - execute ingestion

hycdong:
This pull request is about how pegasus handle ingestion request.
Ingestion request is considered as a write request, but it is different from normal write request. Normal write request will not only write data but also write decree into rocksdb, but rocksdb ingestion won't write decree, so we should write an empty put to flush decree before execute ingestion. Besides, ingestion is time-consuming, it should not be executed in replication thread pool, so ingestion will execute in ingestion thread pool asynchronously, the result of ingestion will be got by ingestion_status. Before rocksdb do actual ingestion, we should verify files again.

### DIFF
--- a/src/server/pegasus_server_impl.h
+++ b/src/server/pegasus_server_impl.h
@@ -158,6 +158,7 @@ public:
 
     std::string dump_write_request(dsn::message_ex *request) override;
 
+    // Not thread-safe
     void set_ingestion_status(dsn::replication::ingestion_status::type status) override;
 
     dsn::replication::ingestion_status::type get_ingestion_status() override
--- a/src/server/pegasus_server_impl.h
+++ b/src/server/pegasus_server_impl.h
@@ -158,9 +158,9 @@ public:
 
     std::string dump_write_request(dsn::message_ex *request) override;
 
-    void set_ingestion_status(::dsn::replication::ingestion_status::type status) override;
+    void set_ingestion_status(dsn::replication::ingestion_status::type status) override;
 
-    ::dsn::replication::ingestion_status::type get_ingestion_status() override
+    dsn::replication::ingestion_status::type get_ingestion_status() override
     {
         return _ingestion_status;
     }
@@ -371,8 +371,8 @@ private:
 
     std::atomic<int32_t> _partition_version;
 
-    ::dsn::replication::ingestion_status::type _ingestion_status{
-        ::dsn::replication::ingestion_status::IS_INVALID};
+    dsn::replication::ingestion_status::type _ingestion_status{
+        dsn::replication::ingestion_status::IS_INVALID};
 
     dsn::task_tracker _tracker;
 
--- a/src/server/pegasus_server_impl.h
+++ b/src/server/pegasus_server_impl.h
@@ -158,6 +158,13 @@ public:
 
     std::string dump_write_request(dsn::message_ex *request) override;
 
+    void set_ingestion_status(::dsn::replication::ingestion_status::type status) override;
+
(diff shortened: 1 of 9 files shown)


src/server/pegasus_server_impl.h
```diff
@@ -158,6 +158,13 @@ class pegasus_server_impl : public ::dsn::apps::rrdb_service
 
     std::string dump_write_request(dsn::message_ex *request) override;
 
+    void set_ingestion_status(::dsn::replication::ingestion_status::type status) override;
+
+    ::dsn::replication::ingestion_status::type get_ingestion_status() override
```
levy5307, line 163: remove the first `::`

src/server/pegasus_write_service_impl.h
```diff
@@ -806,6 +833,23 @@ class pegasus_write_service::impl : public dsn::replication::replica_base
...
 
+    dsn::error_code get_external_files_path(const std::string &bulk_load_dir,
+                                            const dsn::replication::bulk_load_metadata &metadata,
+                                            /*out*/ std::vector<std::string> &files_path)
+    {
+        for (const auto &f_meta : metadata.files) {
+            // const dsn::replication::file_meta &f_meta = *file;
+            const std::string &file_name =
+                dsn::utils::filesystem::path_combine(bulk_load_dir, f_meta.name);
+            if (dsn::utils::filesystem::verify_file(file_name, f_meta.md5, f_meta.size)) {
```
levy5307, line 844: Why should you verfiy these files again?

hycdong: It's for double check, it will cost long time during files downloaded and ingestion.

src/server/pegasus_server_impl.h
```diff
@@ -158,6 +158,13 @@ class pegasus_server_impl : public ::dsn::apps::rrdb_service
 
     std::string dump_write_request(dsn::message_ex *request) override;
 
+    void set_ingestion_status(dsn::replication::ingestion_status::type status) override;
```
neverchanje, line 161:
```suggestion
    // Not thread-safe.
    void set_ingestion_status(dsn::replication::ingestion_status::type status) override;
```

src/server/pegasus_write_service_impl.h
```diff
@@ -806,6 +833,23 @@ class pegasus_write_service::impl : public dsn::replication::replica_base
         return expire_ts;
     }
 
+    dsn::error_code get_external_files_path(const std::string &bulk_load_dir,
```
neverchanje, line 836: Should be a global function.

src/server/pegasus_write_service_impl.h
```diff
@@ -806,6 +833,23 @@ class pegasus_write_service::impl : public dsn::replication::replica_base
...
+                                            const dsn::replication::bulk_load_metadata &metadata,
+                                            /*out*/ std::vector<std::string> &files_path)
+    {
+        for (const auto &f_meta : metadata.files) {
+            // const dsn::replication::file_meta &f_meta = *file;
```
neverchanje, line 841: remove redundant comments.

src/server/pegasus_write_service_impl.h
hycdong: Why should this function be a global function?

neverchanje: Because it doesn't touch any member of this class. It's merely a util. Setting it global helps shortening the size of pegasus_write_service_impl.

hycdong: okay

PR #548, opened 2020-06-10
feat(rocksdb): Support to open rocksdb with or without meta CF

acelyc111:
[Pegasus 2.0](https://github.com/XiaoMi/pegasus/issues/536) has released many new features, as well as many code refactors and bugfixes since the previous release version [1.12.3](https://github.com/XiaoMi/pegasus/releases/tag/v1.12.3). And more important, 2.0 will introduce some uncompatible features, such as update rocksdb from 5.9.2 to 6.6.4, update 'pegasus data version' from 0 to 1, etc.
This patch provide an effort for Pegasus 1.x to be forward compatible with 2.0. When upgrade Pegasus server from 1.x to 2.0 and find some bugs in 2.0, you can downgrade Pegasus to this version.

**NOTE**: 
- Pegasus version (let's suppose it is 1.12.4) with this patch is forward compatible with 2.0.0
- Pegasus 1.12.3 and lower versions are not forward compatible with this patch!!!

### Check List 

Tests 

1. bootstrap a new cluster, with version 1.12.3
2. write and verify data set A
3. upgrade to version 2.0
4. write data set B, verify data set A and B
5. downgrade to version 1.12.4 (the version include this patch)
6. write data set C, verify data set A, B and C
7. upgrade to version 2.0
8. write data set D, verify data set A, B, C and D

Side effects

No
- Increased code complexity
No
No

Related changes

No
Yes
Yes


src/server/pegasus_server_impl.cpp
```diff
@@ -2893,5 +2916,39 @@ void pegasus_server_impl::set_partition_version(int32_t partition_version)
...
+
+    for (const auto &column_family : column_families) {
+        if (column_family == DATA_COLUMN_FAMILY_NAME) {
+            continue;
+        }
+        if (column_family == META_COLUMN_FAMILY_NAME) {
+            *need_open_with_meta_cf = true;
+            continue;
+        }
+        dassert_replica(true, "Column family '{}' should not present");
```
neverchanje, line 2938:
```suggestion
        dassert_replica(false, "Column family '{}' should not present");
```

acelyc111: data CF and meta CF have been checked above, this is a new CF we can't meet here

PR #549, opened 2020-06-11
feat(rocksdb): Support to open rocksdb with or without meta CF

acelyc111:
[Pegasus 2.0](https://github.com/XiaoMi/pegasus/issues/536) has released many new features, as well as many code refactors and bugfixes since the previous release version [1.12.3](https://github.com/XiaoMi/pegasus/releases/tag/v1.12.3). And more important, 2.0 will introduce some uncompatible features, such as update rocksdb from 5.9.2 to 6.6.4, update 'pegasus data version' from 0 to 1, etc.
This patch provide an effort for Pegasus 1.x to be forward compatible with 2.0. When upgrade Pegasus server from 1.x to 2.0 and find some bugs in 2.0, you can downgrade Pegasus to this version.

**NOTE**: 
- Pegasus version (let's suppose it is 1.12.4) with this patch is forward compatible with 2.0.0
- Pegasus 1.12.3 and lower versions are not forward compatible with this patch!!!

### Check List 

Tests 

1. bootstrap a new cluster, with version 1.12.3
2. write and verify data set A
3. upgrade to version 2.0
4. write data set B, verify data set A and B
5. downgrade to version 1.12.4 (the version include this patch)
6. write data set C, verify data set A, B and C
7. upgrade to version 2.0
8. write data set D, verify data set A, B, C and D

Side effects

No
- Increased code complexity
No
No

Related changes

No
Yes
Yes

### DIFF
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -2944,7 +2944,10 @@ void pegasus_server_impl::release_db()
 {
     _db->DestroyColumnFamilyHandle(_data_cf);
     _data_cf = nullptr;
-    _db->DestroyColumnFamilyHandle(_meta_cf);
+    if (_meta_cf != nullptr) {
+      _db->DestroyColumnFamilyHandle(_meta_cf);
+      _meta_cf = nullptr;
+    }
     _meta_cf = nullptr;
     delete _db;
     _db = nullptr;
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -48,6 +48,8 @@ std::shared_ptr<rocksdb::Cache> pegasus_server_impl::_s_block_cache;
 ::dsn::task_ptr pegasus_server_impl::_update_server_rdb_stat;
 ::dsn::perf_counter_wrapper pegasus_server_impl::_pfc_rdb_block_cache_mem_usage;
 const std::string pegasus_server_impl::COMPRESSION_HEADER = "per_level:";
+const std::string pegasus_server_impl::DATA_COLUMN_FAMILY_NAME = "default";
+const std::string pegasus_server_impl::META_COLUMN_FAMILY_NAME = "pegasus_meta_cf";
 
 pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
     : dsn::apps::rrdb_service(r),
@@ -452,8 +454,7 @@ pegasus_server_impl::~pegasus_server_impl()
 {
     if (_is_open) {
         dassert(_db != nullptr, "");
-        delete _db;
-        _db = nullptr;
+        release_db();
     }
 }
 
@@ -1647,6 +1648,7 @@ void pegasus_server_impl::on_clear_scanner(const int64_t &args) { _context_cache
     //          2, we can parse restore info from app env, which is stored in argv
     //          3, restore_dir is exist
     //
+    bool db_exist = true;
     auto path = ::dsn::utils::filesystem::path_combine(data_dir(), "rdb");
     if (::dsn::utils::filesystem::path_exists(path)) {
         // only case 1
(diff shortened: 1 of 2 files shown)


src/server/pegasus_server_impl.cpp
```diff
@@ -2894,5 +2916,39 @@ void pegasus_server_impl::set_partition_version(int32_t partition_version)
...
+
+    for (const auto &column_family : column_families) {
+        if (column_family == DATA_COLUMN_FAMILY_NAME) {
+            continue;
+        }
+        if (column_family == META_COLUMN_FAMILY_NAME) {
+            *need_open_with_meta_cf = true;
+            continue;
+        }
+        dassert_replica(true, "Column family '{}' should not present");
```
neverchanje, line 2938:
```suggestion
        dassert_replica(false, "Column family '{}' should not present");
```

acelyc111: data CF and meta CF have been checked above, this is a new CF we can't meet here

src/server/pegasus_server_impl.cpp
```diff
@@ -2894,5 +2916,39 @@ void pegasus_server_impl::set_partition_version(int32_t partition_version)
...
+    }
+    return ::dsn::ERR_OK;
+}
+
+void pegasus_server_impl::release_db()
+{
+    _db->DestroyColumnFamilyHandle(_data_cf);
+    _data_cf = nullptr;
+    _db->DestroyColumnFamilyHandle(_meta_cf);
```
neverchanje, line 2947:
```suggestion
    if(_meta_cf != nullptr) {
      _db->DestroyColumnFamilyHandle(_meta_cf);
      _meta_cf = nullptr;
    }
```

src/server/pegasus_server_impl.cpp
neverchanje: Yes. But if you are sure that impossible to run into this statement, why not use `dassert(false`?
`dassert(true` is the same as a piece of comment. It makes reader wonder why you write this assert rather than simply a comment.

acelyc111: I think I've made a mistake..

src/server/pegasus_server_impl.cpp
```diff
@@ -2894,5 +2916,42 @@ void pegasus_server_impl::set_partition_version(int32_t partition_version)
...
+    return ::dsn::ERR_OK;
+}
+
+void pegasus_server_impl::release_db()
+{
+    _db->DestroyColumnFamilyHandle(_data_cf);
+    _data_cf = nullptr;
+    if (_meta_cf != nullptr) {
+      _db->DestroyColumnFamilyHandle(_meta_cf);
+      _meta_cf = nullptr;
```
neverchanje, line 2949:
```suggestion
```

src/server/pegasus_server_impl.cpp
```diff
@@ -1702,16 +1706,36 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
 
     ddebug("%s: start to open rocksDB's rdb(%s)", replica_name(), path.c_str());
 
-    auto status = rocksdb::DB::Open(rocksdb::Options(_db_opts, _data_cf_opts), path, &_db);
+    bool need_open_with_meta_cf = false;
+    // Check meta CF only when db exist.
+    if (db_exist && check_meta_cf(path, &need_open_with_meta_cf) != ::dsn::ERR_OK) {
+        derror_replica("check meta column family failed");
+        return ::dsn::ERR_LOCAL_APP_FAILURE;
```
levy5307, line 1713: remove the first `::`

acelyc111: Seems most existing code has `::` before `dsn`, if want to remove all leading `::`, I think we can commit another pr to do it.

PR #550, opened 2020-06-11
refactor: use flag api to define configs in pegasus_reporter

neverchanje:

This PR redefines the configurations in pegasus_reporter module using flag API. No logical changes.

### Check List 

Related changes


src/reporter/pegasus_counter_reporter.cpp
```diff
@@ -2,28 +2,47 @@
...
+DSN_DEFINE_string("pegasus.server", falcon_host, "127.0.0.1", "falcon agent host");
+DSN_DEFINE_uint64("pegasus.server", falcon_port, 1988, "falcon agent port");
+DSN_DEFINE_string("pegasus.server", falcon_path, "/v1/push", "falcon http path");
+
+DSN_DEFINE_uint64("pegasus.server",
+                  perf_counter_update_interval_seconds,
+                  10,
+                  "perf_counter_update_interval_seconds");
+DSN_DEFINE_bool("pegasus.server", perf_counter_enable_logging, true, "perf_counter_enable_logging");
+DSN_DEFINE_string("pegasus.server", perf_counter_sink, "", "perf_counter_sink");
```
acelyc111, line 41: Reorder as top-down, or dict order by config key would be better.

src/reporter/pegasus_counter_reporter.cpp
```diff
@@ -2,28 +2,47 @@
...
-#include <dsn/dist/fmt_logging.h>
+#include "pegasus_counter_reporter.h"
 
 using namespace ::dsn;
 
+DSN_DEFINE_uint64("pegasus.server", prometheus_port, 9091, "prometheus exposer port");
+
+DSN_DEFINE_string("pegasus.server", falcon_host, "127.0.0.1", "falcon agent host");
+DSN_DEFINE_uint64("pegasus.server", falcon_port, 1988, "falcon agent port");
+DSN_DEFINE_string("pegasus.server", falcon_path, "/v1/push", "falcon http path");
```
acelyc111, line 34:
```suggestion
DSN_DEFINE_string("pegasus.server", falcon_path, "/v1/push", "falcon agent http path");
```

src/reporter/pegasus_counter_reporter.cpp
neverchanje: I don't get your point. What's top-down order by config-key? Could you please write a suggestion showing that?

acelyc111: There are 3 types of config group, top-down order is:
- configs for server            # top level, we can select prometheus or falcon
- configs for prometheus  # lower level
- configs for falcon            # lower level
So when developers read code here, he know it's useless to config falcon items when select prometheus sink.

src/reporter/pegasus_counter_reporter.cpp
```diff
@@ -2,28 +2,48 @@
...
 
 #include "base/pegasus_utils.h"
 #include "pegasus_io_service.h"
-
-#include <chrono>
-#include <map>
-#include <memory>
-#include <string>
-#include <dsn/dist/fmt_logging.h>
+#include "pegasus_counter_reporter.h"
```
acelyc111, line 26: The related header should be placed at first of all.

PR #551, opened 2020-06-12
refactor: refactor rrdb_server into pegasus _read_service

Smityz:
This RP refactors rrdb_service because there are many useless codes, It will make the reading and writing path clearer.

No


src/server/pegasus_rrdb.server.h
```diff
@@ -0,0 +1,103 @@
+#pragma once
+#include <iostream>
+#include <dsn/dist/replication/replication_app_base.h>
+#include <dsn/dist/replication/storage_serverlet.h>
+#include <rrdb/rrdb.code.definition.h>
+
+namespace pegasus {
+namespace server {
+class pegasus_rrdb_service : public dsn::replication::replication_app_base,
```
neverchanje, line 9: call it 'pegasus_read_service'. it's really hard to remember what rrdb means.

src/server/pegasus_server_impl.h
```diff
@@ -28,7 +28,7 @@ class meta_store;
 class capacity_unit_calculator;
 class pegasus_server_write;
 
-class pegasus_server_impl : public ::dsn::apps::rrdb_service
+class pegasus_server_impl : public pegasus_read_service
```
levy5307, line 31: What about change the name of this class to `pegasus_read_service_impl`?

Smityz: https://github.com/Smityz/pegasus/blob/master/src/server/pegasus_server_impl.cpp#L577
Here are also some codes related to write

PR #553, opened 2020-06-17
fix: fix incorrect log message in info_collector::on_app_stat

Smityz:
fix incorrect log message in info_collector::on_app_stat

### DIFF
--- a/src/server/info_collector.cpp
+++ b/src/server/info_collector.cpp
@@ -103,13 +103,13 @@ info_collector::~info_collector()
 
 void info_collector::start()
 {
-    _app_stat_timer_task =
-        ::dsn::tasking::enqueue_timer(LPC_PEGASUS_APP_STAT_TIMER,
-                                      &_tracker,
-                                      [this] { on_app_stat(); },
-                                      std::chrono::seconds(_app_stat_interval_seconds),
-                                      0,
-                                      std::chrono::minutes(1));
+    _app_stat_timer_task = ::dsn::tasking::enqueue_timer(
+        LPC_PEGASUS_APP_STAT_TIMER,
+        &_tracker,
+        [this] { on_app_stat(); },
+        std::chrono::seconds(_app_stat_interval_seconds),
+        0,
+        std::chrono::minutes(1));
 
     _capacity_unit_stat_timer_task = ::dsn::tasking::enqueue_timer(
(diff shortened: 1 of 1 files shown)


acelyc111:
> In the past version, stat table is considered a table, but in some cases, it will cause some misunderstanding. If there we drop the table stat and temp(which are initially generated by default), app_count will be -1.

The description let me puzzled, I think you don't need to explain so much :)

src/server/info_collector.cpp
```diff
@@ -163,7 +163,7 @@ void info_collector::on_app_stat()
     get_app_counters(all_stats.app_name)->set(all_stats);
 
     ddebug("stat apps succeed, app_count = %d, total_read_qps = %.2f, total_write_qps = %.2f",
```
levy5307, line 165: `ddebug_f`

levy5307:
> > In the past version, stat table is considered a table, but in some cases, it will cause some misunderstanding. If there we drop the table stat and temp(which are initially generated by default), app_count will be -1.
> 
> The description let me puzzled, I think you don't need to explain so much :)

I agree :(

src/server/info_collector.cpp
```diff
@@ -162,10 +162,10 @@ void info_collector::on_app_stat()
     }
     get_app_counters(all_stats.app_name)->set(all_stats);
 
-    ddebug("stat apps succeed, app_count = %d, total_read_qps = %.2f, total_write_qps = %.2f",
-           (int)(all_rows.size() - 1),
-           all_stats.get_total_read_qps(),
-           all_stats.get_total_write_qps());
+    ddebug_f("stat apps succeed, app_count = {}, total_read_qps = {}, total_write_qps = {}",
+             (int)(all_rows.size()),
```
neverchanje, line 166:
```suggestion
             all_rows.size(),
```

hycdong:
I have another question, why original log use `all_rows.size() -1` , it is to exclude test table or stat table?

Smityz:
> I have another question, why original log use `all_rows.size() -1` , it is to exclude test table or stat table?

Through the comparison of historical versions, I speculate that it was a mistake made during reconstruction/
https://github.com/XiaoMi/pegasus/blob/v1.11/src/server/info_collector.cpp#L140

PR #561, opened 2020-07-21
feat(bulk-load): add bulk load shell command

hycdong:
This pull request adds and implements bulk load shell command, including: start/pause/restart/cancel bulk load.
This pull request also adds query bulk load, but not implement it in this pull request.

### Check List 

Tests 
- Integration test
- Manual test

### DIFF
--- a/src/shell/commands/bulk_load.cpp
+++ b/src/shell/commands/bulk_load.cpp
@@ -66,8 +66,17 @@ bool start_bulk_load(command_executor *e, shell_context *sc, arguments args)
     return true;
 }
 
-bool pause_bulk_load(command_executor *e, shell_context *sc, arguments args)
+// helper function for pause/restart bulk load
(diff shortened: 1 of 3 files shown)


src/shell/commands/bulk_load.cpp
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    if (app_name.empty()) {
+        fprintf(stderr, "app_name should not be empty\n");
+        return false;
+    }
+
+    auto err_resp = sc->ddl_client->control_bulk_load(
+        app_name, dsn::replication::bulk_load_control_type::BLC_RESTART);
```
acelyc111, line 139: Almost same with pause_bulk_load except `bulk_load_control_type `, you can do some encapsulation and reuse some code.

hycdong: done

PR #566, opened 2020-07-22
perf(slog): add new pool for slog callback

Shuo-Jia:
https://github.com/XiaoMi/rdsn/pull/568 split the slog and plog callback pool, this patch add new pool config to support it.

### Check List 
compile it with https://github.com/XiaoMi/rdsn/pull/568 and deploy it on cluster, then run ycsb

Related changes

### New config
```
[apps.replica]
  pools = ...THREAD_POOL_SLOG

[threadpool.THREAD_POOL_SLOG]
  name = slog
  worker_count = 1
```

### Note
* The config depends on https://github.com/XiaoMi/rdsn/pull/568, otherwise the `start_onebox` will crash:
```
invalid enum configuration '[apps.replica] pools = THREAD_POOL_SLOG'
```
* Now the thread pool is named `THREAD_POOL_SLOG` and set `worker_count=1`, which is because the slog write is sequential in single thread, it won't write next log until the previous log callback complete. So, the one worker_count is enough. However if has other task put the pool, you must increase the worker count value.


src/server/config.ini
```diff
@@ -16,7 +16,7 @@
   name = replica
   arguments =
   ports = 34801
-  pools = THREAD_POOL_DEFAULT,THREAD_POOL_REPLICATION_LONG,THREAD_POOL_REPLICATION,THREAD_POOL_FD,THREAD_POOL_LOCAL_APP,THREAD_POOL_FDS_SERVICE,THREAD_POOL_COMPACT,THREAD_POOL_INGESTION
+  pools = THREAD_POOL_DEFAULT,THREAD_POOL_REPLICATION_LONG,THREAD_POOL_REPLICATION,THREAD_POOL_FD,THREAD_POOL_LOCAL_APP,THREAD_POOL_FDS_SERVICE,THREAD_POOL_COMPACT,THREAD_POOL_INGESTION,THREAD_POOL_SLOG
```
levy5307, line 19: Does the config.ini used for unit test should change synchronously？

src/server/config.min.ini
```diff
@@ -89,6 +89,12 @@
   partitioned = false
   worker_count = 2
 
+[threadpool.THREAD_POOL_SLOG]
+  name = slog
+  partitioned = false
+  worker_priority = THREAD_xPRIORITY_NORMAL
```
neverchanje, line 95:
```suggestion
```
THREAD_xPRIORITY_NORMAL is configured by default. So no need to repeat here.

src/server/config.ini
Shuo-Jia: Yeah, should need

PR #568, opened 2020-07-30
feat(bulk-load): add bulk load function test

hycdong:
This pull request adds function test of bulk load and sst files and other files used for function test.
This test will run at least 5 minutes, so it won't run on travis currently.
When try to run function test including bulk load function test, it will download files from remote repository. If the files exists, it won't download repeatedly.

### Check List 

Tests 

- Unit test
- Integration test


src/test/function_test/test_bulk_load.cpp
```diff
@@ -0,0 +1,254 @@
...
+        for (int i = 0; i < COUNT; ++i) {
+            std::string hash_key = hashkey_prefix + boost::lexical_cast<std::string>(i);
+            for (int j = 0; j < COUNT; ++j) {
+                std::string sort_key = sortkey_prefix + boost::lexical_cast<std::string>(j);
+                std::string act_value;
+                int ret = pg_client->get(hash_key, sort_key, act_value);
+                if (ret != 0) {
+                    std::cout << "Failed to get [" << hash_key << "," << sort_key << "], error is "
+                              << ret << std::endl;
+                    return false;
```
acelyc111, line 146: Once this function return `false`, it will lead caller's `ASSERT_*` fail. So you can use ASSERT_* inside this function like:
```
ASSERT_EQ(ret, ERR_OK) << "Failed to get [" << hash_key << "," << sort_key << "], error is "
                       << ret << std::endl;
```

src/test/function_test/test_bulk_load.cpp
```diff
@@ -0,0 +1,254 @@
...
+                ASSERT_EQ(0, ret);
+            } break;
+            case bulk_load_test::operation::NO_VALUE: {
+                std::string act_value;
+                int ret = pg_client->get(hash_key, sort_key, act_value);
+                // -1001 means value not found
+                ASSERT_EQ(ret, -1001);
+            } break;
+            default:
+                break;
```
acelyc111, line 186:
```suggestion
                ASSERT_TRUE(false);
                break;
```

src/test/function_test/test_bulk_load.cpp
```diff
@@ -0,0 +1,254 @@
...
+/// case3: bulk load data consistent:
+///     - old data will be overrided by bulk load data
+///     - get/set/del succeed after bulk load
+///
+TEST_F(bulk_load_test, bulk_load_tests)
+{
+    // bulk load failed because partition[0] `bulk_load_metadata` file is missing
+    remove_file(bulk_load_local_root + "/" + CLUSTER + "/" + APP_NAME + "/0/bulk_load_metadata");
+    ASSERT_EQ(start_bulk_load(), ERR_OK);
+    // bulk load will get FAILED
```
acelyc111, line 229: How do you know bulk load will get FAILED indeed?

src/test/function_test/test_bulk_load.cpp
```diff
@@ -0,0 +1,254 @@
...
+                          bulk_load_local_root + "/" + CLUSTER + "/" + APP_NAME + "/";
+        system(cmd.c_str());
+    }
+
+    void wait_bulk_load_finish(int64_t seconds)
+    {
+        int64_t sleep_time = 5;
+        error_code err = ERR_OK;
+
+        while (seconds > 0 && err == ERR_OK) {
```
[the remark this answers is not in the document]
hycdong: when bulk load finished, err will be ERR_INVALID_STATE, I adds comment for it.

src/test/function_test/test_bulk_load.cpp
hycdong: I add assert statement in function `verify_bulk_load_data`

src/test/function_test/test_bulk_load.cpp
hycdong: I update the function `wait_bulk_load_finish`, it will return last bulk load status.

PR #575, opened 2020-08-14
fix: replace use of insecure sprintf

l2dy:
Use of `sprintf()` is dangerous and may cause buffer overflows.

Replace `sprintf()` with `snprintf()`, limiting bytes written to the buffer.

### Check List 

Tests 

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes

### DIFF
--- a/src/client_lib/pegasus_client_factory_impl.cpp
+++ b/src/client_lib/pegasus_client_factory_impl.cpp
@@ -23,7 +23,7 @@ bool pegasus_client_factory_impl::initialize(const char *config_file)
             // use config file to run
             char exe[] = "client";
             char config[1024];
-            sprintf(config, "%s", config_file);
+            snprintf(config, 1024, "%s", config_file);
             char *argv[] = {exe, config};
             dsn_run(2, argv, false);
         }
--- a/src/server/info_collector.cpp
+++ b/src/server/info_collector.cpp
@@ -181,8 +181,8 @@ info_collector::app_stat_counters *info_collector::get_app_counters(const std::s
     char counter_desc[1024];
 #define INIT_COUNTER(name)                                                                         \
     do {                                                                                           \
-        sprintf(counter_name, "app.stat." #name "#%s", app_name.c_str());                          \
-        sprintf(counter_desc, "statistic the " #name " of app %s", app_name.c_str());              \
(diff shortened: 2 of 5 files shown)


neverchanje:
@l2dy  Thank you for your contribution!  Some of the `sprintf` usages you fixed here are in our plan to be refactored. As a temporary fix, you can use [fmtlib](https://github.com/fmtlib/fmt) that's been widely used in Pegasus, instead of using the C API `snprintf`.

```cpp
char hash_key_count_str[100];
snprintf(hash_key_count_str, 100, " (%ld hash keys)", cur_total_hash_key_count);
```

```cpp
std::string hash_key_count_str = fmt::format(" ({} hash keys)", cur_total_hash_key_count);
```

l2dy:
`snprintf` is also a temporary fix, if code here is to be refactored.

I'm not familiar with `std::string` and fmtlib, so I'll leave this pull request as is. Buffer overflows should be fixed soon anyway, with fmtlib or not.

src/client_lib/pegasus_client_factory_impl.cpp
```diff
@@ -23,7 +23,7 @@ bool pegasus_client_factory_impl::initialize(const char *config_file)
             // use config file to run
             char exe[] = "client";
             char config[1024];
-            sprintf(config, "%s", config_file);
+            snprintf(config, 1024, "%s", config_file);
```
levy5307, line 26: I think you should define a const variable for the value `1024`

l2dy: That's true, but the size of these buffers are already hard-coded and you have plans to refactor them afterwards with maybe fmtlib, so I don't want to bother to introduce new size constants for each of them. Time would be better spent on refactoring with fmtlib.

P.S. In Pegasus' code base, there are at least 25 `char` arrays with hard-coded size and no comment explaining the choice of different sizes. That should be improved in other PRs.

l2dy:
Closing due to inactivity.

acelyc111:
> Closing due to inactivity.

Hi @l2dy , thanks for your advice, as @neverchanje mentioned above, we are planning to use `fmtlib` to replace C `*printf`. We can discuss for the next refactor and you can contribute if you're interested in it.

PR #576, opened 2020-08-14
fix(script): operation return before the cluster balance when using pegasus_rolling_update.sh 

Shuo-Jia:
The script return when get the `op_count=0` at `rebalance stage` when using the `pegasus_rolling_update.sh ` to update cluster `all` job(meta, replica, collector). However, `op_count` may have not be updated. This pr add the `double-check`which will wait 30s and check again when `op_count=0`. if it still zero, the script judge the cluster is balanced.

Beside, this pr split the `rebalance process` to `pegasus_rebalance_cluster.sh` for `pegasus_rolling_update.sh ` and `pegasus_add_node_list.sh `, user can also uses it directly.

### Note
`pegasus_rolling_update.sh ` one job(replica server) won't execute `rebalance` cluster, you need use `pegasus_rebalance_cluster.sh` to rebalance it after only rolling replica server

* Split the `rebalance process` to `pegasus_rebalance_cluster`
* Double check after wait 30s when `op_count=0`

### Check List 
 (1)  deploy the cluster with minos
 (2)  execute pegasus_rebalance_cluster.sh/pegasus_rolling_update.sh/pegasus_add_node_list.sh

Related changes


scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+#
+# Pegasus cluster rebalance 
+#
+
+PID=$$
+
+if [ $# -le 2 ]; then
+  echo "USAGE: $0 <cluster-name> <cluster-meta-list> <only_move_primary>"
```
acelyc111, line 9: unify to use `-` or `_`

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
...
+
+echo "UID=$UID"
+echo "PID=$PID"
+echo "Start time: `date`"
+rebalance_start_time=$((`date +%s`))
+echo
+
+echo "Generating /tmp/$UID.$PID.pegasus.rebalance.cluster_info..."
+echo cluster_info | ./run.sh shell --cluster $meta_list 2>&1 | sed 's/ *$//' >/tmp/$UID.$PID.pegasus.rebalance.cluster_info
+cname=`grep zookeeper_root /tmp/$UID.$PID.pegasus.rebalance.cluster_info | grep -o '/[^/]*$' | grep -o '[^/]*$'`
```
acelyc111, line 40: It's very hard to read these code, JSON format output is supported, could you plz to use JSON?

acelyc111: We are also plan to refactor remote commands to HTTP APIs, the latter will output in JSON format, we can reuse these JSON parser code then.

Shuo-Jia: Well, now all scripts use the `regex` to filter the result, I think they can be refactored uniformly later

acelyc111: OK, could you please add an issue to trace this plan?

Shuo-Jia: Yeah, I have add it https://github.com/apache/incubator-pegasus/issues/578

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
...
+set_ok=`grep 'control meta level ok' /tmp/$UID.$PID.pegasus.rebalance.set_meta_level | wc -l`
+if [ $set_ok -ne 1 ]; then
+  echo "ERROR: set meta level to lively failed"
+  exit 1
+fi
+
+echo "Wait cluster to become balanced..."
+echo "Wait for 3 minutes to do load balance..."
+sleep 180
+op_count_check_time=1
```
levy5307, line 75: I think `op_count_check_remain_times` is better

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
...
+    op_count=$(echo "cluster_info" | ./run.sh shell --cluster $meta_list | grep balance_operation_count | grep -o 'total=[0-9][0-9]*' | cut -d= -f2)
+    if [ -z "op_count" ]; then
+        break
+    fi
+
+    if [ $op_count -eq 0 -a $op_count_check_time -eq 0 ]; then
+        break
+    fi
+
+    if [ $op_count -eq 0 -a $op_count_check_time -gt 0 ]; then
```
levy5307, line 86: ` if [ $op_count -eq 0]` is enough.  there is no need to add `$op_count_check_time -gt 0`

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
...
+echo "Wait for 3 minutes to do load balance..."
+sleep 180
+op_count_check_remain_times=1
+while true; do
+    op_count=$(echo "cluster_info" | ./run.sh shell --cluster $meta_list | grep balance_operation_count | grep -o 'total=[0-9][0-9]*' | cut -d= -f2)
+    if [ -z "op_count" ]; then
+        break
+    fi
+
+    if [ $op_count -eq 0 -a $op_count_check_remain_times -eq 0 ]; then
```
levy5307, line 82:
```
if [ $op_count -eq 0 -a $op_count_check_remain_times -eq 0 ]; then
        break
    fi

    if [ $op_count -eq 0 ]; then
        if [$op_count_check_remain_times -eq 0 ]; then
              break;
        else
              echo "Cluster may be balanced, try wait 30 seconds..."
             ((op_count_check_remain_times--))
             sleep 30
    else
        echo "Still $op_count balance operations to do..."
        sleep 10
    fi
```

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+#
+# Pegasus cluster rebalance 
+#
+
+PID=$$
+
+if [ $# -le 2 ]; then
+  echo "USAGE: $0 <cluster-name> <cluster-meta-list> <only-move-primary>"
```
neverchanje, line 9:
```suggestion
  echo "USAGE: $0 <cluster-name> <cluster-meta-list> <only-move-primary>"
```

The second argument can be optional, default to false.

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
...
+set_ok=`grep 'control meta level ok' /tmp/$UID.$PID.pegasus.rebalance.set_meta_level | wc -l`
+if [ $set_ok -ne 1 ]; then
+  echo "ERROR: set meta level to lively failed"
+  exit 1
+fi
+
+echo "Wait cluster to become balanced..."
+echo "Wait for 3 minutes to do load balance..."
+sleep 180
+op_count_check_remain_times=1
```
neverchanje, line 75:
```suggestion
## Number of check times for balanced state, in case that op_count is 0 but
## the cluster is in fact unbalanced. Each check waits for 30 secs.
op_count_check_remain_times=1
```

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -0,0 +1,121 @@
...
+while true; do
+    op_count=$(echo "cluster_info" | ./run.sh shell --cluster $meta_list | grep balance_operation_count | grep -o 'total=[0-9][0-9]*' | cut -d= -f2)
+    if [ -z "op_count" ]; then
```
neverchanje, line 78:
```suggestion
    if [ -z "$op_count" ]; then
```

Is this a bug?

Issue #571, opened 2020-08-10
Load options from file when open an exist DB


acelyc111:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Replicas may transient lost from consistent protocol and then reopen on the same ReplicaServer shortly, because of network or some other problems.
When replicas reopen rocksdb, its **usage scenario** will reset to **NORMAL** and set to **actual scenario** of this table after ENVS synced from MetaServer. It will cause flushs and compactions when call rocksdb's `SetOptions()`, we'd better avoid it.

**Describe the feature you'd like:**

Avoid reset usage scenario and DB options, for example:
- store usage scenario into meta column family
- load options from file when open an exist DB (using `rocksdb::LoadLatestOptions`)

PR #587, opened 2020-08-26
fix: Load options from file when open an exist DB

zhangyifan27:

#571 

RocksDB database automatically persisted its current set of options into a file since RocksDB 4.3[1], 
so when DB exists, we should open the rocksdb instance using the options file stored in the db directory.

This patch also adds usage scenario into meta column family to initialize `_usage_scenario`.

[1] https://github.com/facebook/rocksdb/wiki/RocksDB-Options-File

### Check List 

Tests 

- Unit test
- Integration test

Code changes

Side effects

- Increased code complexity

Related changes

### DIFF
--- a/src/server/meta_store.cpp
+++ b/src/server/meta_store.cpp
@@ -64,9 +64,10 @@ std::string meta_store::get_usage_scenario() const
     std::string usage_scenario = ROCKSDB_ENV_USAGE_SCENARIO_NORMAL;
     auto ec = get_string_value_from_meta_cf(false, ROCKSDB_ENV_USAGE_SCENARIO_KEY, &usage_scenario);
     dassert_replica(ec == ::dsn::ERR_OK || ec == ::dsn::ERR_OBJECT_NOT_FOUND,
-                    "rocksdb {} get {} from meta column family failed",
+                    "rocksdb {} get {} from meta column family failed: {}",
                     _db->GetName(),
-                    ROCKSDB_ENV_USAGE_SCENARIO_KEY);
+                    ROCKSDB_ENV_USAGE_SCENARIO_KEY,
+                    ec.to_string());
     return usage_scenario;
 }
 
--- a/src/server/meta_store.cpp
+++ b/src/server/meta_store.cpp
@@ -92,7 +92,6 @@ std::string meta_store::get_usage_scenario() const
     dassert_f(dsn::buf2uint64(data, *value),
               "rocksdb {} get \"{}\" from meta column family failed to parse into uint64",
               db->GetName(),
-              key,
               data);
     return ::dsn::ERR_OK;
 }
--- a/src/server/meta_store.cpp
+++ b/src/server/meta_store.cpp
@@ -52,38 +52,63 @@ uint64_t meta_store::get_last_manual_compact_finish_time() const
 uint64_t meta_store::get_decree_from_readonly_db(rocksdb::DB *db,
                                                  rocksdb::ColumnFamilyHandle *meta_cf) const
 {
+    std::string str_last_flushed_decree;
     uint64_t last_flushed_decree = 0;
-    auto ec = get_value_from_meta_cf(db, meta_cf, true, LAST_FLUSHED_DECREE, &last_flushed_decree);
+    auto ec = get_string_value_from_meta_cf(
+        db, meta_cf, true, LAST_FLUSHED_DECREE, &str_last_flushed_decree);
     dcheck_eq_replica(::dsn::ERR_OK, ec);
+    dassert_f(dsn::buf2uint64(str_last_flushed_decree, last_flushed_decree),
+              "rocksdb {} get LAST_FLUSHED_DECREE from meta column family got error value {}",
+              db->GetName(),
+              str_last_flushed_decree);
     return last_flushed_decree;
+}
+
+std::string meta_store::get_usage_scenario() const
+{
+    std::string usage_scenario;
+    auto ec = get_string_value_from_meta_cf(false, ROCKSDB_ENV_USAGE_SCENARIO_KEY, &usage_scenario);
+    dcheck_eq_replica(::dsn::ERR_OK, ec);
+    return usage_scenario;
 }
 
 ::dsn::error_code meta_store::get_value_from_meta_cf(bool read_flushed_data,
                                                      const std::string &key,
                                                      uint64_t *value) const
 {
-    return get_value_from_meta_cf(_db, _meta_cf, read_flushed_data, key, value);
+    std::string data;
+    auto ec = get_string_value_from_meta_cf(_db, _meta_cf, read_flushed_data, key, &data);
+    if (ec != ::dsn::ERR_OK) {
+        return ec;
+    }
+    dassert_f(dsn::buf2uint64(data, *value),
+              "rocksdb {} get {} from meta column family got error value {}",
+              _db->GetName(),
+              key,
+              data);
+    return ::dsn::ERR_OK;
 }
 
-::dsn::error_code meta_store::get_value_from_meta_cf(rocksdb::DB *db,
-                                                     rocksdb::ColumnFamilyHandle *cf,
-                                                     bool read_flushed_data,
-                                                     const std::string &key,
-                                                     uint64_t *value)
+::dsn::error_code meta_store::get_string_value_from_meta_cf(bool read_flushed_data,
+                                                            const std::string &key,
+                                                            std::string *value) const
 {
-    std::string data;
+    return get_string_value_from_meta_cf(_db, _meta_cf, read_flushed_data, key, value);
+}
+
+::dsn::error_code meta_store::get_string_value_from_meta_cf(rocksdb::DB *db,
+                                                            rocksdb::ColumnFamilyHandle *cf,
+                                                            bool read_flushed_data,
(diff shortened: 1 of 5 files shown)


acelyc111:
> If store usage scenario into meta column family, we need to open DB with default options and then read
meta data and set options, that would also lead to unexpected flushs and compactions.

Sure, usage scenario stored in meta column family is not used to initialize options to open db, but I think it's worthy to initialize member variable `_usage_scenario`, which is not well initialized currently, and it's an empty string after constructing. If restore it from meta CF, we can prevent calling `set_options` which may cause flush and compaction by https://github.com/apache/incubator-pegasus/blob/ea1689f5ca46dbdc27a15054d73e2a9a5ae0a37e/src/server/pegasus_server_impl.cpp#L2453

src/server/meta_store.cpp
```diff
@@ -52,38 +52,64 @@ uint64_t meta_store::get_last_manual_compact_finish_time() const
 uint64_t meta_store::get_decree_from_readonly_db(rocksdb::DB *db,
                                                  rocksdb::ColumnFamilyHandle *meta_cf) const
 {
+    std::string str_last_flushed_decree;
     uint64_t last_flushed_decree = 0;
-    auto ec = get_value_from_meta_cf(db, meta_cf, true, LAST_FLUSHED_DECREE, &last_flushed_decree);
+    auto ec = get_string_value_from_meta_cf(
```
Shuo-Jia, line 57: Why not use `get_value_from_meta_cf` directly?

src/server/pegasus_server_impl.cpp
```diff
@@ -1375,6 +1414,7 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
         _meta_store->set_data_version(PEGASUS_DATA_VERSION_MAX);
         _meta_store->set_last_flushed_decree(0);
         _meta_store->set_last_manual_compact_finish_time(0);
+        _meta_store->set_usage_scenario(ROCKSDB_ENV_USAGE_SCENARIO_NORMAL);
```
Shuo-Jia, line 1417: Defautly, the `get_envs` will show `null` result when using `shell`,  if here `set_usage_scenario(ROCKSDB_ENV_USAGE_SCENARIO_NORMAL)` at  first `start`, does that mean the `get_envs` will show `normal` after reopen when using `shell`?

src/server/meta_store.cpp
zhangyifan27: `get_value_from_meta_cf` reads data from `_db`, but `get_string_value_from_meta_cf` reads data from a readonly db.

src/server/pegasus_server_impl.cpp
```diff
@@ -1360,6 +1398,7 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
     if (db_exist) {
         _last_committed_decree = _meta_store->get_last_flushed_decree();
         _pegasus_data_version = _meta_store->get_data_version();
+        _usage_scenario = _meta_store->get_usage_scenario();
```
hycdong, line 1401: I think it will cause core while rolling_update server. The table created before won't have `usage_scenario` in meta column family, while rolling update, the `db_exist` is true, it will assert on Line72 in function `get_usage_scenario`

src/server/meta_store.cpp
Shuo-Jia: So is `get_value_from_meta_cf` useless？

I see `get_value_from_meta_cf` has been refactor using `get_string_value_from_meta_cf` and the code is almost duplicate with here, maybe should offer  `get_value_from_meta_cf` overload-version supporting pass `db` and `meta_cf`?

src/server/pegasus_server_impl.cpp
Shuo-Jia: Should not, she `set_usage_scenario`  at first open
https://github.com/apache/incubator-pegasus/blob/ec3b23092e4c677ac14a2ab0e7b0c3966dec5a15/src/server/pegasus_server_impl.cpp#L1417

src/server/pegasus_server_impl.cpp
zhangyifan27: It depends on where we read data from when using pegasus_shell `get_app_envs`.
I think we will read data from meta server, so we would still get an empty string.

src/server/pegasus_server_impl.cpp
zhangyifan27: Indeed, maybe `get_usage_scenario()` should return `ROCKSDB_ENV_USAGE_SCENARIO_NORMAL` if we couldn't find usage_scenario in rocksdb meta column family.

hycdong (approved):

src/server/pegasus_server_impl.cpp
```diff
@@ -1375,7 +1414,12 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
         _meta_store->set_data_version(PEGASUS_DATA_VERSION_MAX);
         _meta_store->set_last_flushed_decree(0);
         _meta_store->set_last_manual_compact_finish_time(0);
+        _meta_store->set_usage_scenario(ROCKSDB_ENV_USAGE_SCENARIO_NORMAL);
```
acelyc111, line 1417: We can specify envs when create new table, so actual usage scenario may be included in envs which is passed from meta server, see https://github.com/apache/incubator-pegasus/blob/dea9756663847acb326539f61b80c6c48f92456e/src/server/pegasus_server_impl.cpp#L1256

You can read it from envs.

src/server/meta_store.cpp
```diff
@@ -58,6 +58,18 @@ uint64_t meta_store::get_decree_from_readonly_db(rocksdb::DB *db,
...
 
+std::string meta_store::get_usage_scenario() const
+{
+    // If couldn't find rocksdb usage scenario in meta column family, return normal in default.
+    std::string usage_scenario = ROCKSDB_ENV_USAGE_SCENARIO_NORMAL;
+    auto ec = get_string_value_from_meta_cf(false, ROCKSDB_ENV_USAGE_SCENARIO_KEY, &usage_scenario);
+    dassert_replica(ec == ::dsn::ERR_OK || ec == ::dsn::ERR_OBJECT_NOT_FOUND,
+                    "rocksdb {} get {} from meta column family failed",
+                    _db->GetName(),
+                    ROCKSDB_ENV_USAGE_SCENARIO_KEY);
```
neverchanje, line 69:
```suggestion
    dassert_replica(ec == ::dsn::ERR_OK || ec == ::dsn::ERR_OBJECT_NOT_FOUND,
                    "rocksdb {} get {} from meta column family failed: {}",
                    _db->GetName(),
                    ROCKSDB_ENV_USAGE_SCENARIO_KEY,
                    ec.to_string());
```

src/server/meta_store.cpp
```diff
@@ -72,18 +84,38 @@ ::dsn::error_code meta_store::get_value_from_meta_cf(rocksdb::DB *db,
                                                      uint64_t *value)
 {
     std::string data;
+    auto ec = get_string_value_from_meta_cf(db, cf, read_flushed_data, key, &data);
+    if (ec != ::dsn::ERR_OK) {
+        return ec;
+    }
+    dassert_f(dsn::buf2uint64(data, *value),
+              "rocksdb {} get {} from meta column family got error value {}",
```
neverchanje, line 92:
```suggestion
              "rocksdb {} get \"{}\" from meta column family failed to parse into uint64",
```

I think you cannot simply print a value out, there could be unprintable characters.

src/server/pegasus_server_impl.cpp
```diff
@@ -1326,22 +1327,59 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
...
             return ::dsn::ERR_LOCAL_APP_FAILURE;
         }
+        for (int i = 0; i < loaded_cf_descs.size(); ++i) {
+            if (loaded_cf_descs[i].name == META_COLUMN_FAMILY_NAME) {
+                missing_meta_cf = false;
+            } else if (loaded_cf_descs[i].name == DATA_COLUMN_FAMILY_NAME) {
+                missing_data_cf = false;
+                loaded_data_cf_opts = loaded_cf_descs[i].options;
+            } else {
+                derror_replica("unkown column family name.");
```
neverchanje, line 1357:
```suggestion
                derror_replica("unknown column family name.");
```

src/server/pegasus_server_impl.cpp
```diff
@@ -1326,22 +1327,59 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
...
+                return ::dsn::ERR_LOCAL_APP_FAILURE;
+            }
+        }
+        // When DB exist, meta CF and default CF must be present.
         dassert_replica(!missing_meta_cf, "You must upgrade Pegasus server from 2.0");
+        dassert_replica(!missing_data_cf, "Missing default column family");
```
neverchanje, line 1363:
```suggestion
        // When DB exists, meta CF and data CF must be present.
        dassert_replica(!missing_meta_cf, "You must upgrade Pegasus server from 2.0"
        dassert_replica(!missing_data_cf, "Missing data column family");
```

acelyc111 (approved):

neverchanje (approved):

PR #593, opened 2020-09-03
fix: Use default options to open db when latest option file has incompatible db options

zhangyifan27:
When we reopen an existing db, we will use latest option file to construct options. However,
if this file is an old version file which has incompatible options with official rocksdb, we would
get an error and fail to open db.
https://github.com/apache/incubator-pegasus/blob/ae7caa2599357f83cbbf79c4367a5d0a72f04288/src/server/pegasus_server_impl.cpp#L1341-L1349

If the option file has known incompatible options `pegasus_data_version` and `pegasus_data`,
we skip loading options from file and use default options to open db.
A new version option file will be generated when we succeed to open db, after that we could use
existing option file to construct options. That means the first time we open an old version db, we
must use default options, and that may lead to unexpected flushes and compactions.

### Check List 

Tests 

1. deploy an old version cluster.
2. create some old version replicas.
3. upgrade replica servers to the new version with #587 and this patch.
4. test if new version servers could start apps sccessfully.

Code changes

Side effects

- Increased code complexity

Related changes

#587 

### DIFF
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -1353,7 +1353,10 @@ void pegasus_server_impl::on_clear_scanner(const int64_t &args) { _context_cache
                                                  &loaded_cf_descs,
                                                  /*ignore_unknown_options=*/true);
         if (!status.ok()) {
-            if (status.ToString().find("pegasus") == std::string::npos) {
+            // Here we ignore an invalid argument error related to `pegasus_data_version` and
+            // `pegasus_data` options, which were used in old version rocksdbs (before 2.1.0).
+            if (status.code() != rocksdb::Status::kInvalidArgument ||
+                status.ToString().find("pegasus_data") == std::string::npos) {
                 derror_replica("load latest option file failed: {}.", status.ToString());
                 return ::dsn::ERR_LOCAL_APP_FAILURE;
             }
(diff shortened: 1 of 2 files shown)


acelyc111 (dismissed):

src/server/pegasus_server_impl.cpp
```diff
@@ -1330,9 +1330,18 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
     // Here we create a `tmp_data_cf_opts` because we don't want to modify `_data_cf_opts`, which
     // will be used elsewhere.
     rocksdb::ColumnFamilyOptions tmp_data_cf_opts = _data_cf_opts;
+    bool has_uncompatible_db_options = false;
```
neverchanje, line 1333: `uncompatible` is misspelled. The correct one is `incompatible`.

acelyc111 (approved):

src/server/pegasus_server_impl.cpp
```diff
@@ -1344,28 +1353,28 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
...
-                missing_meta_cf = false;
-            } else if (loaded_cf_descs[i].name == DATA_COLUMN_FAMILY_NAME) {
-                missing_data_cf = false;
-                loaded_data_cf_opts = loaded_cf_descs[i].options;
-            } else {
-                derror_replica("unknown column family name.");
+            if (status.ToString().find("pegasus") == std::string::npos) {
+                derror_replica("load latest option file failed: {}.", status.ToString());
                 return ::dsn::ERR_LOCAL_APP_FAILURE;
             }
```
neverchanje, line 1359:
```suggestion
            if (status.code() != rocksdb::Status::kInvalidArgument || status.ToString().find(" Unrecognized DBOptions") == std::string::npos) {
                derror_replica("load latest option file failed: {}.", status.ToString());
                return ::dsn::ERR_LOCAL_APP_FAILURE;
            }
```

acelyc111 (approved):

hycdong (approved):

PR #594, opened 2020-09-04
fix: fix db is not nullptr when release_db

hycdong:
Pegasus use `release_db` to close rocksdb instance, like the code below:
```
void pegasus_server_impl::release_db() { release_db(_db, {_data_cf, _meta_cf}); }

void pegasus_server_impl::release_db(rocksdb::DB *db,
                                     const std::vector<rocksdb::ColumnFamilyHandle *> &handles)
{
    if (db) {
        for (auto handle : handles) {
            dassert_replica(handle != nullptr, "");
            db->DestroyColumnFamilyHandle(handle);
            handle = nullptr;
        }
        delete db;
        db = nullptr;
    }
}
```
However, the function would close the copy of `_db`, the original `_db` is still not nullptr.

This bug will lead to core when double stop a pegasus instance, this pull request is to fix this problem.

### Check List 

Tests 

- Unit test
- Integration test
- Manual test

Related changes

### DIFF
--- a/.asf.yml
+++ b/.asf.yml
@@ -1,0 +1,41 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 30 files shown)


acelyc111 (dismissed):

zhangyifan27:
Could you also add a test wherein we close db twice to make sure the bug was fixed?

hycdong:
> Could you also add a test wherein we close db twice to make sure the bug was fixed?

Sure, I have added the unit test for it.

zhangyifan27 (approved):

neverchanje (approved):

levy5307 (approved):

PR #598, opened 2020-09-08
refactor: use db_get to implement incr

neverchanje:

We repeat `rocksdb::DB::Get` in every places where a get is performed. The code around a get is repeated too, including retrival of expire_ts and hashkey, sortkey.

To reduce the repetition I added `db_get` previously. In this PR I use this function to read old value during `incr`.

### Check List 

Tests 

- Unit test


src/server/meta_store.h
```diff
@@ -37,6 +37,11 @@ class meta_store : public dsn::replication::replica_base
     void set_last_manual_compact_finish_time(uint64_t last_manual_compact_finish_time) const;
     void set_usage_scenario(const std::string &usage_scenario) const;
 
+    // Keys of meta data wrote into meta column family.
```
acelyc111, line 40: I didn't find you use them, why move them to public scope?

src/server/pegasus_write_service_impl.h
```diff
@@ -223,29 +227,14 @@ class pegasus_write_service::impl : public dsn::replication::replica_base
...
-                else // update.expire_ts_seconds > 0
+                } else { // update.expire_ts_seconds > 0
                     new_expire_ts = update.expire_ts_seconds;
+                }
             }
-        } else if (s.IsNotFound()) {
-            // old value is not found, set to 0 before increment
-            new_value = update.increment;
-            new_expire_ts = update.expire_ts_seconds > 0 ? update.expire_ts_seconds : 0;
-        } else {
```
Shuo-Jia, line 237: why delete the case？

src/server/test/pegasus_write_service_impl_test.cpp
```diff
@@ -140,5 +156,66 @@ TEST_F(pegasus_write_service_impl_test, verify_timetag_compatible_with_version_0
...
+    ASSERT_EQ(resp.new_value, 100);
+
+    db_get(req.key, &get_ctx);
+    ASSERT_TRUE(get_ctx.found);
+}
+
+TEST_F(pegasus_write_service_impl_test, negative_incr_and_zero_incr)
+{
+    dsn::apps::incr_request req;
+    req.increment = -100;
```
levy5307, line 181: there are too many same code, may be you can refactor it, such as using for look to do it.

neverchanje: Take a look at the new code. I've refactored much. I think it is compacted enough.

src/server/pegasus_write_service_impl.h
neverchanje: It was moved here:

```
        db_get_context get_ctx;
        int err = db_get(raw_key, &get_ctx);
        if (err == 0) {
            if (!get_ctx.found) {
                // old value is not found, set to 0 before increment
                new_value = update.increment;
                new_expire_ts = update.expire_ts_seconds > 0 ? update.expire_ts_seconds : 0;
            } else if (get_ctx.expired) {
```

Because NotFound is treated as a success in `db_get`. 0 is returned.

acelyc111 (approved):

src/server/pegasus_write_service_impl.h
```diff
@@ -675,7 +664,7 @@ class pegasus_write_service::impl : public dsn::replication::replica_base
         return status.code();
     }
 
-    // The resulted `expire_ts` is -1 if record is expired.
+    /// \returns 0 if success.
```
hycdong, line 667: I think the comment of this function should be more detailed, such as the expired case, not_found case, these case will also return 0, but in my view, they are not simply success.

hycdong (approved):

acelyc111 (approved):

PR #602, opened 2020-09-16
build: remove restrict to build from non-git-repo

neverchanje:

This PR removes the restriction that we must build under the git repo of Pegasus, thereby users can download the zip/tar and build directly from the source.

There are generally two cases of building:
1. Building from a git-repo 
2. Buiding from a downloaded source.

Every build will generate automatically a file git_commit.h consisting of `PEGASUS_GIT_COMMIT`.
Since we do not want to update git_commit.h file for every single commit, we ignore it in .gitignore. It's fine for the git-repo case.

However, while building from a non-git-repo, it fails because git_commit.h is unable to be generated.

I modified src/build.sh, which generates a PEGASUS_GIT_COMMIT, "non-git-repo" while building non-git-repo.

### Check List 

Tests 

1. No git-repo.

```
> mv .git .git-tmp # delete git repo info
> rm -rf src/include/pegasus/git_commit.h src/builder # delete build cache
> ./run.sh build

# Build perfectly.

> cat src/include/pegasus/git_commit.h                                                                                                                                                                          
#pragma once
#define PEGASUS_GIT_COMMIT "non-git-repo"
```

2. Git Repo.

```
> rm -rf src/include/pegasus/git_commit.h src/builder GIT_COMMIT # delete build cache

> ./run.sh build

# Build perfectly.

> cat src/include/pegasus/git_commit.h
#pragma once
#define PEGASUS_GIT_COMMIT "2034415aa0833c93f44bc4ab9cb57ffe44f61a18"
```


acelyc111:
What’s the benefit of generating a commit Id when build from non-git-repo? When we release a .tar package, the branch/tag is filled in a version file, I think it is enough to describe what version user use.
On the other hand, how this GIT_COMMIT file update if cherry-pick more patches to the release branch?

neverchanje:
`PEGASUS_GIT_COMMIT` is a complement of `PEGASUS_VERSION` (src/include/pegasus/version.h), in my opinion. So I just keep consistent with what it is in non-git repo. But truth is, I agree that `PEGASUS_GIT_COMMIT` has no much use.

`PEGASUS_GIT_COMMIT` is not "generated" from the downloaded source. It exists in the source package as the `GIT_COMMIT` file which is provided by who produced the package.

levy5307 (approved):

Shuo-Jia (approved):

PR #608, opened 2020-09-23
fix: bugfix on incr

neverchanje:

https://github.com/apache/incubator-pegasus/pull/598 introduced a bug, which makes incr continue to proceed even the rocksdb::get failed.

This PR fixes this bug and adds some unit tests using fail-point library to mock rocksdb get failure.

### Check List 

Tests 

- Unit test

### DIFF
--- a/src/server/pegasus_write_service_impl.h
+++ b/src/server/pegasus_write_service_impl.h
@@ -182,58 +182,60 @@ public:
         uint32_t new_expire_ts = 0;
         db_get_context get_ctx;
         int err = db_get(raw_key, &get_ctx);
-        if (err == 0) {
-            if (!get_ctx.found) {
-                // old value is not found, set to 0 before increment
+        if (err != 0) {
+            resp.error = err;
+            return err;
+        }
+        if (!get_ctx.found) {
+            // old value is not found, set to 0 before increment
+            new_value = update.increment;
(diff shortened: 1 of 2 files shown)


src/server/pegasus_write_service_impl.h
```diff
@@ -182,59 +182,61 @@ class pegasus_write_service::impl : public dsn::replication::replica_base
...
-        if (err == 0) {
-            if (!get_ctx.found) {
-                // old value is not found, set to 0 before increment
-                new_value = update.increment;
-                new_expire_ts = update.expire_ts_seconds > 0 ? update.expire_ts_seconds : 0;
-            } else if (get_ctx.expired) {
-                // ttl timeout, set to 0 before increment
-                _pfc_recent_expire_count->increment();
+        if (err != 0) {
+            resp.error = err;
```
Shuo-Jia, line 186: The line seem to be useless?

neverchanje: Certainly useful. It sets response error.

Shuo-Jia: But you `return error` , `resp.error` is not be used.

Shuo-Jia: Well, I konw

Shuo-Jia (approved):

acelyc111 (approved):

PR #615, opened 2020-10-12
feat(hotkey): add an interface of hotkey capture

Smityz:
hotkey capturing is executed in `capacity_unit_calculator`.
Both `capture_raw_key` and `capture_hash_key` have two params: `key` `weight`
`key`: raw_key or hash_key
`weight`: In some scenarios, some keys will be counted several times, we can capture these keys just one time.

### DIFF
--- a/src/server/capacity_unit_calculator.cpp
+++ b/src/server/capacity_unit_calculator.cpp
@@ -138,11 +138,11 @@ void capacity_unit_calculator::add_multi_get_cu(int32_t status,
 
     if (status == rocksdb::Status::kNotFound) {
         add_read_cu(1);
-        _read_hotkey_collector->capture_raw_key(hash_key, key_count);
+        _read_hotkey_collector->capture_hash_key(hash_key, key_count);
         return;
     }
     add_read_cu(data_size);
-    _read_hotkey_collector->capture_raw_key(hash_key, key_count);
+    _read_hotkey_collector->capture_hash_key(hash_key, key_count);
 }
 
 void capacity_unit_calculator::add_scan_cu(int32_t status,
--- a/src/server/capacity_unit_calculator.cpp
+++ b/src/server/capacity_unit_calculator.cpp
@@ -81,26 +81,6 @@ capacity_unit_calculator::capacity_unit_calculator(
         "app.pegasus", name, COUNTER_TYPE_RATE, "statistic the check and mutate bytes");
 }
 
-void capacity_unit_calculator::count_read_data(const dsn::blob &key, key_type type, int64_t size)
-{
-    add_read_cu(size);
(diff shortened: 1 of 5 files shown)


src/server/capacity_unit_calculator.cpp
```diff
@@ -110,9 +123,11 @@ void capacity_unit_calculator::add_multi_get_cu(int32_t status,
 {
     int64_t data_size = 0;
     int64_t multi_get_bytes = 0;
+    int64_t key_count = 0;
```
levy5307, line 126: uint64_t

src/server/capacity_unit_calculator.cpp
```diff
@@ -144,25 +161,28 @@ void capacity_unit_calculator::add_scan_cu(int32_t status,
...
 }
 
-void capacity_unit_calculator::add_ttl_cu(int32_t status)
+void capacity_unit_calculator::add_ttl_cu(int32_t status, const dsn::blob &key)
 {
     if (status != rocksdb::Status::kOk && status != rocksdb::Status::kNotFound) {
         return;
     }
     add_read_cu(1);
+    _read_hotkey_collector->capture_raw_key(key, 1);
```
levy5307, line 185: refactor to move L184~L185 to a function

Smityz: `add_read_cu(1);` and `_read_hotkey_collector->capture_raw_key(key, 1);` have different semantics, they are not suitable for merging
@zhangyifan27

levy5307: I mean you should wrap these two lines of code into a new function, but not merge `_read_hotkey_collector->capture_raw_key(key, 1);` into `add_read_cu(1)`.
Because these two lines are always called in the same time.

zhangyifan27: If these two functions are always called at the same time, you can put them into a new function `collect_hotkey_and_cu`

src/server/capacity_unit_calculator.cpp
levy5307, line 126: int64_t key_count = kvs.size();

src/server/capacity_unit_calculator.cpp
```diff
@@ -256,9 +284,11 @@ void capacity_unit_calculator::add_check_and_mutate_cu(
 {
     int64_t data_size = 0;
     int64_t check_and_mutate_bytes = 0;
+    int64_t key_count = 0;
```
levy5307, line 287: `uint64_t key_count = mutate_list.size();`

hycdong (approved):

levy5307 (approved):

PR #620, opened 2020-10-15
chore: fix stale urls and remove unmaintained files

neverchanje:

Some stale links are fixed as pegasus URL are now at https://github.com/apache/incubator-pegasus.

Docker files are currently maintained at https://github.com/pegasus-kv/pegasus-docker.
The image files are still undocumented though they give an approach to build pegasus simply. I will update those files in the "pegasus-docker" project later on. They are not maintained in the "pegasus" repo anymore.

installation.md and client-development.md are removed. Given that all our docs (RFC is an exception) are maintained at https://github.com/apache/incubator-pegasus-website (source of the truth for docs), there should not be any "docs" remained in the core repo. I think the README playing a guide-map role to the docs is enough.

Notice that removing those files doesn't mean the files are completely deleted. They still remain in the git history. So don't worry they gonna lost.

### Check List 

Tests 

- No code

Related changes


docs/client-development.md
```diff
@@ -1,81 +0,0 @@
-Notices For Client Library Development
```
hycdong, line 1: I agree that client-develpment.md should be removed in pegasus repo, but I suggest adding this document in client repo or official website.

neverchanje: Certainly. But we don't have to wait after this doc added to website to merge this PR. I will do it later this day.

levy5307 (approved):

acelyc111 (approved):

hycdong (approved):

PR #624, opened 2020-10-20
feat(hotkey): capture data part2 - declare coarse collector 

Smityz:
the implement of coarse collector 
## Config changes
```diff
[pegasus.server]
+ coarse_data_variance_threshold = 3
+ data_capture_hash_bucket_num = 37
```


src/server/hotkey_collector.h
```diff
@@ -26,6 +26,12 @@ namespace server {
 
 class internal_collector_base;
 
+struct detect_hotkey_result
+{
+    int coarse_bucket_index;
+    detect_hotkey_result() : coarse_bucket_index(-1) {}
```
acelyc111, line 32:
```suggestion
    int coarse_bucket_index = -1;
```

src/server/hotkey_collector.h
```diff
@@ -75,28 +81,48 @@ class hotkey_collector : public dsn::replication::replica_base
     void analyse_data();
     void handle_rpc(const dsn::replication::detect_hotkey_request &req,
                     /*out*/ dsn::replication::detect_hotkey_response &resp);
+    static int get_bucket_id(dsn::string_view data);
```
levy5307, line 84: you shouldn't make this function to be a member of class `hotkey_collector`

src/server/hotkey_collector.cpp
```diff
@@ -113,5 +144,63 @@ void hotkey_collector::on_stop_detect(dsn::replication::detect_hotkey_response &
...
+    double table_captured_key_sum = 0;
+    int hot_index = 0;
+    int hot_value = 0;
+    for (int i = 0; i < data_size; i++) {
+        table_captured_key_sum += captured_keys[i];
+        if (captured_keys[i] > hot_value) {
+            hot_index = i;
+            hot_value = captured_keys[i];
+        }
+    }
```
levy5307, line 187: add a blank line

src/server/hotkey_collector.h
```diff
@@ -75,28 +81,48 @@ class hotkey_collector : public dsn::replication::replica_base
...
+    virtual void analyse_data(detect_hotkey_result &result) = 0;
 };
 
 // used in hotkey_collector_state::STOPPED and hotkey_collector_state::FINISHED, avoid null pointers
 class hotkey_empty_data_collector : public internal_collector_base
 {
 public:
-    void capture_data(const dsn::blob &hash_key, uint64_t size) {}
-    void analyse_data() {}
+    explicit hotkey_empty_data_collector(replica_base *base) : internal_collector_base(base){};
```
acelyc111, line 108:
```suggestion
    explicit hotkey_empty_data_collector(replica_base *base) : internal_collector_base(base) {}
```

src/server/hotkey_collector.cpp
```diff
@@ -113,5 +144,63 @@ void hotkey_collector::on_stop_detect(dsn::replication::detect_hotkey_response &
...
+    double standard_deviation = 0;
+    for (int i = 0; i < data_size; i++) {
+        if (i != hot_index) {
+            standard_deviation += pow((captured_keys[i] - captured_keys_avg_count), 2);
+        }
+    }
+    standard_deviation = sqrt(standard_deviation / (data_size - 2));
+    double hot_point = (hot_value - captured_keys_avg_count) / standard_deviation;
+    detect_hotkey_result result;
+    hot_point > threshold ? result.coarse_bucket_index = hot_index
```
levy5307, line 200:
```
if (hot_point > threshold) {
     result.coarse_bucket_index = hot_index;
}
```

src/server/hotkey_collector.cpp
```diff
@@ -113,5 +144,63 @@ void hotkey_collector::on_stop_detect(dsn::replication::detect_hotkey_response &
...
+        _hash_buckets[i].store(0);
+    }
+    result = internal_analysis_method(buckets, FLAGS_coarse_data_variance_threshold);
+}
+
+detect_hotkey_result
+hotkey_coarse_data_collector::internal_analysis_method(const std::vector<uint64_t> &captured_keys,
+                                                       int threshold)
+{
+
```
acelyc111, line 174:
```suggestion
```

src/server/hotkey_collector.cpp
```diff
@@ -113,5 +144,63 @@ void hotkey_collector::on_stop_detect(dsn::replication::detect_hotkey_response &
...
+    int data_size = captured_keys.size();
+    dassert(captured_keys.size() > 2, "data_capture_hash_bucket_num is too small");
```
acelyc111, line 176:
```suggestion
    dcheck_gt(captured_keys.size(), 2);
```

src/server/hotkey_collector.h
Smityz: it will be used in `hotkey_coarse_data_collector` and `hotkey_fine_data_collector`
I think it is a public method

acelyc111 (approved):

acelyc111:
> the implement of coarse collector
> 
> ## Config changes
> ```diff
> [pegasus.server]
> + coarse_data_variance_threshold
> + data_capture_hash_bucket_num
> ```
Better to add the default value in PR description.

acelyc111 (dismissed):

src/server/hotkey_collector.cpp
```diff
@@ -113,5 +144,64 @@ void hotkey_collector::on_stop_detect(dsn::replication::detect_hotkey_response &
     ddebug_replica(hint);
 }
 
+hotkey_coarse_data_collector::hotkey_coarse_data_collector(replica_base *base)
+    : internal_collector_base(base), _hash_buckets(FLAGS_data_capture_hash_bucket_num)
+{
+    for (std::atomic<uint64_t> &bucket : _hash_buckets) {
```
levy5307, line 150: `for (auto &bucket: _hash_buckets)`

src/server/hotkey_collector.cpp
```diff
@@ -19,18 +19,30 @@
...
 
+DSN_DEFINE_int32("pegasus.server",
+                 coarse_data_variance_threshold,
+                 3,
+                 "the threshold of variance calculate to find the outliers");
+
+DSN_DEFINE_int32("pegasus.server",
+                 data_capture_hash_bucket_num,
+                 37,
+                 "the number of data capture hash buckets");
```
levy5307, line 38: you can add `DSN_DEFINE_validator` to make sure the `data_capture_hash_bucket_num` must be greater than 0

src/server/hotkey_collector.cpp
```diff
@@ -32,11 +32,26 @@ DSN_DEFINE_int32("pegasus.server",
...
                  "the number of data capture hash buckets");
 
+DSN_DEFINE_validator(data_capture_hash_bucket_num, [](int32_t bucket_num) -> bool {
+    if (bucket_num < 3) {
+        return false;
+    }
+    // data_capture_hash_bucket_num should be a prime number
+    for (int i = 2; i <= bucket_num / i; i++) {
+        if (bucket_num % i == 0)
+            return false;
```
acelyc111, line 50:
```suggestion
        if (bucket_num % i == 0) {
            return false;
        }
```

src/server/hotkey_collector.cpp
```diff
@@ -164,43 +179,49 @@ void hotkey_coarse_data_collector::analyse_data(detect_hotkey_result &result)
         buckets[i] = _hash_buckets[i].load();
         _hash_buckets[i].store(0);
     }
-    result = internal_analysis_method(buckets, FLAGS_coarse_data_variance_threshold);
+    outlier_detection detection(buckets, FLAGS_coarse_data_variance_threshold);
+    int hotindex = -1;
```
acelyc111, line 183:
```suggestion
    int hot_index = -1;
```

src/server/hotkey_collector.cpp
```diff
@@ -19,18 +19,90 @@
...
+    } else {
+        hot_index = -1;
+        return false;
+    }
+}
+
+static int get_bucket_id(dsn::string_view data)
+{
+    size_t hash_value = boost::hash_range(data.begin(), data.end());
+    return static_cast<int>(hash_value % FLAGS_data_capture_hash_bucket_num);
```
acelyc111, line 97: Calculate hash value of every key may be wastly, is it possible to use https://github.com/XiaoMi/rdsn/blob/master/src/dsn.layer2.thrift#L92 directly?

Smityz: In the current design, it can't get this info

acelyc111 (dismissed):

src/server/hotkey_collector.cpp
acelyc111: Remember to improve it https://github.com/apache/incubator-pegasus/issues/628

acelyc111 (dismissed):

src/server/hotkey_collector.cpp
```diff
@@ -19,25 +19,98 @@
...
+
 DSN_DEFINE_int32(
     "pegasus.server",
     max_seconds_to_detect_hotkey,
     150,
     "the max time (in seconds) allowed to capture hotkey, will stop if hotkey's not found");
 
+// 68–95–99.7 rule, same algorithm as hotspot_partition_calculator::stat_histories_analyse
+static bool
+find_outlier_index(const std::vector<uint64_t> &captured_keys, int threshold, int &hot_index)
```
hycdong, line 65: The function parameter `threshold` is not needed, you can use `FLAGS_coarse_data_variance_threshold` directly.

hycdong, line 65: `captured_keys` is a copy of `_hash_bucket`, it seems that you wouldn't update the value of `_hash_buckets`, why would you copy another vector here?

src/server/hotkey_collector.cpp
```diff
@@ -19,25 +19,98 @@
...
+    // TODO: (Tangyanzhao) increase a judgment of table_captured_key_sum
+    double captured_keys_avg_count =
+        (table_captured_key_sum - captured_keys[hot_index]) / (data_size - 1);
+    double standard_deviation = 0;
+    for (int i = 0; i < data_size; i++) {
+        if (i != hot_index) {
+            standard_deviation += pow((captured_keys[i] - captured_keys_avg_count), 2);
+        }
+    }
+    standard_deviation = sqrt(standard_deviation / (data_size - 2));
```
hycdong, line 89: Why use `data_size - 2` rather than `data_size - 1`?

src/server/hotkey_collector.cpp
```diff
@@ -19,25 +19,98 @@
...
+    double hot_point = (hot_value - captured_keys_avg_count) / standard_deviation;
+    detect_hotkey_result result;
```
hycdong, line 91: What is the usage of `result` here? It seems is useless.

src/server/hotkey_collector.cpp
Smityz: It is a public function actually, the `fine_analyse` will use this function also

Smityz: `_hash_bucket` is dynamic changed, because in other threads it is always capturing data.
in other words, it has to keep `captured_keys` const when analyzing, otherwise, the result is incorrect.

src/server/hotkey_collector.cpp
Smityz: image
In the wiki's desc, the denominator should be `N-1`, so except for the data to be tested, it should be `data_size - 2`
but in some implements of this algorithm, it can also be `N`, that is `data_size`

src/server/hotkey_collector.cpp
Smityz: On the one hand, it can pass the `detect_hotkey_result::coarse_bucket_index` to `fine_collector`
on the other hand, it can be a snapshot to validate the final result

hycdong: I don't understand your explanation, `result` is the local variety of this function, your explanation seems to describe `_result` defined in `hotkey_collector`.

src/server/hotkey_collector.cpp
hycdong: Okay, I have an another question now. If `_hash_buckets` will be updated in different threads, why don't you add a lock when read or write it?

src/server/hotkey_collector.cpp
Smityz: I will add more variable in `detect_hotkey_result`

src/server/hotkey_collector.cpp
```diff
@@ -19,25 +19,97 @@
...
 #include <dsn/utility/flags.h>
 
 namespace pegasus {
 namespace server {
 
+DSN_DEFINE_int32("pegasus.server",
+                 coarse_data_variance_threshold,
+                 3,
+                 "the threshold of variance calculate to find the outliers");
+
```
levy5307, line 35: remove the blank line

hycdong (approved):

acelyc111 (approved):

PR #625, opened 2020-10-20
feat(hotkey): collector can be terminated by timeout

Smityz:
We can set config `hotkey_collector_max_work_time` to determine how long will the hotkey detection takes.
If the running time is greater than the `hotkey_collector_max_work_time`, the hotkey collection will be terminated automatically.
## Config changes
```diff
[pegasus.server]
+ max_seconds_to_detect_hotkey = 150
```


src/server/hotkey_collector.cpp
```diff
@@ -105,13 +117,30 @@ void hotkey_collector::on_start_detect(dsn::replication::detect_hotkey_response
...
+void hotkey_collector::terminate_colletor()
+{
+    _state.store(hotkey_collector_state::STOPPED);
+    _internal_collector.reset();
+    _collector_start_time = 0;
+}
+
+void hotkey_collector::terminate_by_timeout()
+{
+    if (dsn_now_s() == 0)
```
acelyc111, line 136: How can it happen?

src/server/hotkey_collector.h
```diff
@@ -79,9 +79,13 @@ class hotkey_collector : public dsn::replication::replica_base
 private:
     void on_start_detect(dsn::replication::detect_hotkey_response &resp);
     void on_stop_detect(dsn::replication::detect_hotkey_response &resp);
+    void terminate_colletor();
+    void terminate_by_timeout();
```
acelyc111, line 83: How about rename to `check_timeout` ?

src/server/hotkey_collector.cpp
```diff
@@ -105,13 +117,30 @@ void hotkey_collector::on_start_detect(dsn::replication::detect_hotkey_response
...
+        return;
```
neverchanje, line 137:
```suggestion
    if (dsn_now_s() == 0)
        return;
```

No such thing that now_s returns 0. What does this supposed to mean?

src/server/hotkey_collector.h
Smityz: It not only checked but also terminate the hotkey_collector

src/server/hotkey_collector.cpp
Smityz: https://github.com/apache/incubator-pegasus/pull/625/files/df197b1e14bc8b60f9ba1d60079005edaab7d166#diff-7d14c5784befda2ae449c643a1ec2dd2b9988db5fc72325f48e855b6a566e02eR40
https://github.com/apache/incubator-pegasus/pull/625/files/df197b1e14bc8b60f9ba1d60079005edaab7d166#diff-7d14c5784befda2ae449c643a1ec2dd2b9988db5fc72325f48e855b6a566e02eR131
In `hotkey_collector_state::STOPPED` and `hotkey_collector_state::STOPPED` it should be 0

src/server/hotkey_collector.cpp
```diff
@@ -21,16 +21,23 @@
...
 #include <dsn/dist/fmt_logging.h>
+#include <dsn/utility/flags.h>
 
 namespace pegasus {
 namespace server {
 
+DSN_DEFINE_int32("pegasus.server",
+                 hotkey_collector_max_work_time,
+                 150,
+                 "the max time allowed to capture hotkey, will stop if hotkey's not found");
```
neverchanje, line 32:
```suggestion
                 "the max time (in seconds) allowed to capture hotkey, will stop if hotkey's not found");
```

src/server/hotkey_collector.cpp
Smityz: It is a mistake

src/server/hotkey_collector.cpp
Smityz: It is a mistake

neverchanje (dismissed):

src/server/hotkey_collector.cpp
```diff
@@ -105,13 +118,30 @@ void hotkey_collector::on_start_detect(dsn::replication::detect_hotkey_response
...
+    _state.store(hotkey_collector_state::STOPPED);
+    _internal_collector.reset();
+    _collector_start_time = 0;
+}
+
+void hotkey_collector::terminate_by_timeout()
+{
+    if (_collector_start_time == 0)
+        return;
+    if (dsn_now_s() - _collector_start_time >= FLAGS_hotkey_collector_max_work_time) {
```
levy5307, line 139: I think ` if (dsn_now_s() >= FLAGS_hotkey_collector_max_work_time) +  _collector_start_time` is better

src/server/hotkey_collector.h
acelyc111: how about `terminate_if_timeout ` ?

src/server/hotkey_collector.cpp
```diff
@@ -65,7 +73,11 @@ void hotkey_collector::capture_hash_key(const dsn::blob &hash_key, int64_t weigh
     _internal_collector->capture_data(hash_key, weight);
 }
 
-void hotkey_collector::analyse_data() { _internal_collector->analyse_data(); }
+void hotkey_collector::analyse_data()
+{
+    terminate_by_timeout();
+    _internal_collector->analyse_data();
```
acelyc111, line 79: How to comprehend the logic when collector is terminated but still execute analyse_data? I think you can add a boolean return value for `terminate_by_timeout `, and judge it at first.

src/server/hotkey_collector.cpp
```diff
@@ -21,16 +21,24 @@
...
 #include "base/pegasus_key_schema.h"
 #include <dsn/dist/fmt_logging.h>
+#include <dsn/utility/flags.h>
 
 namespace pegasus {
 namespace server {
 
+DSN_DEFINE_int32(
+    "pegasus.server",
+    hotkey_collector_max_work_time,
```
acelyc111, line 31: How about `max_seconds_to_detect_hotkey` or something like that, unit should be present for admins.

src/server/hotkey_collector.cpp
```diff
@@ -105,13 +118,31 @@ void hotkey_collector::on_start_detect(dsn::replication::detect_hotkey_response
...
-    _state.store(hotkey_collector_state::STOPPED);
-    _internal_collector.reset();
+    terminate_colletor();
     resp.err = dsn::ERR_OK;
     std::string hint =
         fmt::format("{} hotkey stopped, cache cleared", dsn::enum_to_string(_hotkey_type));
     ddebug_replica(hint);
 }
 
+void hotkey_collector::terminate_colletor()
```
levy5307, line 128: I think `terminate` is OK, no need to add `_collector`

levy5307 (approved):

src/server/hotkey_collector.cpp
```diff
@@ -138,17 +139,15 @@ void hotkey_collector::terminate()
...
-        return;
-    }
     if (dsn_now_s() >= _collector_start_time + FLAGS_max_seconds_to_detect_hotkey) {
         ddebug_replica("hotkey collector work time is exhausted but no hotkey has been found");
         terminate();
-        return;
+        return false;
     }
-};
+    return true;
```
acelyc111: I think it's more reasonable to return true when terminated.

src/server/hotkey_collector.h
```diff
@@ -79,9 +79,13 @@ class hotkey_collector : public dsn::replication::replica_base
 private:
     void on_start_detect(dsn::replication::detect_hotkey_response &resp);
     void on_stop_detect(dsn::replication::detect_hotkey_response &resp);
+    void terminate();
+    bool terminate_if_timeout();
+
     std::atomic<hotkey_collector_state> _state;
     const dsn::replication::hotkey_type::type _hotkey_type;
     std::shared_ptr<internal_collector_base> _internal_collector;
+    uint64_t _collector_start_time;
```
hycdong, line 88: start_time is second? If is, I suggest rename it `_collector_start_time_second`

src/server/hotkey_collector.cpp
```diff
@@ -88,6 +107,7 @@ void hotkey_collector::on_start_detect(dsn::replication::detect_hotkey_response
         dwarn_replica(hint);
         return;
     case hotkey_collector_state::STOPPED:
+        _collector_start_time = dsn_now_s();
```
hycdong, line 110: Why update the start time in stopped state?

Smityz: because in these codes, the state will be transformed from `hotkey_collector_state::STOPPED` to `hotkey_collector_state::COARSE_DETECTING` which means the collecting action is beginning.

hycdong (approved):

acelyc111 (approved):

PR #632, opened 2020-10-28
fix(hotkey): fix two hidden dangers in hotkey_collector

Smityz:
This PR is to solve two hidden dangers in hotkey_collector.
Because the related codes of two dangers are overlapped, I solve them in one PR.
The first hidden danger is if `FLAGS_xxx` variable is dynamically changing in the future, `FLAGS_data_capture_hash_bucket_num` might be changing when capturing data, which will cause the memory overflow.
The second hidden danger is `_internal_collector` might be reset while other threads are using it.


src/server/hotkey_collector.cpp
```diff
@@ -162,6 +162,14 @@ void hotkey_collector::analyse_data()
     }
 }
 
+inline void hotkey_collector::init_internal_collector()
```
levy5307, line 165: I think `reset_internal_collector` is better

hycdong:
> The second hidden danger is _internal_collector might be reset while other threads are using it.
I don't understand what is the hidden danger of it, and how you fix it, maybe you can add more description about it.

src/server/hotkey_collector.cpp
```diff
@@ -142,17 +139,19 @@ void hotkey_collector::capture_raw_key(const dsn::blob &raw_key, int64_t weight)
 void hotkey_collector::capture_hash_key(const dsn::blob &hash_key, int64_t weight)
 {
     // TODO: (Tangyanzhao) add a unit test to ensure data integrity
-    _internal_collector->capture_data(hash_key, weight);
+    if (_state.load() == hotkey_collector_state::COARSE_DETECTING &&
```
Smityz, line 142: `load()`  will lose performance, find a way to delete it

src/server/hotkey_collector.cpp
```diff
@@ -102,20 +101,78 @@ find_outlier_index(const std::vector<uint64_t> &captured_keys, int threshold, in
...
-      _collector_start_time_second(0)
+    : replica_base(r_base), _hotkey_type(hotkey_type)
 {
+    int now_hash_bucket_num = FLAGS_hotkey_buckets_num;
+    _internal_coarse_collector =
+        std::make_shared<hotkey_coarse_data_collector>(this, now_hash_bucket_num);
+    _internal_fine_collector =
+        std::make_shared<hotkey_fine_data_collector>(this, now_hash_bucket_num);
+    _internal_empty_collector = std::make_shared<hotkey_empty_data_collector>(this);
+    hotkey_collector::change_state_to_stopped();
```
Shuo-Jia, line 119: is the line necessary? At the beginning of the `hotkey_collector`, the all varible should have been default value.

acelyc111 (dismissed):

src/server/hotkey_collector.cpp
```diff
@@ -145,19 +202,24 @@ void hotkey_collector::capture_raw_key(const dsn::blob &raw_key, int64_t weight)
 void hotkey_collector::capture_hash_key(const dsn::blob &hash_key, int64_t weight)
 {
     // TODO: (Tangyanzhao) add a unit test to ensure data integrity
-    _internal_collector->capture_data(hash_key, weight);
+    switch (_state.load()) {
+    case hotkey_collector_state::COARSE_DETECTING:
+    case hotkey_collector_state::FINE_DETECTING:
+        _internal_coarse_collector->capture_data(hash_key, weight > 0 ? weight : 1);
```
acelyc111, line 208: Why the two states both use coarse collector?

Smityz: fiexd

acelyc111 (approved):

levy5307 (approved):

PR #640, opened 2020-11-07
feat(hotkey): add thread_safe method to clear collector

Smityz:


src/server/hotkey_collector.cpp
```diff
@@ -332,5 +339,12 @@ void hotkey_fine_data_collector::analyse_data(detect_hotkey_result &result)
     }
 }
 
+void hotkey_fine_data_collector::clear()
+{
+    std::pair<dsn::blob, uint64_t> key_weight_pair;
+    while (_capture_key_queue.try_dequeue(key_weight_pair)) {
+    };
```
acelyc111, line 346:
```suggestion
    }
```

src/server/hotkey_collector.cpp
```diff
@@ -332,5 +339,12 @@ void hotkey_fine_data_collector::analyse_data(detect_hotkey_result &result)
     }
 }
 
+void hotkey_fine_data_collector::clear()
+{
+    std::pair<dsn::blob, uint64_t> key_weight_pair;
```
levy5307, line 344: what about change it like this?
```
    moodycamel::ConcurrentQueue<std::pair<dsn::blob, uint64_t>> temp;
    _capture_key_queue.swap(temp);
```

Smityz:
```
concurrentqueue.h:890
	// Swaps this queue's state with the other's. Not thread-safe.
	// Swapping two queues does not invalidate their tokens, however
	// the tokens that were created for one queue must be used with
	// only the swapped queue (i.e. the tokens are tied to the
	// queue's movable state, not the object itself).
```

src/server/hotkey_collector.h
```diff
@@ -102,6 +102,7 @@ class internal_collector_base : public dsn::replication::replica_base
     explicit internal_collector_base(replica_base *base) : replica_base(base){};
     virtual void capture_data(const dsn::blob &hash_key, uint64_t weight) = 0;
     virtual void analyse_data(detect_hotkey_result &result) = 0;
+    virtual void clear() = 0;
```
acelyc111, line 105: Where you call this function?

Smityz: https://github.com/apache/incubator-pegasus/pull/632/files#diff-7d14c5784befda2ae449c643a1ec2dd2b9988db5fc72325f48e855b6a566e02eR126
In another PR

acelyc111 (approved):

neverchanje (approved):

PR #646, opened 2020-11-23
feat: support costing memtable and index and filter blocks to block c…

empiredan:

While multiple services are deployed on a node, memory usage of each service has to be restricted. As for pegasus, memory is consumed by both rDSN and rocksdb. We first restrict rocksdb's memory since it's easier to implement.

This PR involves some rocksdb's options, to manage memtable, block cache, index and filter blocks into block cache size, by which we can take precise control over the memory used by rocksdb.

If `rocksdb_enable_write_buffer_manager = true`, memtable will be managed into block cache size, and `rocksdb_block_cache_capacity` will be considered as the total memory assigned to rocksdb,  `rocksdb_total_size_across_write_buffer` is the maximum memory assigned to memtables. 

If `rocksdb_cache_index_and_filter_blocks = true`, index and filter blocks will be stored in block cache, and meanwhile you'd better use partitioned filters; If `rocksdb_cache_index_and_filter_blocks = false`, you have to use `rocksdb_max_open_files` to take control.

As for details of both write buffer manager and index and filter blocks, see the docs as below: 
https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB
https://github.com/facebook/rocksdb/wiki/Write-Buffer-Manager
https://github.com/facebook/rocksdb/wiki/Partitioned-Index-Filters

### Check List 

Tests 

- Unit test
- Integration test
- No code

Code changes

- Has persistent data change

Side effects

- Increased code complexity

Related changes


src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,130 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+         rocksdb::BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch},
+        {"kBinarySearchWithFirstKey",
+         rocksdb::BlockBasedTableOptions::IndexType::kBinarySearchWithFirstKey}
+    };
+    auto index_type_item = index_type_string_map.find(index_type);
+    dassert(index_type_item != index_type_string_map.end(),
+            "[pegasus.server]rocksdb_index_type shold be one among kBinarySearch, "
+            "kHashSearch, kTwoLevelIndexSearch or kBinarySearchWithFirstKey.");
+    tbl_opts.index_type = index_type_item->second;
+    ddebug("rocksdb_index_type = %s", index_type.c_str());
```
neverchanje, line 359: This is irrelevant to the topic of this PR. But I do curious if you test the performance effect of this configuration.

I'd suggest this configuration naming to:

```
[pegasus.server]
rocksdb_index_type = binary_search | hash_search | two_level_index_search | binary_search_with_first_key
```

without prefix "k".

neverchanje:
>  If `rocksdb_cache_index_and_filter_blocks = false`, you have to use `rocksdb_max_open_files` to take control.

Why is the two configs relevant?

neverchanje (changes requested):

src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,130 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+        _db_opts.write_buffer_manager = _s_write_buffer_manager;
+    }
+
+    int64_t max_open_files = dsn_config_get_value_int64(
+            "pegasus.server",
+            "rocksdb_max_open_files",
+            -1,
+            "number of open files that can be used by a replica(namely a DB instance)");
+    _db_opts.max_open_files = static_cast<int>(max_open_files);
+    ddebug("rocksdb_max_open_files = %d", _db_opts.max_open_files);
```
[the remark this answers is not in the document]
empiredan: That's right, keeping per-replica  (i.e. a rocksdb instance) size not very large is the best practice. However, our case is somewhat complex. We deploy the services on our customer's private cluster for security. Some customer's configuration, for example, say max-open-files per process, is small (e.g. 65536). A case that once we experienced is making import faster, we set bulk-load mode, but it was failed at last due to too-many-open-files.

src/server/pegasus_server_impl_init.cpp
empiredan: Actually, since I'd decided to cache index and filter blocks, I just took this opportunity to involve partitioned indexes in. However we've been developing pegasus based on v1.12.3, whose rocksdb's version is 5.9.2 and not high enough to support index and filter blocks cached well. After several related core-dumps had happened we had to disable all the options about the cache and partitioned indexes. Until I saw the version of rocksdb used by latest pegasus had reached 6.6.4, I thought those options could be enabled again albeit disabled still by default. Unfortunately we've not done any performance test since.  I think these options should be enabled by users if necessary

src/server/pegasus_server_impl_init.cpp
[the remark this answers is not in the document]
neverchanje: OK, sounds very reasonable.

empiredan:
> > If `rocksdb_cache_index_and_filter_blocks = false`, you have to use `rocksdb_max_open_files` to take control.
> 
> Why is the two configs relevant?

If `rocksdb_cache_index_and_filter_blocks=false`, table reader will pre-load index and filter blocks from files, whose number can be limited by `rocksdb_max_open_files`, to control the memory consumption.

src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,130 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+                "rocksdb_enable_write_buffer_manager",
+                false,
+                "enable write buffer manager to limit total memory "
+                "used by memtables and block caches across multiple replicas");
+    ddebug("rocksdb_enable_write_buffer_manager = %d", enable_write_buffer_manager);
+    if (enable_write_buffer_manager) {
+        // If write buffer manager is enabled, all replicas(one DB instance for each
+        // replica) on this server will share the same write buffer manager object,
+        // thus the same block cache object. It's convenient to control the total memory
+        // of memtables and bloack caches used by this server
```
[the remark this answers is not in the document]
empiredan: Ok, fixed~

src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,128 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+                static_cast<size_t>(total_size_across_write_buffer), tbl_opts.block_cache);
+        });
+        _db_opts.write_buffer_manager = _s_write_buffer_manager;
+    }
+
+    int64_t max_open_files = dsn_config_get_value_int64(
+        "pegasus.server",
+        "rocksdb_max_open_files",
+        -1,
+        "number of open files that can be used by a replica(namely a DB instance)");
```
levy5307, line 337: number of opened files...

src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,128 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+    _db_opts.max_open_files = static_cast<int>(max_open_files);
+    ddebug("rocksdb_max_open_files = %d", _db_opts.max_open_files);
+
+    std::string index_type =
+        dsn_config_get_value_string("pegasus.server",
+                                    "rocksdb_index_type",
+                                    "binary_search",
+                                    "The index type that will be used for this table.");
+    const std::unordered_map<std::string, rocksdb::BlockBasedTableOptions::IndexType>
```
levy5307, line 346: maybe you can make index_type_string_map as a global static variable

src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,128 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+            if (total_size_across_write_buffer < 0) {
+                total_size_across_write_buffer = 0;
+            }
+            _s_write_buffer_manager = std::make_shared<rocksdb::WriteBufferManager>(
+                static_cast<size_t>(total_size_across_write_buffer), tbl_opts.block_cache);
+        });
+        _db_opts.write_buffer_manager = _s_write_buffer_manager;
+    }
+
+    int64_t max_open_files = dsn_config_get_value_int64(
```
levy5307, line 333: uint64_t

src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,128 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+            int64_t total_size_across_write_buffer = dsn_config_get_value_int64(
+                "pegasus.server",
+                "rocksdb_total_size_across_write_buffer",
+                0,
+                "total size limit used by memtables across multiple replicas");
+            ddebug("rocksdb_total_size_across_write_buffer = %" PRId64,
+                   total_size_across_write_buffer);
+            if (total_size_across_write_buffer < 0) {
+                total_size_across_write_buffer = 0;
+            }
```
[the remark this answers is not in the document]
levy5307: use uint64_t, so you can remove this check

src/server/pegasus_server_impl.cpp
```diff
@@ -68,6 +68,7 @@ static bool chkpt_init_from_dir(const char *name, int64_t &decree)
 std::shared_ptr<rocksdb::RateLimiter> pegasus_server_impl::_s_rate_limiter;
 int64_t pegasus_server_impl::_rocksdb_limiter_last_total_through;
 std::shared_ptr<rocksdb::Cache> pegasus_server_impl::_s_block_cache;
+std::shared_ptr<rocksdb::WriteBufferManager> pegasus_server_impl::_s_write_buffer_manager;
```
levy5307, line 71: what about unique_ptr?

src/server/pegasus_server_impl_init.cpp
```diff
@@ -299,6 +300,128 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
         _db_opts.rate_limiter = _s_rate_limiter;
     }
 
+    bool enable_write_buffer_manager =
+        dsn_config_get_value_bool("pegasus.server",
+                                  "rocksdb_enable_write_buffer_manager",
+                                  false,
+                                  "enable write buffer manager to limit total memory "
+                                  "used by memtables and block caches across multiple replicas");
+    ddebug("rocksdb_enable_write_buffer_manager = %d", enable_write_buffer_manager);
```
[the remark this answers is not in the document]
empiredan: Ok, I see. I've replaced all ddebug.

src/server/pegasus_server_impl_init.cpp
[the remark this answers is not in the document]
empiredan: Ok, I use uint64_t instead. BTW, after I set total_size_across_write_buffer with a negative, I tried to start pegasus_server and I found total_size_across_write_buffer=0, is this as expected ?

src/server/pegasus_server_impl_init.cpp
empiredan: Actually max_open_files is defined as int type in include/rocksdb/options.h, and -1 means files opened are always kept open. So here it has to be signed.

src/server/pegasus_server_impl.cpp
empiredan: In include/rocksdb/options.h, it is defined as std::shared_ptr<WriteBufferManager> as a member of rocksdb::DBOptions, so we'd better keep shared_ptr here ?

src/server/pegasus_server_impl_init.cpp
neverchanje: Agree with @empiredan

neverchanje (approved):

src/server/pegasus_server_impl_init.cpp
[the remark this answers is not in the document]
levy5307: Yes.

levy5307 (approved):

PR #648, opened 2020-11-27
feat(hotkey): add unit test of hotkey collector

Smityz:


src/server/test/hotkey_collector_test.cpp
```diff
@@ -195,5 +198,174 @@ TEST_F(fine_collector_test, fine_collector)
...
+    }
+    collector->analyse_data();
+    ASSERT_EQ(get_collector_stat(collector), hotkey_collector_state::COARSE_DETECTING);
+
+    on_detect_hotkey(generate_control_rpc(dsn::replication::hotkey_type::READ,
+                                          dsn::replication::detect_action::STOP),
+                     resp);
+    ASSERT_EQ(resp.err, dsn::ERR_OK);
+    ASSERT_EQ(get_collector_stat(collector), hotkey_collector_state::STOPPED);
+    _tracker.wait_outstanding_tasks();
```
neverchanje, line 332: This entire block of tests are redundant. It repeates the previous lines, from STOPPED to COARSE_DETECTING, and STOPPED. You don't have to repeat.

src/server/test/hotkey_collector_test.cpp
```diff
@@ -195,5 +198,174 @@ TEST_F(fine_collector_test, fine_collector)
...
+    on_detect_hotkey(generate_control_rpc(dsn::replication::hotkey_type::READ,
+                                          dsn::replication::detect_action::START),
+                     resp);
+    ASSERT_EQ(resp.err, dsn::ERR_OK);
+    on_detect_hotkey(generate_control_rpc(dsn::replication::hotkey_type::WRITE,
+                                          dsn::replication::detect_action::START),
+                     resp);
+    ASSERT_EQ(resp.err, dsn::ERR_OK);
+
+    auto writes = new dsn::message_ex *[1000];
```
neverchanje, line 347: Use message_ptr here. This line has definitely memory leak.

Smityz: `auto cleanup = dsn::defer([=]() { delete[] writes; });`
I defer deleting this object, is it not correct?

src/server/test/hotkey_collector_test.cpp
Smityz: https://github.com/apache/incubator-pegasus/pull/648/files/b7927974330ad1b93d5f28cd2b798d2e7ce2d244#diff-d16dd94725655577d994a6cb037e9e369e6ddcd03f4c5fdbe6dfd99d69362236R321
this block is to test the state transform in normal data
https://github.com/apache/incubator-pegasus/pull/648/files/b7927974330ad1b93d5f28cd2b798d2e7ce2d244#diff-d16dd94725655577d994a6cb037e9e369e6ddcd03f4c5fdbe6dfd99d69362236R287
And the previous block is to test it in hotspot data

src/server/test/hotkey_collector_test.cpp
neverchanje: `delete[] writes` only deletes the array, but not deletes the elements. Use STL& smart ptr whenever possible.

src/server/test/pegasus_server_write_test.cpp
```diff
@@ -67,7 +67,12 @@ class pegasus_server_write_test : public pegasus_server_test_base
                 for (int i = put_rpc_cnt; i < total_rpc_cnt; i++) {
                     writes[i] = pegasus::create_remove_request(key);
                 }
-                auto cleanup = dsn::defer([=]() { delete[] writes; });
+                auto cleanup = dsn::defer([=]() {
```
levy5307, line 70: `for (auto &write : writes)`

Smityz: `error: ‘begin’ was not declared in this scope`
It is inconvenient to use iter in pointer

neverchanje (approved):

levy5307 (approved):

PR #651, opened 2020-12-03
refactor:  use rocksdb_wrapper::get to reimplement incr

levy5307:
1. Add a  class `rocksdb_wrapper` to wrap the operation of rocksdb, and put `db_get` into it. 
2. Use `rocksdb_wrapper::get` to reimplement incr function.

### Check List 

Tests 

- Unit test


neverchanje (approved):

src/server/rocksdb_wrapper.cpp
```diff
@@ -0,0 +1,68 @@
...
+    rocksdb::Status s = _db->Get(_rd_opts, utils::to_rocksdb_slice(raw_key), &(ctx->raw_value));
+    if (dsn_likely(s.ok())) {
+        // success
+        ctx->found = true;
+        ctx->expire_ts = pegasus_extract_expire_ts(_pegasus_data_version, ctx->raw_value);
+        if (check_if_ts_expired(utils::epoch_now(), ctx->expire_ts)) {
+            ctx->expired = true;
+            _pfc_recent_expire_count->increment();
+        }
+        return 0;
```
acelyc111, line 50: When do refactor, ERR_OK is more meanful than '0'

src/server/rocksdb_wrapper.h
```diff
@@ -0,0 +1,58 @@
...
+namespace dsn {
+class perf_counter_wrapper;
+} // namespace dsn
+
+namespace pegasus {
+namespace server {
+struct db_get_context;
+class pegasus_server_impl;
+
+static constexpr int FAIL_DB_GET = -104;
```
acelyc111, line 38: Could you define it in cpp?
And what does '-104' mean?

levy5307: It it internal error codes used for fail injection, and there are three other error codes defined in `pegasus_write_service_impl.h`:
```
static constexpr int FAIL_DB_WRITE_BATCH_PUT = -101;
static constexpr int FAIL_DB_WRITE_BATCH_DELETE = -102;
static constexpr int FAIL_DB_WRITE = -103;
```
I think `104` has no special meaning. And I don't want to change it, because it is not the main purpose for this pull reqeust.

acelyc111 (dismissed):

neverchanje (approved):

acelyc111 (approved):

PR #661, opened 2020-12-17
fix: add thirdparty lib when pack server

hycdong:
This pull request adds `libsasl2.so.3` and `libcom_err.so.3` when pack server.


acelyc111:
What problem you encountered?

levy5307 (approved):

hycdong:
> What problem you encountered?

Missing those shared libraries, replica server won't be started when deploying cluster, I will close this pr, liwei will raise another pr to resolve it.

PR #684, opened 2021-01-18
fix(split): add flush_memtable for copy_checkpoint_to_dir

hycdong:
As rdsn pr [729](https://github.com/XiaoMi/rdsn/pull/729) shows, this pr add an option flush_memtable for function `copy_checkpoint_to_dir`. If `flush_memtable=true`, it will flush memtable before create a checkpoint, otherwise, not flush memtable.

### DIFF
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -1802,19 +1802,21 @@ private:
 
 // Must be thread safe.
 ::dsn::error_code pegasus_server_impl::copy_checkpoint_to_dir(const char *checkpoint_dir,
-                                                              /*output*/ int64_t *last_decree)
+                                                              /*output*/ int64_t *last_decree,
+                                                              bool flush_memtable)
 {
     CheckpointingTokenHelper token_helper(_is_checkpointing);
     if (!token_helper.token_got()) {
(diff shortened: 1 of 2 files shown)


neverchanje (approved):

src/server/pegasus_server_impl.cpp
```diff
@@ -1802,19 +1802,21 @@ ::dsn::error_code pegasus_server_impl::async_checkpoint(bool flush_memtable)
 
 // Must be thread safe.
 ::dsn::error_code pegasus_server_impl::copy_checkpoint_to_dir(const char *checkpoint_dir,
-                                                              /*output*/ int64_t *last_decree)
+                                                              /*output*/ int64_t *last_decree,
+                                                              bool flush_memtable)
```
neverchanje, line 1806: Usually argument is before output variables. `copy_checkpoint_to_dir(const char *checkpoint_dir, const bool flush_memtable, /*output*/ int64_t last_decree)`

hycdong: It will update all functions who call `copy_checkpoint_to_dir`, and its default value is `false`, so I just add it as the last parameter.

acelyc111 (approved):

levy5307 (approved):

PR #685, opened 2021-01-20
fix(asan): repeated memory release in pegasus_server_write_test.batch_writes

levy5307:
writes[0] ~ writes[total_rpc_cnt-1] will be destructed by their corresponding rpc_holder which created in on_batched_write_requests, so we don't need to release them in this unit test. Or it will produce coredump.

### Check List 

Tests 

- No code

### DIFF
--- a/src/server/test/pegasus_server_write_test.cpp
+++ b/src/server/test/pegasus_server_write_test.cpp
@@ -60,19 +60,17 @@ public:
                 int put_rpc_cnt = dsn::rand::next_u32(1, 10);
                 int remove_rpc_cnt = dsn::rand::next_u32(1, 10);
                 int total_rpc_cnt = put_rpc_cnt + remove_rpc_cnt;
-                auto writes = new dsn::message_ex *[total_rpc_cnt];
+                /**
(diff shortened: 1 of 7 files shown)


src/server/test/pegasus_server_write_test.cpp
```diff
@@ -60,19 +60,18 @@ class pegasus_server_write_test : public pegasus_server_test_base
...
+                 * release them here
+                 **/
+                dsn::message_ex *writes[total_rpc_cnt];
                 for (int i = 0; i < put_rpc_cnt; i++) {
                     writes[i] = pegasus::create_put_request(req);
                 }
                 for (int i = put_rpc_cnt; i < total_rpc_cnt; i++) {
                     writes[i] = pegasus::create_remove_request(key);
                 }
-                auto cleanup = dsn::defer([=]() {
```
Smityz, line 70: https://github.com/apache/incubator-pegasus/blob/51578db98da9f1c7278148d6df08b5d4c220b553/src/server/test/hotkey_collector_test.cpp#L353
May have the same problem

levy5307: Done

hycdong (approved):

neverchanje (approved):

PR #689, opened 2021-01-30
feat: limit rocksdb info logs

neverchanje:

The rocksdb info logs could occupy much disk capacity especially when you have many nearly-empty replicas assigned on the server. In this case, most replica will have rocksdb info logs many times larger than their actual data.

I use rocksdb options for rolling logs to limit the total size.

Related changes


acelyc111 (approved):

src/server/pegasus_server_impl_init.cpp
```diff
@@ -348,6 +348,22 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
...
+        "pegasus.server", "rocksdb_keep_log_file_num", 10, "Maximal RocksDB log files to be kept.");
+    _db_opts.log_file_time_to_roll =
+        dsn_config_get_value_int64("pegasus.server",
+                                   "rocksdb_log_file_time_to_roll",
+                                   24 * 3600, /*86400*/
+                                   "Time for the info log file to roll (in seconds).");
+    _db_opts.max_log_file_size =
+        dsn_config_get_value_int64("pegasus.server",
+                                   "rocksdb_max_log_file_size",
+                                   32 * 1024 * 1024, /*33554432*/
```
acelyc111, line 364: No need to leave comment, it's less readable than the code.

src/server/pegasus_server_impl_init.cpp
```diff
@@ -348,6 +348,22 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
     _db_opts.max_open_files = static_cast<int>(max_open_files);
     ddebug_replica("rocksdb_max_open_files = {}", _db_opts.max_open_files);
 
+    // Configurations of rocksdb info-log.
+    // By default, we keep 10 days of info-logs at maximum, combined with the total size limit
+    // of 10*32MB = 320MB.
+    _db_opts.keep_log_file_num = dsn_config_get_value_int64(
```
levy5307, line 354: DSN_DEFINE_***

src/server/pegasus_server_impl_init.cpp
neverchanje: We in configuration need the finally calculated value. So I want to make people clearly get the number.

acelyc111 (approved):

PR #698, opened 2021-03-12
feat(split): gc useless data after partition split

hycdong:
If app's `validate_partition_hash=true`, it will check partition_hash when execute rocksdb compaction, to gc old data during partition split.

### DIFF
--- a/src/base/pegasus_const.h
+++ b/src/base/pegasus_const.h
@@ -62,4 +62,6 @@ extern const std::string PEGASUS_CLUSTER_SECTION_NAME;
 extern const std::string ROCKSDB_ENV_SLOW_QUERY_THRESHOLD;
 
 extern const std::string ROCKSDB_ITERATION_THRESHOLD_TIME_MS;
+
+extern const std::string SPLIT_VALIDATE_PARTITION_HASH;
 } // namespace pegasus
--- a/src/server/key_ttl_compaction_filter.h
+++ b/src/server/key_ttl_compaction_filter.h
@@ -39,13 +39,13 @@ public:
                                bool enabled,
                                int32_t pidx,
                                int32_t partition_version,
-                               bool check_hash)
+                               bool validate_hash)
         : _pegasus_data_version(pegasus_data_version),
           _default_ttl(default_ttl),
           _enabled(enabled),
           _partition_index(pidx),
           _partition_version(partition_version),
-          _check_partition_hash(check_hash)
+          _validate_partition_hash(validate_hash)
     {
     }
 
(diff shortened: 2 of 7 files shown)


acelyc111 (dismissed):

src/server/key_ttl_compaction_filter.h
```diff
@@ -58,16 +69,28 @@ class KeyWithTTLCompactionFilter : public rocksdb::CompactionFilter
             *value_changed = true;
             return false;
         }
-        return check_if_ts_expired(utils::epoch_now(), expire_ts);
+        return check_if_ts_expired(utils::epoch_now(), expire_ts) || check_if_stale_split_data(key);
     }
 
     const char *Name() const override { return "KeyWithTTLCompactionFilter"; }
 
+    bool check_if_stale_split_data(const rocksdb::Slice &key) const
```
neverchanje, line 77:
```suggestion
    // Check if the record is stale after partition-split,
    // which splits the partition into two halves.
    // The stale record belongs to the other half.
    bool check_if_stale_split_data(const rocksdb::Slice &key) const
```

src/server/pegasus_server_impl.cpp
```diff
@@ -2438,6 +2442,25 @@ void pegasus_server_impl::update_rocksdb_iteration_threshold(
     }
 }
 
+void pegasus_server_impl::update_validate_partition_hash(
```
neverchanje, line 2445:
```suggestion
void pegasus_server_impl::update_check_partition_hash(
```

Be consistent with the concept "check_partition_hash", it would be ambiguous.

src/base/pegasus_const.h
```diff
@@ -62,4 +62,6 @@ extern const std::string PEGASUS_CLUSTER_SECTION_NAME;
 extern const std::string ROCKSDB_ENV_SLOW_QUERY_THRESHOLD;
 
 extern const std::string ROCKSDB_ITERATION_THRESHOLD_TIME_MS;
+
+extern const std::string SPLIT_VALIDATE_PARTITION_HASH;
```
neverchanje, line 66:
```suggestion
extern const std::string SPLIT_CHECK_PARTITION_HASH;
```

neverchanje (changes requested):

src/base/pegasus_const.h
hycdong: Actually, the app envs is called `replica.split.validate_partition_hash`, so I rename all `check_partition_hash` into `validate_partition_hash`

neverchanje (approved):

acelyc111 (approved):

PR #702, opened 2021-03-23
feat(split): scan support validate_partition_hash

hycdong:
This pull request supports validate partition_hash while scan, including c++ client and server implementation.
when scan_context.validate_partition_hash = true and app env validate_partition_hash = true, each iterator of scan will check partition_hash to avoid return stale data after partition split. If request.validate_partition_hash is not set, scan_context.validate_partition_hash = true, it is compatibility for old client. Besides, only full_scan need validate hash, hash_scan will not return stale data. 

The result of different client and server will like:
| client | server | hash_scan validate | full_scan validate |
| :-----|  :----- |  :----- |  :----- |
| old | splitting table | true | true |
| old | not-splitting table | false | false |
| new | splitting table | false | true |
| new | not-splitting table | false | false |


src/server/pegasus_server_impl.h
```diff
@@ -206,6 +206,7 @@ class pegasus_server_impl : public pegasus_read_service
     // return 1 if value is appended
     // return 2 if value is expired
     // return 3 if value is filtered
+    // return 4 if key is not pass partition_hash validation
```
acelyc111, line 209: Better to refactor these code and define them as an enum class

src/client_lib/pegasus_client_impl.h
```diff
@@ -265,12 +265,14 @@ class pegasus_client_impl : public pegasus_client
 
         pegasus_scanner_impl(::dsn::apps::rrdb_client *client,
                              std::vector<uint64_t> &&hash,
-                             const scan_options &options);
+                             const scan_options &options,
+                             bool validate_partition_hash);
```
levy5307, line 269: What about move validate_partition_hash into scan_options?

src/server/pegasus_server_impl.cpp
```diff
@@ -2064,6 +2070,16 @@ int pegasus_server_impl::append_key_value_for_scan(
         return 2;
     }
 
+    if (request_validate_hash && _validate_partition_hash) {
+        if (_partition_version < 0 || _gpid.get_partition_index() > _partition_version ||
+            !check_pegasus_key_hash(key, _gpid.get_partition_index(), _partition_version)) {
+            if (_verbose_log) {
+                derror_replica("not serve hash key while scan");
+            }
+            return 4;
```
levy5307, line 2079: What's the meaning of 1 2 3 and 4? could you make it a enum class?

src/server/pegasus_server_impl.h
hycdong: Sure~ I will refactor in next pull request, this pr only contains functional code.

src/server/pegasus_server_impl.cpp
hycdong: This is a history problem, you can reference `.h` file, I will refactor in next pull request, this pr only contains functional code.

src/client_lib/pegasus_client_impl.h
hycdong: Scan_options is a structure for user-defined options, `validate_partition_hash` should not be aware for user. For example, `get_scanner` interface is like:
```
int pegasus_client_impl::get_scanner(const std::string &hash_key,
                                     const std::string &start_sort_key,
                                     const std::string &stop_sort_key,
                                     const scan_options &options,
                                     pegasus_scanner *&scanner);
```

levy5307 (approved):

acelyc111 (approved):

PR #709, opened 2021-03-30
refactor: add some base classes to refactor pegasus_value_schema

levy5307:
Add some base classes to refactor pegasus_value_schema, for example: `pegasus_value_schema`, `pegasus_value_segment` and `pegasus_value_manager`.

The global design is here:
image

### Check List 

Tests 

- Unit test
No node actually. I will add unit tests in later pull requests.


src/base/pegasus_value_segment.h
```diff
@@ -0,0 +1,34 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+namespace pegasus {
+
+enum segment_type
```
neverchanje, line 24:
```suggestion
enum value_field_type
```

I would recommend calling this "field". Actually, no system that I have seen so far is calling a "field" to "segment".

src/base/pegasus_value_segment.h
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// TBD(zlw)
+};
+
+struct pegasus_value_segment
+{
+    virtual ~pegasus_value_segment() = default;
+    virtual segment_type type() = 0;
+};
```
neverchanje, line 33:
```suggestion
struct value_field
{
    virtual ~value_field() = default;
    virtual value_field_type type() = 0;
};
```

src/base/pegasus_value_schema.h
```diff
@@ -234,4 +236,31 @@ class pegasus_value_generator
...
+{
+    /// TBD(zlw)
+};
+
+struct generate_value_params
+{
+    /// TBD(zlw)
+};
+
+class pegasus_value_schema
```
neverchanje, line 249:
```suggestion
class value_schema
```

src/base/value_manager.h
```diff
@@ -0,0 +1,43 @@
...
+#include "pegasus_value_schema.h"
+#include <dsn/utility/singleton.h>
+
+namespace pegasus {
+
+class value_manager : public dsn::utils::singleton<value_manager>
+{
+public:
+    void register_schema(value_schema *schema);
+    /// using raw value in rocksdb and data version stored in meta column family to get data version
```
Smityz, line 31:
```suggestion
    /// using the raw value in rocksdb and data version stored in meta column family to get data version
```

src/base/value_manager.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#pragma once
+
+#include "pegasus_value_schema.h"
+#include <dsn/utility/singleton.h>
+
+namespace pegasus {
+
+class value_manager : public dsn::utils::singleton<value_manager>
```
neverchanje, line 27:
```suggestion
class value_schema_manager : public dsn::utils::singleton<value_manager>
```

value_schema_manager is more accurate

src/base/value_schema_manager.h
```diff
@@ -0,0 +1,43 @@
...
+    /// version
+    value_schema *get_value_schema(uint32_t meta_cf_data_version, dsn::string_view value) const;
+    value_schema *get_value_schema(uint32_t version) const;
+    value_schema *get_latest_value_schema() const;
+
+private:
+    value_schema_manager() = default;
+    friend class dsn::utils::singleton<value_schema_manager>;
+
+    std::map<data_version, value_schema *> _schemas;
```
neverchanje, line 41: Use array instead of map. You need to ensure the cost is as least as possible.

levy5307: There won't be too many schemas, so I think the cost won't be too much.
And I use std::map here is to maintain the order for these schemas,  which will be useful in function `get_latest_value_schema`

neverchanje (approved):

Shuo-Jia (approved):

PR #533, opened 2020-05-12
feat: overload dump_write_request of replication_app_base.h

Shuo-Jia:
### What problem does this PR solve? 
https://github.com/XiaoMi/rdsn/pull/414 forbids the large write size request, but the log information is simple, https://github.com/XiaoMi/rdsn/pull/458 provide the interface of `dump_write_request`, this pr overload it.

### Check List 

Tests 
- Manual test (add detailed scripts or steps below)
  -  start onebox with config `max_allowed_write_size = 5`
  -  use shell `set` any key-value whose size > 5
  -  check the log:
```
[2.4@127.0.0.1:34801] client from 127.0.0.1:1 write request body size exceed threshold, request=[put:hash_key=a,sort_key=abc], request_body_size = 99, max_allowed_write_size = 5, it will be rejected!
``` 

Related changes

- Need to cherry-pick to the release branch(**depends on https://github.com/XiaoMi/rdsn/pull/458 merged**)
- Need to be included in the release note

PR #718, opened 2021-04-09
fix: overload dump_write_request shouldn't auto-reply to client

Shuo-Jia:
#533 overload `dump_write_request` for https://github.com/XiaoMi/rdsn/pull/414 and cause the problem that the server will auto-reply to client with `ERR_OK`, so all the `response_client_write` after `dump_write_request` will be ignored for client, it also mean client will always receive `ERR_OK`, though pass `ERR_INVALID_DATA` in https://github.com/XiaoMi/rdsn/pull/414.

This PR use rpc_holder constructor replace `rpc::auto_replay`to dump request

###  Checklist 

##### Side effects
- large write request will be responsed with `ERR_INVALID_DATA` and throw exception at client side

##### Related changes

### DIFF
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -2846,49 +2846,35 @@ std::string pegasus_server_impl::dump_write_request(dsn::message_ex *request)
 {
     dsn::task_code rpc_code(request->rpc_code());
     if (rpc_code == dsn::apps::RPC_RRDB_RRDB_PUT) {
-        auto put = put_rpc::auto_reply(request).request();
+        auto put = put_rpc(request).request();
         ::dsn::blob hash_key, sort_key;
         pegasus_restore_key(put.key, hash_key, sort_key);
-        std::string request("put:");
-        request.append("hash_key=")
-            .append(pegasus::utils::c_escape_string(hash_key))
-            .append(",sort_key=")
-            .append(pegasus::utils::c_escape_string(sort_key));
-        return request;
+        return fmt::format("put: hash_key={}, sort_key={}",
+                           pegasus::utils::c_escape_string(hash_key),
+                           pegasus::utils::c_escape_string(sort_key));
     }
 
     if (rpc_code == dsn::apps::RPC_RRDB_RRDB_MULTI_PUT) {
-        auto multi_put = multi_put_rpc::auto_reply(request).request();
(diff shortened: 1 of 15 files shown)


levy5307 (dismissed):

src/server/pegasus_server_impl.cpp
```diff
@@ -2846,49 +2846,35 @@ std::string pegasus_server_impl::dump_write_request(dsn::message_ex *request)
...
-        std::string request("multi_put:");
-        request.append("hash_key=")
-            .append(pegasus::utils::c_escape_string((multi_put.hash_key))
-                        .append(",multi_put_count=")
-                        .append(std::to_string(multi_put.kvs.size())));
-        return request;
+        auto multi_put = multi_put_rpc(request).request();
+        return fmt::format("multi_put: hash_key={}, multi_put_count={}",
+                           pegasus::utils::c_escape_string(multi_put.hash_key),
+                           std::to_string(multi_put.kvs.size()));
```
Smityz, line 2861: no need to use `std::to_string`

src/server/pegasus_server_impl.cpp
```diff
@@ -2846,49 +2846,35 @@ std::string pegasus_server_impl::dump_write_request(dsn::message_ex *request)
...
-            .append(",check_sort_key=")
-            .append(pegasus::utils::c_escape_string(check_and_mutate.check_sort_key))
-            .append(",set_value_count=")
-            .append(std::to_string(check_and_mutate.mutate_list.size()));
-        return request;
+        auto check_and_mutate = check_and_mutate_rpc(request).request();
+        return fmt::format("check_and_mutate: hash_key={}, check_sort_key={}, set_value_count={}",
+                           pegasus::utils::c_escape_string(check_and_mutate.hash_key),
+                           pegasus::utils::c_escape_string(check_and_mutate.check_sort_key),
+                           std::to_string(check_and_mutate.mutate_list.size()));
```
Smityz, line 2877: no need to use `std::to_string`

levy5307 (approved):

neverchanje:
Good job. Is there any case of bug caused by this problem?

Shuo-Jia:
> Good job. Is there any case of bug caused by this problem?

`dump_request_write` is used for `large size limit`, I haven't found other case

neverchanje (approved):

PR #722, opened 2021-04-20
refactor: implement value_shcema_v0

levy5307:
implement value_schema_v0, which represents data version 0.

##### Tests 

- Unit test


src/base/value_schema_v0.h
```diff
@@ -0,0 +1,45 @@
...
+class value_schema_v0 : public value_schema
+{
+public:
+    value_schema_v0() = default;
+
+    std::unique_ptr<value_field> extract_field(dsn::string_view value, value_field_type type);
+    dsn::blob extract_user_data(std::string &&value);
+    void update_field(std::string &value, std::unique_ptr<value_field> segment);
+    rocksdb::SliceParts generate_value(const value_params &params);
+    data_version version() const { return VERSION_0; }
```
neverchanje, line 39:
```suggestion
    std::unique_ptr<value_field> extract_field(dsn::string_view value, value_field_type type) override;
    dsn::blob extract_user_data(std::string &&value) override;
    void update_field(std::string &value, std::unique_ptr<value_field> segment) override;
    rocksdb::SliceParts generate_value(const value_params &params) override;
    data_version version() override const { return VERSION_0; }
```

src/base/pegasus_value_schema.h
```diff
@@ -238,13 +238,24 @@ class pegasus_value_generator
...
 };
 
 struct value_params
 {
+    value_params(std::string &buf, std::vector<rocksdb::Slice> &slices)
+        : write_buf(buf), write_slices(slices)
+    {
+    }
+
     std::map<value_field_type, std::unique_ptr<value_field>> fields;
```
Smityz, line 254: Maybe we can replace it with `std::array" to get more efficiency by both time complexity(O(logN) vs O(1)) and [cache-friendly ](https://stackoverflow.com/questions/16699247/what-is-a-cache-friendly-code).

Smityz:
```
std::map<value_field_type, std::unique_ptr<value_field>> fields;
```
=>
```
std::array<std::unique_ptr<value_field>> fields;
// fields[value_field_type] = dsn::make_unique<time_tag_field>(time_tag);
```

levy5307: done

neverchanje:
If your intention of this PR is to add new value schema more conveniently, I can give a simple solution:

```cpp
struct value_schema_v2
{
    static constexpr size_t version_field_offset = 0;
    typedef version_type = uint8_t;

    static constexpr size_t expire_ts_field_offset = 1;
    typedef expire_ts_type = uint32_t;

    static constexpr size_t timetag_field_offset = 5;
    typedef timetag_type = uint64_t;

    value_schema_v2() {
      _write_buf.resize(sizeof(uint8_t) + sizeof(uint32_t) + sizeof(uint64_t));
      _write_slices.resize(2);
    }

    rocksdb::SliceParts generate_value(uint8_t version, uint32_t expire_ts, uint64_t timetag, dsn::string_view user_data) {
      data_output out(_write_buf);
      out.write<uint8_t>(version);
      out.write<uint32_t>(expire_ts);
      out.write<uint64_t>(user_data);

      _write_buf[0] = rocksdb::Slice(_write_buf.data, )
      return {&_write_slices[0], static_cast<int>(_write_slices.size())};
    }

    static uint8_t extract_version(string_view val);
    static uint32_t extract_expire_ts(string_view val);
    static uint64_t extract_timetag(string_view val);
    static blob:: extract_user_data(std::string&& val);

private:
    std::string _write_buf;
    std::vector<rocksdb::Slice> _write_slices;
};

void extract_expire_ts_from_value(int version, string_view user_data);
void extract_timetag_from_value(int version, string_view user_data);
void extract_version_from_value(int version, string_view user_data);
```

As you can see, this code is very similar to our thrift-generated code. It can be generated with a very simple template language, for example, yaml.

```yaml
0:
  - expire_ts : 
      type: uint32_t
      size: 4

1:
  - expire_ts : 
      type: uint32_t
      size: 4
  - timetag : 
      type: uint64_t
      size: 8

2:
  - version : 
      type: uint8_t
      size: 1
  - expire_ts : 
      type: uint32_t
      size: 4
  - timetag : 
      type: uint64_t
      size: 8
```

I would suggest implementing a tool that can generate the above code according to the template, instead of using class inheritance.
The value encoding/decoding is extremely critical and requires efficiency. I think we should avoid runtime type deduction as much as we can, for example:

```cpp
void extract_expire_ts_from_value(value_schema *schema, string_view user_data) {
    auto field = schema->extract_field(raw_value, pegasus::value_field_type::EXPIRE_TIMESTAMP); // type deduction #1
    auto expire_ts_field = static_cast<expire_timestamp_field *>(segment.get()); // type deduction #2
    return expire_ts_field->expire_ts;
```

There're 2 deductions here. I'm not indicating that this code will definitely perform poorly. But technically, I would prefer to write code without potential slow points.

Using a code generator can also avoid bugs so long as the tool itself is stable. So when we add another field in the future, we only need to modify the template file, rather than introduce more code.

levy5307:
> ```yaml
> uint64_t
> ```

I don't think it's a good idea. Since we should add a different interface for each filed, for example: `extract_version_from_value`. This means you should add a interface if you add a filed. And In this interface, you should deal with all of these version even if this version does't support it.

levy5307:
> If your intention of this PR is to add new value schema more conveniently, I can give a simple solution:
> 
> ```c++
> struct value_schema_v2
> {
>     static constexpr size_t version_field_offset = 0;
>     typedef version_type = uint8_t;
> 
>     static constexpr size_t expire_ts_field_offset = 1;
>     typedef expire_ts_type = uint32_t;
> 
>     static constexpr size_t timetag_field_offset = 5;
>     typedef timetag_type = uint64_t;
> 
>     value_schema_v2() {
>       _write_buf.resize(sizeof(uint8_t) + sizeof(uint32_t) + sizeof(uint64_t));
>       _write_slices.resize(2);
>     }
> 
>     rocksdb::SliceParts generate_value(uint8_t version, uint32_t expire_ts, uint64_t timetag, dsn::string_view user_data) {
>       data_output out(_write_buf);
>       out.write<uint8_t>(version);
>       out.write<uint32_t>(expire_ts);
>       out.write<uint64_t>(user_data);
> 
>       _write_buf[0] = rocksdb::Slice(_write_buf.data, )
>       return {&_write_slices[0], static_cast<int>(_write_slices.size())};
>     }
> 
>     static uint8_t extract_version(string_view val);
>     static uint32_t extract_expire_ts(string_view val);
>     static uint64_t extract_timetag(string_view val);
>     static blob:: extract_user_data(std::string&& val);
> 
> private:
>     std::string _write_buf;
>     std::vector<rocksdb::Slice> _write_slices;
> };
> 
> void extract_expire_ts_from_value(int version, string_view user_data);
> void extract_timetag_from_value(int version, string_view user_data);
> void extract_version_from_value(int version, string_view user_data);
> ```
> 
> As you can see, this code is very similar to our thrift-generated code. It can be generated with a very simple template language, for example, yaml.
> 
> ```yaml
> 0:
>   - expire_ts : 
>       type: uint32_t
>       size: 4
> 
> 1:
>   - expire_ts : 
>       type: uint32_t
>       size: 4
>   - timetag : 
>       type: uint64_t
>       size: 8
> 
> 2:
>   - version : 
>       type: uint8_t
>       size: 1
>   - expire_ts : 
>       type: uint32_t
>       size: 4
>   - timetag : 
>       type: uint64_t
>       size: 8
> ```
> 
> I would suggest implementing a tool that can generate the above code according to the template, instead of using class inheritance.
> The value encoding/decoding is extremely critical and requires efficiency. I think we should avoid runtime type deduction as much as we can, for example:
> 
> ```c++
> void extract_expire_ts_from_value(value_schema *schema, string_view user_data) {
>     auto field = schema->extract_field(raw_value, pegasus::value_field_type::EXPIRE_TIMESTAMP); // type deduction #1
>     auto expire_ts_field = static_cast<expire_timestamp_field *>(segment.get()); // type deduction #2
>     return expire_ts_field->expire_ts;
> ```
> 
> There're 2 deductions here. I'm not indicating that this code will definitely perform poorly. But technically, I would prefer to write code without potential slow points.
> 
> Using a code generator can also avoid bugs so long as the tool itself is stable. So when we add another field in the future, we only need to modify the template file, rather than introduce more code.

I don't think it's a good idea
1.  we should add a different interface for each filed, for example: `extract_version_from_value`. This means you should add a interface if you add a filed. And In this interface, you should deal with all of these version even if this version does't support it. The code is not elegant enough if we implement it like this.
2. `static_cast` doesn't cost too much time. And these is only one type deduction, not two.
3. It's too complex to implement a tool to generate these code.

src/base/pegasus_value_schema.h
```diff
@@ -238,13 +238,24 @@ class pegasus_value_generator
...
 {
-    std::map<value_field_type, std::unique_ptr<value_field>> fields;
-    /// TBD(zlw)
+    value_params(std::string &buf, std::vector<rocksdb::Slice> &slices)
+        : write_buf(buf), write_slices(slices)
+    {
+    }
+
+    std::array<std::unique_ptr<value_field>, FIELD_COUNT> fields;
+    // write_buf and write_slices are transferred from `pegasus_value_generator`, Which are used to
```
acelyc111, line 255:
```suggestion
    // write_buf and write_slices are transferred from `pegasus_value_generator`, which are used to
```

src/base/value_field.h
```diff
@@ -23,12 +23,39 @@ namespace pegasus {
...
 struct value_field
 {
     virtual ~value_field() = default;
     virtual value_field_type type() = 0;
 };
+
+struct expire_timestamp_field : public value_field
+{
+    explicit expire_timestamp_field(uint32_t timestamp) { expire_ts = timestamp; }
+    value_field_type type() { return value_field_type::EXPIRE_TIMESTAMP; };
```
acelyc111, line 41:
```suggestion
    value_field_type type() { return value_field_type::EXPIRE_TIMESTAMP; }
```

src/base/value_field.h
```diff
@@ -23,12 +23,39 @@ namespace pegasus {
...
 
 struct value_field
 {
     virtual ~value_field() = default;
     virtual value_field_type type() = 0;
 };
+
+struct expire_timestamp_field : public value_field
+{
+    explicit expire_timestamp_field(uint32_t timestamp) { expire_ts = timestamp; }
```
acelyc111, line 40:
```suggestion
    explicit expire_timestamp_field(uint32_t timestamp) : expire_ts(timestamp) {}
```

src/base/value_field.h
```diff
@@ -23,12 +23,39 @@ namespace pegasus {
...
+    value_field_type type() { return value_field_type::EXPIRE_TIMESTAMP; };
+
+    uint32_t expire_ts;
+};
+
+struct time_tag_field : public value_field
+{
+    explicit time_tag_field(uint64_t tag) { time_tag = tag; }
```
acelyc111, line 48:
```suggestion
    explicit time_tag_field(uint64_t tag) : time_tag(tag) {}
```

src/base/value_field.h
```diff
@@ -23,12 +23,39 @@ namespace pegasus {
...
+    value_field_type type() { return value_field_type::TIME_TAG; };
```
acelyc111, line 49:
```suggestion
    value_field_type type() { return value_field_type::TIME_TAG; }
```

src/base/value_field.h
```diff
@@ -23,12 +23,39 @@ namespace pegasus {
...
+
+    uint64_t time_tag;
+};
+
+struct user_data_field : public value_field
+{
+    explicit user_data_field(dsn::string_view data) { user_data = data; }
```
acelyc111, line 56:
```suggestion
    explicit user_data_field(dsn::string_view data) : user_data(data) {}
```

src/base/value_field.h
```diff
@@ -23,12 +23,39 @@ namespace pegasus {
...
+    value_field_type type() { return value_field_type::USER_DATA; };
```
acelyc111, line 57:
```suggestion
    value_field_type type() { return value_field_type::USER_DATA; }
```

src/base/value_schema_v0.cpp
```diff
@@ -0,0 +1,98 @@
...
+    default:
+        dassert_f(false, "Unsupported field type: {}", type);
+    }
+    return field;
+}
+
+dsn::blob value_schema_v0::extract_user_data(std::string &&value)
+{
+    auto ret = dsn::blob::create_from_bytes(std::move(value));
+    ret.range(sizeof(uint32_t));
```
acelyc111, line 43: What does this line mean?

acelyc111 (approved):

hycdong (approved):

PR #737, opened 2021-05-26
feat: implement data version 2

levy5307:
proposal https://github.com/apache/incubator-pegasus/blob/master/rfcs/2020-10-09-data-version-v3.md
implement value_schema_v2, which represents data version 2.
The format of data version 2:
```
|- 1bit -|- version(7bits) -|- expire_ts(4bytes) -|- timetag(8 bytes) -|- user value(bytes) -|
```

##### Tests 

- Unit test

### DIFF
--- a/src/base/value_schema_v2.cpp
+++ b/src/base/value_schema_v2.cpp
@@ -76,7 +76,7 @@ rocksdb::SliceParts value_schema_v2::generate_value(const value_params &params)
     }
 
     params.write_buf.resize(sizeof(uint8_t) + sizeof(uint32_t) + sizeof(uint64_t));
-    uint8_t version_2 = 2;
+    const uint8_t version_2 = 2;
     dsn::data_output output(params.write_buf);
     output.write_u8(0x80 | version_2);            // version
     output.write_u32(expire_ts_field->expire_ts); // expire_ts
--- a/src/base/value_schema_v2.cpp
+++ b/src/base/value_schema_v2.cpp
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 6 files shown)


acelyc111 (approved):
LGTM

src/base/value_schema_v2.cpp
```diff
@@ -0,0 +1,117 @@
...
+    default:
+        dassert_f(false, "Unsupported field type: {}", type);
+    }
+    return field;
+}
+
+dsn::blob value_schema_v2::extract_user_data(std::string &&value)
+{
+    auto ret = dsn::blob::create_from_bytes(std::move(value));
+    return ret.range(sizeof(uint8_t) + sizeof(uint32_t) + sizeof(uint64_t));
```
Smityz, line 49: Can these values calculate in advance as a const value?

hycdong:
In my view, update data_version may raise compatible problems, which is NOT included in this pr. Will you raise another pr to display how to apply data_version 2 in future? This pr only adds version 2 format.

levy5307:
> In my view, update data_version may raise compatible problems, which is NOT included in this pr. Will you raise another pr to display how to apply data_version 2 in future? This pr only adds version 2 format.

Yes.

src/base/value_schema_v2.cpp
levy5307: How to calculate it in advance?

hycdong (approved):

src/base/value_schema_v2.cpp
acelyc111: I think @Smityz mean add a static const variable like `header_length = sizeof(uint8_t) + sizeof(uint32_t) + sizeof(uint64_t)` ?

levy5307: OK, I will refactor it in next pull request

PR #739, opened 2021-05-28
feat: implement value_schema_manager::get_value_schema

levy5307:
proposal https://github.com/apache/incubator-pegasus/blob/master/rfcs/2020-10-09-data-version-v3.md

implement value_schema_manager::get_value_schema

##### Tests 

- Unit test

### DIFF
--- a/rfcs/2021-05-27-user-specified-compaction.md
+++ b/rfcs/2021-05-27-user-specified-compaction.md
@@ -1,0 +1,40 @@
+# User Specified Compaction
+
+## Summary
+
+In Pegasus, sometimes we should add user specified compaction policy to reduce disk usage. This RFC proposes a user specified compaction design.
+
+## Design
+
+Add two classes named with `compaction_operation` and `compaction_rule`.
+
+`compaction_filter_rule` represents the compaction rule to filter the keys which are stored in rocksdb.
+There are three types of compaction operation:
+- Hashkey rule, which supports prefix match, postfix match and anywhere match.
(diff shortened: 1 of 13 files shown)


hycdong:
The old `value_schema` is a table level concept defined in `data_version` in meta column family. Now, `value_schema` version 2 updates it, `value_schema` becomes a value level concept, a table is permitted having different value schema. However, it will raise two problems.
The first problem is the compatibility problem. New server version won't be downgraded because the old server can not figure out value_schema version 2 data. I think it is okay for new server can not downgrade, but our recent releases are almost all not-downgraded, should we merge this commit into master branch? 
The second problem is the `data_version` in meta cf. This value is actually useless for new server, but it can not be removed for version0 and version1, would you have any solution to remove this value in future?

levy5307:
> The old `value_schema` is a table level concept defined in `data_version` in meta column family. Now, `value_schema` version 2 updates it, `value_schema` becomes a value level concept, a table is permitted having different value schema. However, it will raise two problems.
> The first problem is the compatibility problem. New server version won't be downgraded because the old server can not figure out value_schema version 2 data. I think it is okay for new server can not downgrade, but our recent releases are almost all not-downgraded, should we merge this commit into master branch?
> The second problem is the `data_version` in meta cf. This value is actually useless for new server, but it can not be removed for version0 and version1, would you have any solution to remove this value in future?

I don't think so. Because value_schema version 2 will not updates it. It only reads it from meta cf.

hycdong (dismissed):

acelyc111 (approved):

hycdong (approved):

PR #741, opened 2021-05-31
feat: add hashkey pattern rule

levy5307:

##### Tests 

- Unit test

### DIFF
--- a/src/server/compaction_filter_rule.cpp
+++ b/src/server/compaction_filter_rule.cpp
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "compaction_filter_rule.h"
+
+#include <dsn/dist/fmt_logging.h>
+#include <dsn/utility/string_view.h>
+#include <dsn/c/api_utilities.h>
+
+namespace pegasus {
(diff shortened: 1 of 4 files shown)


src/server/compaction_filter_rule.h
```diff
@@ -36,5 +38,34 @@ class compaction_filter_rule
...
+ENUM_BEGIN(string_match_type, SMT_INVALID)
+ENUM_REG(SMT_MATCH_ANYWHERE)
+ENUM_REG(SMT_MATCH_PREFIX)
+ENUM_REG(SMT_MATCH_POSTFIX)
+ENUM_END(string_match_type)
+
+class hashkey_pattern_rule : public compaction_filter_rule
+{
+public:
+    explicit hashkey_pattern_rule(uint32_t pegasus_data_version);
```
hycdong, line 58: Why do hashkey pattern need `pegasus_data_version`? Is `pegasus_data_version` used for value schema?

src/server/compaction_filter_rule.cpp
```diff
@@ -0,0 +1,62 @@
...
+namespace pegasus {
+namespace server {
+bool string_pattern_match(const std::string &value,
+                          string_match_type type,
+                          const std::string &filter_pattern)
+{
+    if (filter_pattern.length() == 0)
+        return false;
+    if (value.length() < filter_pattern.length())
+        return false;
```
acelyc111, line 35:
```suggestion
    if (filter_pattern.empty()) {
        return false;
    }
    if (value.length() < filter_pattern.length()) {
        return false;
    }
```

acelyc111, line 35: UT doesn't cover this case.

src/server/compaction_filter_rule.cpp
```diff
@@ -0,0 +1,62 @@
...
+    case string_match_type::SMT_MATCH_ANYWHERE:
+        return dsn::string_view(value).find(filter_pattern) != dsn::string_view::npos;
+    case string_match_type::SMT_MATCH_PREFIX:
+        return memcmp(value.data(), filter_pattern.data(), filter_pattern.length()) == 0;
+    case string_match_type::SMT_MATCH_POSTFIX:
+        return memcmp(value.data() + value.length() - filter_pattern.length(),
+                      filter_pattern.data(),
+                      filter_pattern.length()) == 0;
+    default:
+        derror_f("invalid match type {}", type);
```
acelyc111, line 47: UT doesn't cover this case.

src/server/compaction_filter_rule.h
levy5307: Yes, I can remove it in this pull request.

hycdong (approved):

acelyc111 (approved):
LGTM

Issue #727, opened 2021-04-27
Support preserving TTL for copy_data


empiredan:
### Background

Sometimes our customs demand a migration to new cluster, which inevitably includes copying data. On the other hand, the customs tend to use TTL to support some business logic. Therefore,  missing TTL in the new cluster will impact greatly on the customs' business, for example, the business reads a data that should have disappeared. 

However, currently copy_data has not yet supported preserving TTL, since it first scans user_data without TTL, leaving ttl_seconds 0 for the later async_set.

### Solution

We can add an option for copy_data, for example, "-k/--preserve-ttl". With this option, data will be copied to the target cluster with remaining TTL, if any.

### Implementation

A possible implementation is described as below:
* add "-k/--preserve-ttl" option for copy_data in shell commands;
* support scanning user_data with TTL from rocksdb data;
* async_set with the remaining TTL in the target cluster

Smityz:
It sounds like a great idea, would you like to implement it in https://github.com/pegasus-kv/admin-cli ?

Shuo-Jia:
> It sounds like a great idea, would you like to implement it in https://github.com/pegasus-kv/admin-cli ?

@empiredan cplus shell will be abandoned, we can consider add the feature in admin-cli.

PR #752, opened 2021-06-03
feat: support preserving TTL for copy_data

empiredan:
## What problem does this PR solve? 

As we discussed in https://github.com/apache/incubator-pegasus/issues/727, this PR is to support preserving TTL for copy_data both in the client and the server side.

## What is changed and how does it work?

An option `-e/--no_ttl` is added for copy_data, with an optional boolean `return_expire_ts` added in get_scanner_request in the client side to tell the server that the expiration timestamp should be returned if `-e` is not set, by default. In the server side,  scan_response will return key/value with expiration timestamp if requested.

To facilitate the test, C++ pegasus shell is also changed, though it is considered deprecated soon (discussed in https://github.com/apache/incubator-pegasus/issues/727). Later this option will be implemented in https://github.com/pegasus-kv/admin-cli together with copy_data command.

#### copy_data without -e by default
copy the data from an app to another app without `-e` to check if TTL is retained.

#### copy_data with -e
copy the data from an app to another app with `-e` to check if TTL is ignored and lifetime is set infinitely.

#### copy_data without -e and with -n
copy the data from an app to another app without `-e` and with `-n` to check only the data that are not existing be copied with TTL.

#### compatibility test
- Backward compatibility test: use new shell to connect the old server without `-e` (where data should be copied with no TTL as infinite lifetime)
- Forward compatibility test: use old shell to connect the new server to ensure copy_data can work well


src/idl/rrdb.thrift
```diff
@@ -253,6 +253,7 @@ struct get_scanner_request
     9:filter_type  sort_key_filter_type;
     10:dsn.blob    sort_key_filter_pattern;
     11:optional bool    validate_partition_hash;
+    12:optional bool    return_expire_ts;
```
acelyc111, line 256: Would it better to add `return_expire_ts` to `struct key_value`, like:
```
struct key_value
{
    1:dsn.blob      key;
    2:dsn.blob      value;
    3:optional bool expire_ts;
}
```
It would be more general and extensible.

empiredan: Good idea ! expire_ts should be a member of key_value.  Should we use `i32` instead of `bool` for `expire_ts` which is used in `scan_response` to return all of the key-value pairs with optional expire_ts:
```
struct key_value
{
    1:dsn.blob      key;
    2:dsn.blob      value;
    3:optional i32 expire_ts;
}
```
where I've used an `expire_ts_seconds_list` to separate the optional `expire_ts` to another list. However, since `return_expire_ts` in get_scanner_request tells the server that the client need expire_ts, should we keep it ?

acelyc111: Oh, yes, it's type of `i32`, and you can keep it in `get_scanner_request`.

empiredan: OK

empiredan: To follow the previous style, `expire_ts` is fixed as `expire_ts_seconds`.

src/server/pegasus_server_impl.cpp
```diff
@@ -1009,6 +1009,8 @@ void pegasus_server_impl::on_get_scanner(get_scanner_rpc rpc)
     uint32_t batch_count = std::min(request_batch_size, _rng_rd_opts.rocksdb_max_iteration_count);
     resp.kvs.reserve(batch_count);
 
+    bool return_expire_ts = request.__isset.return_expire_ts ? request.return_expire_ts : false;
```
acelyc111, line 1012:
```suggestion
    bool return_expire_ts = request.__isset.return_expire_ts && request.return_expire_ts;
```

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -156,6 +161,11 @@ void pegasus_client_impl::pegasus_scanner_impl::_async_next_internal()
         std::string hash_key, sort_key;
         pegasus_restore_key(_kvs[_p].key, hash_key, sort_key);
         std::string value(_kvs[_p].value.data(), _kvs[_p].value.length());
+        int ttl_seconds = 0;
+        if (_kvs[_p].__isset.expire_ts_seconds) {
+            auto expire_ts_seconds = static_cast<uint32_t>(_kvs[_p].expire_ts_seconds);
+            ttl_seconds = static_cast<int>(expire_ts_seconds - utils::epoch_now());
```
acelyc111, line 167: We should prevent a corner case that `ttl_seconds` overflowed. Because it's a RPC, means it is possible to cost some seconds to get scan response( or there is clock drift), suppose `expire_ts_seconds` is `1600000000`, `epoch_now` on server side is `1599999999`, so it's not expired and will return to client. But on client side `epoch_now` is `1600000001`, then `ttl_seconds` is overflowed and will set this kv a large ttl in callback which is not what we expect.

acelyc111 (approved):
LGTM

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -61,8 +61,12 @@ int pegasus_client_impl::pegasus_scanner_impl::next(std::string &hashkey,
...
     ::dsn::utils::notify_event op_completed;
     int ret = -1;
-    auto callback = [&](
-        int err, std::string &&hash, std::string &&sort, std::string &&str, internal_info &&ii) {
+    auto callback = [&](int err,
+                        std::string &&hash,
+                        std::string &&sort,
+                        std::string &&str,
+                        internal_info &&ii,
+                        uint32_t expire_ts_seconds) {
```
levy5307, line 69: I think it's a bug for not copying ttl in `copy_data`. Maybe we should copy it in default

acelyc111 (dismissed):
LGTM

levy5307 (approved):

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -61,8 +61,12 @@ int pegasus_client_impl::pegasus_scanner_impl::next(std::string &hashkey,
 {
     ::dsn::utils::notify_event op_completed;
     int ret = -1;
-    auto callback = [&](
-        int err, std::string &&hash, std::string &&sort, std::string &&str, internal_info &&ii) {
+    auto callback = [&](int err,
+                        std::string &&hash,
+                        std::string &&sort,
+                        std::string &&str,
```
Shuo-Jia, line 67:
```suggestion
                        std::string &&value,
```

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -156,6 +161,10 @@ void pegasus_client_impl::pegasus_scanner_impl::_async_next_internal()
         std::string hash_key, sort_key;
         pegasus_restore_key(_kvs[_p].key, hash_key, sort_key);
         std::string value(_kvs[_p].value.data(), _kvs[_p].value.length());
+        uint32_t expire_ts_seconds = 0;
+        if (_kvs[_p].__isset.expire_ts_seconds) {
+            expire_ts_seconds = static_cast<uint32_t>(_kvs[_p].expire_ts_seconds);
+        }
```
Shuo-Jia, line 167:
```suggestion
        uint32_t expire_ts_seconds = _kvs[_p].__isset.expire_ts_seconds ? static_cast<uint32_t>(_kvs[_p].expire_ts_seconds : expire_ts_seconds;

```

src/server/pegasus_server_impl.cpp
```diff
@@ -2135,6 +2142,13 @@ pegasus_server_impl::append_key_value_for_scan(std::vector<::dsn::apps::key_valu
     ::memcpy(key_buf.get(), raw_key.data(), raw_key.length());
     kv.key.assign(std::move(key_buf), 0, raw_key.length());
 
+    // extract expire ts if necessary
+    if (request_expire_ts) {
+        auto expire_ts_seconds =
+            pegasus_extract_expire_ts(_pegasus_data_version, utils::to_string_view(value));
+        kv.__set_expire_ts_seconds(static_cast<int32_t>(expire_ts_seconds));
```
Shuo-Jia, line 2149: should `static_cast<uint32_t>`?

src/shell/commands/data_operations.cpp
```diff
@@ -1540,6 +1540,7 @@ bool copy_data(command_executor *e, shell_context *sc, arguments args)
                                            {"no_overwrite", no_argument, 0, 'n'},
                                            {"no_value", no_argument, 0, 'i'},
                                            {"geo_data", no_argument, 0, 'g'},
+                                           {"no_ttl", no_argument, 0, 'e'},
```
Shuo-Jia, line 1543: you declare `no_ttl` here,  but options  declare `return_expire_ts`, I suggest they should be coincident

hycdong: +1

src/include/pegasus/client.h
```diff
@@ -251,14 +251,16 @@ class pegasus_client
...
         scan_options()
             : timeout_ms(5000),
               batch_size(100),
               start_inclusive(true),
               stop_inclusive(false),
               hash_key_filter_type(FT_NO_FILTER),
               sort_key_filter_type(FT_NO_FILTER),
-              no_value(false)
+              no_value(false),
+              return_expire_ts(true)
```
hycdong, line 263: I suggest that the default value of `return_expire_ts` should be `false`, copy data is only a case of scan, for most scan operations, expire timestamp is not a significant option.

src/shell/commands/data_operations.cpp
empiredan: Actually I think `no_ttl` and `return_expire_ts` are different. `no_ttl` tells that by default ttl will be kept while we use shell to copy data. `return_expire_ts` is used not only for copy data,  in fact  at any time that `get_scanner_request` is called the developer should decide whether `expire_ts` is needed. For now, at most cases,  `expire_ts` is not needed by the caller of `get_scanner_request`, therefore I set `return_expire_ts` as ture to tell server that `expire_ts` should be returned. However, for the most cases the callers just ignore `return_expire_ts` and the server only return key/value without `expire_ts` which is not needed.

src/include/pegasus/client.h
empiredan: Yes, I'll follow this.

src/shell/commands/data_operations.cpp
hycdong: Okay, I have been persuaded, you can wait for @Shuo-Jia 's opinion.

src/client_lib/pegasus_scanner_impl.cpp
Shuo-Jia: yeah, better named `meaningful` name, left value can be short, val is ok

src/client_lib/pegasus_scanner_impl.cpp
Shuo-Jia: you are right

src/server/pegasus_server_impl.cpp
Shuo-Jia: Oh, OK

src/shell/commands/data_operations.cpp
Shuo-Jia: Yeah, I see

acelyc111 (approved):

Shuo-Jia (approved):

hycdong:
How about adding a test for scan when return_expire_ts = true?

empiredan:
> How about adding a test for scan when return_expire_ts = true?

I think we could add the test into `src/test/function_test/test_scan.cpp`, where first we put some data with ttl, then scan with `return_expire_ts` set as `true` or `false` to check whether the responded data has `expire_ts_seconds`. Is this alright, @hycdong ?

hycdong:
> > How about adding a test for scan when return_expire_ts = true?
> 
> I think we could add the test into `src/test/function_test/test_scan.cpp`, where first we put some data with ttl, then scan with `return_expire_ts` set as `true` or `false` to check whether the responded data has `expire_ts_seconds`. Is this alright, @hycdong ?

Sure, you're right~

Shuo-Jia (approved):

hycdong (approved):

acelyc111 (approved):

PR #768, opened 2021-06-16
feat: add compaction filter rule creator

levy5307:

##### Tests 

- Unit test

### DIFF
--- a/src/server/test/compaction_filter_rule_test.cpp
+++ b/src/server/test/compaction_filter_rule_test.cpp
@@ -131,5 +131,43 @@ TEST(ttl_range_rule_test, match)
         ASSERT_EQ(rule.match("", "", svalue.parts[0]), test.match);
     }
 }
+
+TEST(compaction_filter_rule_test, create)
+{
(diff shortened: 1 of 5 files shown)


src/server/test/compaction_filter_rule_test.cpp
```diff
@@ -131,5 +131,43 @@ TEST(ttl_range_rule_test, match)
         ASSERT_EQ(rule.match("", "", svalue.parts[0]), test.match);
     }
 }
+
+TEST(compaction_filter_rule_test, create)
+{
+    const uint32_t data_version = 1;
+
+    // ttl_range_rule
+    std::string params_json = "{\"start_ttl\":1000,\"stop_ttl\":2000}";
```
hycdong, line 140: I recommend using [string_literal](https://en.cppreference.com/w/cpp/language/string_literal) to simplify the json string.

levy5307: Done

hycdong (approved):

acelyc111 (approved):

PR #769, opened 2021-06-18
feat: add creator for compaction operation

levy5307:

##### Tests 

- Unit test

### DIFF
--- a/src/server/test/compaction_operation_test.cpp
+++ b/src/server/test/compaction_operation_test.cpp
@@ -222,5 +222,24 @@ TEST(update_ttl_test, filter)
         }
     }
 }
+
+TEST(compaction_filter_operation_test, creator)
+{
+    uint32_t data_version = 1;
+    std::string params_json = R"({"type":"UTOT_FROM_CURRENT","value":2000})";
+    update_ttl *update_ttl_op =
+        static_cast<update_ttl *>(update_ttl::creator(params_json, {}, data_version));
+    ASSERT_EQ(update_ttl_op->value, 2000);
+    ASSERT_EQ(update_ttl_op->type, UTOT_FROM_CURRENT);
+    delete update_ttl_op;
+
+    // invalid operation
+    params_json = R"({"type_xxx":"UTOT_FROM_CURRENT","value":2000})";
+    compaction_operation *invalid_op = update_ttl::creator(params_json, {}, data_version);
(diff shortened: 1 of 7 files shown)


src/server/test/compaction_operation_test.cpp
```diff
@@ -222,5 +222,60 @@ TEST(update_ttl_test, filter)
...
+    ASSERT_EQ(invalid_op, nullptr);
+    params_json = R"({"type":"UTOT_FROM_CURRENT","value_xxx":2000})";
+    invalid_op = update_ttl::creator(params_json, data_version);
+    ASSERT_EQ(invalid_op, nullptr);
+}
+
+TEST(compaction_filter_operation_test, create_operations)
+{
+    std::string json =
+        "{\"ops\":[{\"type\":\"COT_DELETE\",\"params\":\"\",\"rules\":[{\"type\":\"FRT_HASHKEY_"
```
levy5307, line 248: I copied this json string from app env and I don't know how to change it to string literal :(
Actuallly, I have tried some ways but they can't work well.

hycdong: Have you ever tried to use it like the following format? 
1. https://github.com/XiaoMi/rdsn/blob/7ccb58ed9c6319f1a8ccb5fe28276a8a1014379a/src/replica/duplication/test/replica_http_service_test.cpp#L47-L53
2. https://github.com/XiaoMi/rdsn/blob/v2.2/src/meta/test/meta_http_service_test.cpp#L42-L43

levy5307: Yes, but it can't work. I think it is because the rule is formatting twice time for json

hycdong: How about using the following format?
```
std::string str1 = R"{}";
std::string str2 = R"{}";
...
// json is concat result of str1 and str2
std::string json = str1 + str2;
```

levy5307: I think it's ok to stay the same. Because I copied it from app env, so we can use this unit-test to show whether the app env can works well.

hycdong: Okay, I agree with you.

hycdong (approved):

Shuo-Jia (approved):

PR #774, opened 2021-06-25
feat: add more rocksdb  perf-counter support

Shuo-Jia:
some rocksdb has influence with read/write latency, we need add perf-counter support

##### Tests 

- Manual test 

##### Side effects
read amp need external memory to obtain, detail see: https://github.com/XiaoMi/pegasus-rocksdb/blob/v6.6.4-compatible/include/rocksdb/table.h#L247, **so we disable it by default**

### Config update
```diff
+ read_amp_bytes_per_bit = 0 # 0 means disble read amp counter
- update_rdb_stat_interval = 600
+ update_rdb_stat_interval = 60
```
### Counter update
```diff
+ replica*app.pegasus*rdb.read_amplification@gpid
+ replica*app.pegasus*rdb.write_amplification@gpid
+ replica*app.pegasus.rdb.read_memtable_total_count@gpid
+ replica*app.pegasus.rdb.read_memtable_hit_count@gpid
+ replica*app.pegasus*rdb.read_l0_hit_count@gpid
+ replica*app.pegasus*rdb.read_l1_hit_count@gpid
+ replica*app.pegasus*rdb.read_l2andup_hit_count@gpid
```


src/server/pegasus_server_impl.cpp
```diff
@@ -1532,13 +1535,20 @@ ::dsn::error_code pegasus_server_impl::start(int argc, char **argv)
                                       [this]() { this->update_replica_rocksdb_statistics(); },
                                       _update_rdb_stat_interval);
 
+    dinfo_replica("start the update rocksdb amp statistics timer task");
+    _update_replica_amp_stat =
+        ::dsn::tasking::enqueue_timer(LPC_REPLICATION_LONG_COMMON,
+                                      nullptr,
```
Smityz, line 1541: why not use tracker here

src/server/pegasus_server_impl.h
```diff
@@ -39,7 +40,7 @@
 
 namespace pegasus {
 namespace server {
-
+DSN_DECLARE_int32(read_amp_bytes_per_bit);
```
levy5307, line 43: move this to the corresponding cpp file

src/server/pegasus_server_impl_init.cpp
```diff
@@ -44,6 +44,20 @@ DSN_DEFINE_bool("pegasus.server",
...
+// https://github.com/XiaoMi/pegasus-rocksdb/blob/v6.6.4-compatible/include/rocksdb/table.h#L247
+DSN_DEFINE_int32("pegasus.server",
+                 read_amp_bytes_per_bit,
+                 0,
+                 "config for using to calculate the "
+                 "read amplification, must be a power "
+                 "of 2, zero means disable count read "
+                 "amplification");
+
+DSN_DEFINE_validator(read_amp_bytes_per_bit, [](const int64_t read_amp_bytes_per_bit) -> bool {
```
Smityz, line 57: you can validator here is `read_amp_bytes_per_bi`t a power of 2 or equal to 0

src/server/pegasus_server_impl_init.cpp
```diff
@@ -44,6 +44,20 @@ DSN_DEFINE_bool("pegasus.server",
...
+    return read_amp_bytes_per_bit >= 0;
```
levy5307, line 58: ``` return read_amp_bytes_per_bit >= 0 && read_amp_bytes_per_bit & (read_amp_bytes_per_bit - 1) == 0;```

src/server/pegasus_server_impl.cpp
```diff
@@ -2274,6 +2288,33 @@ void pegasus_server_impl::update_replica_rocksdb_statistics()
...
+        auto write_amplification = write_amplification_iter == props.end()
+                                       ? 1
+                                       : std::stod(write_amplification_iter->second);
+        _pfc_rdb_write_amplification->set(write_amplification);
+        dinfo_replica("_pfc_rdb_write_amplification: {}", write_amplification);
+    }
+
+    // Update _pfc_rdb_read_amplification
+    if (FLAGS_read_amp_bytes_per_bit > 0) {
+        int64_t estimate_useful_bytes =
```
levy5307, line 2306: uint64_t

src/server/pegasus_server_impl.cpp
Shuo-Jia: should use

src/server/pegasus_server_impl.h
```diff
@@ -26,6 +26,7 @@
 #include <rocksdb/options.h>
 #include <dsn/perf_counter/perf_counter_wrapper.h>
 #include <dsn/dist/replication/replication.codes.h>
+#include <dsn/utility/flags.h>
```
levy5307, line 29: remove this line

levy5307 (dismissed):

src/server/pegasus_server_impl.cpp
```diff
@@ -46,6 +46,7 @@ namespace pegasus {
 namespace server {
 
 DEFINE_TASK_CODE(LPC_PEGASUS_SERVER_DELAY, TASK_PRIORITY_COMMON, ::dsn::THREAD_POOL_DEFAULT)
+DSN_DECLARE_int32(read_amp_bytes_per_bit);
```
hycdong, line 49: I suggest this config mutable.

Smityz:
LGTM

src/server/pegasus_server_impl.cpp
```diff
@@ -2274,6 +2289,33 @@ void pegasus_server_impl::update_replica_rocksdb_statistics()
...
+                                       ? 1
+                                       : std::stod(write_amplification_iter->second);
+        _pfc_rdb_write_amplification->set(write_amplification);
+        dinfo_replica("_pfc_rdb_write_amplification: {}", write_amplification);
+    }
+
+    // Update _pfc_rdb_read_amplification
+    if (FLAGS_read_amp_bytes_per_bit > 0) {
+        auto estimate_useful_bytes =
+            _statistics->getTickerCount(rocksdb::READ_AMP_ESTIMATE_USEFUL_BYTES);
```
hycdong, line 2308: https://github.com/facebook/rocksdb/blob/v6.6.4/monitoring/statistics.cc
I notice that there are other statistics similar to what we like, such as `GET_HIT_L0`, `GET_HIT_L1`, `GET_HIT_L2_AND_UP`, what it the difference between them?

src/server/pegasus_server_impl.cpp
Shuo-Jia: it actually pass to rocksdb, so won't be updated rocksdb though set mutable

src/server/pegasus_server_impl.cpp
Shuo-Jia: Oh, I haven't noticed, it seem hitRate under different level

hycdong: Yes~ I think they are actually what we need, more accurate than read amplification rate

Shuo-Jia: https://github.com/facebook/rocksdb/blob/v6.6.4/include/rocksdb/statistics.h#L101-L106 it hit count, we need calc `hitRate` via them

hycdong (approved):

Smityz (approved):

PR #780, opened 2021-07-13
feat: start compaction operation in CompactionFilter::Filter

levy5307:
start compaction operation in CompactionFilter::Filter

##### Tests 

1. insert a value into app temp
```
>>> use temp
OK
>>> set a b c 3000
OK

app_id          : 2
partition_index : 4
decree          : 4
server          : 10.231.57.100:34802
```
2. add a compaction operation which delete key with ttl in [1000, 10000]
3. start manual compact
4. get the value, and found it was deleted
```
>>> get a b 
Not found

app_id          : 2
partition_index : 4
server          : 10.231.57.100:34802
>>> 
```


src/server/key_ttl_compaction_filter.h
```diff
@@ -60,6 +62,23 @@ class KeyWithTTLCompactionFilter : public rocksdb::CompactionFilter
             return false;
         }
 
+        // ignore empty write
+        if (key.size() < 2) {
```
Shuo-Jia, line 66: if user's key < 2 actually?

src/server/key_ttl_compaction_filter.h
```diff
@@ -60,6 +62,23 @@ class KeyWithTTLCompactionFilter : public rocksdb::CompactionFilter
...
+            pegasus_restore_key(dsn::blob(key.data(), 0, key.size()), hash_key, sort_key);
+            for (const auto &op : _user_specified_operations) {
+                if (op->filter(hash_key, sort_key, existing_value, new_value, value_changed)) {
+                    // return true if this data need to be deleted
+                    return true;
+                }
+            }
+        }
+
         uint32_t expire_ts =
```
Shuo-Jia, line 82: suggest follow for undestanding the compaction action:
```c++
return user_specified_operations_filter || records_ttl_expired

function user_specified_operations_filter() {
return op->filter(hash_key, sort_key, existing_value, new_value, value_changed)
}

function records_ttl_expired() {
return check_if_ts_expired(utils::epoch_now(), expire_ts) || check_if_stale_split_data(key);
}
```

src/server/key_ttl_compaction_filter.h
levy5307: In the compaction operation,  It will extract field from key. If key.size() < 2, it will cause coredump

Smityz (approved):

Shuo-Jia (approved):

hycdong (approved):

Issue #617, opened 2020-10-13
thrift unmarshall bug in 1.12.3 release


hycdong:
### Pegasus version
Pegasus Server 1.12.3 (a948e89b180b6a5c82d298d0dcc65f7bb770a8be) 

### Coredump stack
```
(gdb) bt
#0  0x00007fbb92c8c1d7 in raise () from /lib64/libc.so.6
#1  0x00007fbb92c8d8c8 in abort () from /lib64/libc.so.6
#2  0x00007fbb92c85146 in __assert_fail_base () from /lib64/libc.so.6
#3  0x00007fbb92c851f2 in __assert_fail () from /lib64/libc.so.6
#4  0x00007fbb96c0cb9f in dsn::binary_reader::read (this=0x7fbb58330d20, buffer=buffer@entry=0x70b900000 "\340\031\367\256\005", sz=sz@entry=16777216)
    at /home/wutao1/pegasus-release/rdsn/src/core/core/binary_reader.cpp:80
#5  0x00000000005e30e5 in read (len=16777216, buf=0x70b900000 "\340\031\367\256\005", this=0x7fbb58330cc0)
    at /home/wutao1/pegasus-release/DSN_ROOT/include/dsn/cpp/serialization_helper/thrift_helper.h:66
#6  apache::thrift::transport::readAll<dsn::binary_reader_transport> (trans=..., buf=0x70b900000 "\340\031\367\256\005", len=16777216)
    at /home/wutao1/pegasus-release/rdsn/thirdparty/output/include/thrift/transport/TTransport.h:41
#7  0x00000000005e9702 in readAll (len=16777216, buf=<optimized out>, this=<optimized out>)
    at /home/wutao1/pegasus-release/rdsn/thirdparty/output/include/thrift/transport/TTransport.h:121
#8  apache::thrift::protocol::TBinaryProtocolT<apache::thrift::transport::TTransport, apache::thrift::protocol::TNetworkBigEndian>::readStringBody<dsn::blob_string> (
    this=this@entry=0x7fbb58330ce0, str=..., size=16777216) at /home/wutao1/pegasus-release/rdsn/thirdparty/output/include/thrift/protocol/TBinaryProtocol.tcc:445
#9  0x00000000005eb949 in readString<dsn::blob_string> (str=..., this=0x7fbb58330ce0)
    at /home/wutao1/pegasus-release/rdsn/thirdparty/output/include/thrift/protocol/TBinaryProtocol.tcc:408
[...]

PR #790, opened 2021-07-23
fix: catch exception if unmarshall the rpc request encounters error

levy5307:
In previous implementation, the cluster will crash if we receive a dirty muation log, and will not recover unless the dirty logs are clear. For more details: https://github.com/apache/incubator-pegasus/issues/617
So in pr(https://github.com/XiaoMi/rdsn/pull/863), I changed the way from coredump to throw exception, and we can catch the exception to avoid cluster crash. 

##### Tests 

1. send a dirty request by java client
2. to see whether the perf counter increase
```[replica*app.pegasus*recent_corrupt_write_count@2.4, VOLATILE_NUMBER, 1.00]```
3. to see whether there is a log about this situation.
```pegasus_server_write.cpp:106:on_batched_writes(): pegasus batch writes handler failed, from ={ip}:{port}, exception =no more data to read after end-of-buffer```

### DIFF
--- a/src/server/pegasus_server_write.cpp
+++ b/src/server/pegasus_server_write.cpp
@@ -36,7 +36,9 @@ pegasus_server_write::pegasus_server_write(pegasus_server_impl *server, bool ver
     char name[256];
     snprintf(name, 255, "recent_corrupt_write_count@%s", get_gpid().to_string());
     _pfc_recent_corrupt_write_count.init_app_counter("app.pegasus",
-            name, COUNTER_TYPE_VOLATILE_NUMBER, "statistic the recent corrupt write count");
+                                                     name,
+                                                     COUNTER_TYPE_VOLATILE_NUMBER,
+                                                     "statistic the recent corrupt write count");
 
     init_non_batch_write_handlers();
 }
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: src/server/pegasus_server_write.cpp [hunk 1]
```diff
@@ -51,11 +58,20 @@ int pegasus_server_write::on_batched_write_requests(dsn::message_ex **requests,
         return _write_svc->empty_put(_decree);
     }
 
-    auto iter = _non_batch_write_handlers.find(requests[0]->rpc_code());
-    if (iter != _non_batch_write_handlers.end()) {
-        dassert_f(count == 1, "count = {}", count);
-        return iter->second(requests[0]);
+    try {
```
Smityz, line 61: Why use `try ... catch ..` rather than `if ... else ...`

levy5307: What to do here with `if ... else ...`?

src/server/pegasus_server_write.cpp
```diff
@@ -51,11 +58,20 @@ int pegasus_server_write::on_batched_write_requests(dsn::message_ex **requests,
...
+        auto iter = _non_batch_write_handlers.find(requests[0]->rpc_code());
+        if (iter != _non_batch_write_handlers.end()) {
+            dassert_f(count == 1, "count = {}", count);
```
hycdong, line 64: How about add gpid in logs in this pr?

hycdong (approved):

Shuo-Jia (approved):

PR #792, opened 2021-07-23
feat: add an env to enable or disable block cache of an app

ZhongChaoqiang:
Sometimes, there are some apps of different business in a cluster.Some apps  want to enable block cache, others my be not.
So we need add an env to enable or disable block cache of an app.

##### Related changes

### DIFF
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -2558,6 +2558,27 @@ void pegasus_server_impl::update_rocksdb_iteration_threshold(
     }
 }
 
+void pegasus_server_impl::update_rocksdb_block_cache_enabled(
+    const std::map<std::string, std::string> &envs)
+{
+    bool cache_enabled = _data_cf_rd_opts.fill_cache;
+    auto find = envs.find(ROCKSDB_BLOCK_CACHE_ENABLED);
+    if (find != envs.end()) {
+        if (!dsn::buf2bool(find->second, cache_enabled)) {
+            derror_replica("{}={} is invalid.", find->first, find->second);
+            return;
+        }
+    }
+
+    if (_data_cf_rd_opts.fill_cache != cache_enabled) {
+        ddebug_replica("update app env[{}] from \"{}\" to \"{}\" succeed",
+                       ROCKSDB_BLOCK_CACHE_ENABLED,
+                       _data_cf_rd_opts.fill_cache,
+                       cache_enabled);
+        _data_cf_rd_opts.fill_cache = cache_enabled;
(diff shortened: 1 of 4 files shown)


ZhongChaoqiang:
The issue is apache/incubator-pegasus#795

levy5307 (dismissed):

src/server/pegasus_server_impl.cpp
```diff
@@ -2558,6 +2558,27 @@ void pegasus_server_impl::update_rocksdb_iteration_threshold(
...
+{
+    bool cache_enabled = _data_cf_rd_opts.fill_cache;
+    auto find = envs.find(ROCKSDB_BLOCK_CACHE_ENABLED);
+    if (find != envs.end()) {
+        if (!dsn::buf2bool(find->second, cache_enabled)) {
+            derror_replica("{}={} is invalid.", find->first, find->second);
+            return;
+        }
+    }
+
```
hycdong, line 2572: If this `ROCKSDB_BLOCK_CACHE_ENABLED` is not found in app_envs,  `cache_enabled` should be set as default value, but in your implementation, this value won't be updated.
For example, this default value is `true`. Firstly, admin user update it as `false` through adding an envs. Then **this app_envs is deleted**, the value is expected to be `true`, not `false`.

src/server/pegasus_server_impl.cpp
```diff
@@ -2558,6 +2558,28 @@ void pegasus_server_impl::update_rocksdb_iteration_threshold(
...
+    bool cache_enabled = true;
+    auto find = envs.find(ROCKSDB_BLOCK_CACHE_ENABLED);
+    if (find != envs.end()) {
+        if (!dsn::buf2bool(find->second, cache_enabled)) {
+            derror_replica("{}={} is invalid.", find->first, find->second);
+            return;
+        }
+    }
+
+    if (_data_cf_rd_opts.fill_cache != cache_enabled) {
```
levy5307, line 2574: I think you can remove the `if` statement:
```
        ddebug_replica("update app env[{}] from \"{}\" to \"{}\" succeed",
                       ROCKSDB_BLOCK_CACHE_ENABLED,
                       _data_cf_rd_opts.fill_cache,
                       cache_enabled);
        _data_cf_rd_opts.fill_cache = cache_enabled;
```

levy5307: Yes, you are right. Remain unchanged

levy5307 (approved):

hycdong (approved):

acelyc111 (approved):

PR #817, opened 2021-09-18
feat: add a config of periodic_compaction_seconds in rocksdb

ZhongChaoqiang:
sometimes we don't need periodic compaction of rocksdb, so we add a config of periodic_compaction_seconds.
For details of periodic compacion to see https://github.com/facebook/rocksdb/wiki/Leveled-Compaction#periodic-compaction


src/server/pegasus_server_impl_init.cpp
```diff
@@ -527,6 +527,12 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
     _key_ttl_compaction_filter_factory = std::make_shared<KeyWithTTLCompactionFilterFactory>();
     _data_cf_opts.compaction_filter_factory = _key_ttl_compaction_filter_factory;
 
+    _data_cf_opts.periodic_compaction_seconds =
+        dsn_config_get_value_uint64("pegasus.server",
+                                    "rocksdb_periodic_compaction_seconds",
+                                    30 * 24 * 60 * 60,
```
levy5307, line 533: I think the default value should be UINT64_MAX-1, which is default value in rocksdb

Smityz:
Can `periodic_compaction_seconds` be set by remote_command or `FLAGS`, so that we can change it dynamiclly?

levy5307:
> Can `periodic_compaction_seconds` be set by remote_command or `FLAGS`, so that we can change it dynamiclly?

I am afraid that RocksDB don't support dynamic change. But I think setting it by `FLAGS`is a good idea.

src/server/pegasus_server_impl_init.cpp
ZhongChaoqiang: The default value of periodic_compaction_seconds is 30 days.It is turned on in column_family.
Below is code of rocksdb.
```
  const uint64_t kAdjustedPeriodicCompSecs = 30 * 24 * 60 * 60;

  // Turn on periodic compactions and set them to occur once every 30 days if
  // compaction filters are used and periodic_compaction_seconds is set to the
  // default value.
  if (result.compaction_style != kCompactionStyleFIFO) {
    if ((result.compaction_filter != nullptr ||
         result.compaction_filter_factory != nullptr) &&
        result.periodic_compaction_seconds == kDefaultPeriodicCompSecs &&
        is_block_based_table) {
      result.periodic_compaction_seconds = kAdjustedPeriodicCompSecs;
    }
  }
```

levy5307: In my opinion, the default value is  kDefaultPeriodicCompSecs(UINT64_MAX -1). And rocksdb set it to 30days when you use default value. maybe someday rocksdb will change it to 20days or 10days, that would lead to inconsistencies

ZhongChaoqiang: @levy5307 the default value has been changed to 0

acelyc111 (approved):
Thanks for the contribution!

levy5307 (approved):

PR #828, opened 2021-10-18
fix: coredump when table name contains '_' and prometheus is enabled

padmejin:
After we enable prometheus(set "perf_counter_sink" to "prometheus"), pegasus ended up core dumped after restart.
We debuged the core, found out it was cored while analysing table name that contains "_".
This PR fixes the bug by changing the order of formating table names.

Prometheus does not support some special charactor in metric name.
This original code first formated the table name by replacing special charactor to "_" or "@".
Then, it try to parse gpid as prometheus labels by spliting the foramted name by ":" or "_".
Unfortunately, tables that already contains "_" may results in unexpected group result, which eventually ended up core.

##### Tests 

- Unit test
- Integration test
- No code

##### Code changes

- Has persistent data change

##### Side effects

- Increased code complexity

##### Related changes

### DIFF
--- a/src/reporter/pegasus_counter_reporter.cpp
+++ b/src/reporter/pegasus_counter_reporter.cpp
@@ -234,8 +234,9 @@ void pegasus_counter_reporter::update()
             const dsn::perf_counters::counter_snapshot &cs) {
             std::string metrics_name = cs.name;
 
-            // split metric_name like "collector*app.pegasus*app_stat_multi_put_qps@1.0.p999" or
-            // "collector*app.pegasus*app_stat_multi_put_qps@1.0"
+            // Splits metric_name like:
+            //   "collector*app.pegasus*app_stat_multi_put_qps@1.0.p999"
+            //   "collector*app.pegasus*app_stat_multi_put_qps@1.0"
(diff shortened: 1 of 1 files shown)


padmejin:
As this picture shows below, when core dumps it was trying to analyze metric naming `replica_econ_replica_table_level_RPC_RRDB_RRDB_REMOVE_latency_ns_:impala_historical_profile_p999`, where `impala_historical_profile` is a table we created earlier that contains multiple underscore.

wecom-temp-979f68e2dd409e3c69aadc62fab16057

As the code shown below on line 256, it tries to split metrics_name by `:` first, then split the last part by `_` and put the splited tokens in an array that length is 3. Because this table already contains underscore, there are more than 3 tokens, so it got core dumped after at line 259 trying to set the fourth element of the array.
wecom-temp-b491dd1192d580d28b13da889d23196e

This patch fix it by switching these format order: 
Firstly, it try to parse gpid as prometheus labels by spliting the foramted name by `:` or `@`.
Then, it  formated the table name by replacing special charactor to `_` or "@".

In this order tables that contains `_` can be parsed expectedly in the first step.

PR #829, opened 2021-10-19
feat: support read throttling by size

Smityz:
Ref: https://github.com/apache/incubator-pegasus/issues/830

### DIFF
--- a/src/base/pegasus_const.h
+++ b/src/base/pegasus_const.h
@@ -68,4 +68,6 @@ extern const std::string ROCKSDB_BLOCK_CACHE_ENABLED;
 extern const std::string SPLIT_VALIDATE_PARTITION_HASH;
 
 extern const std::string USER_SPECIFIED_COMPACTION;
+
+extern const std::string READ_SIZE_THROTTLING;
 } // namespace pegasus
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -31,6 +31,7 @@
 #include <dsn/dist/fmt_logging.h>
 #include <dsn/dist/replication/replication.codes.h>
 #include <dsn/utility/flags.h>
+#include <dsn/utils/token_bucket_throttling_controller.h>
 
 #include "base/pegasus_key_schema.h"
 #include "base/pegasus_value_schema.h"
@@ -52,6 +53,12 @@ DSN_DEFINE_int32("pegasus.server",
                  hotkey_analyse_time_interval_s,
                  10,
                  "hotkey analyse interval in seconds");
+
+DSN_DEFINE_int32("pegasus.server",
+                 read_throttling_by_size_request_count,
+                 1,
+                 "the default value of throttling controller of read_throttling_by_size");
+DSN_TAG_VARIABLE(read_throttling_by_size_request_count, FT_MUTABLE);
 
 static std::string chkpt_get_dir_name(int64_t decree)
 {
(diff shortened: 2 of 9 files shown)


src/server/pegasus_server_impl.cpp
```diff
@@ -53,6 +54,12 @@ DSN_DEFINE_int32("pegasus.server",
                  10,
                  "hotkey analyse interval in seconds");
 
+DSN_DEFINE_int32("pegasus.server",
+                 read_throttling_by_size_request_count,
+                 1,
+                 "the default value of throttling controller of read_throttling_by_size");
```
Shuo-Jia, line 60: the config seem to be useless, which case need change the config?

levy5307 (dismissed):

src/server/pegasus_server_impl.cpp
```diff
@@ -274,6 +281,12 @@ void pegasus_server_impl::on_get(get_rpc rpc)
     resp.partition_index = _gpid.get_partition_index();
     resp.server = _primary_address;
 
+    if (!_read_size_throttling_controller->get_token(FLAGS_read_throttling_by_size_request_count)) {
```
Shuo-Jia, line 284: exactly should getToken(rpc.size)

src/server/pegasus_server_impl.cpp
```diff
@@ -274,6 +281,12 @@ void pegasus_server_impl::on_get(get_rpc rpc)
...
+        rpc.error() = dsn::ERR_BUSY;
+        _counter_recent_read_throttling_reject_count->increment();
```
Shuo-Jia, line 286: I suggest the reject count can be distinct the `size` and `qps`, which we can get the request why exceed the throld

src/server/pegasus_server_impl.cpp
Smityz: Actually, it has nothing to do with any size, it only checks if the token basket is empty. If it's empty, then reject all requests

Shuo-Jia: You have define the default value in get_token, no need pass the FLAGS_read_throttling_by_size_request_count

Smityz: it can be remotely changed if necessary

src/server/pegasus_server_impl.cpp
Smityz: ref https://github.com/XiaoMi/rdsn/blob/master/src/replica/replica_throttle.cpp#L56
For compatibility,  I think it's ok to combine them. We can distinguish `size` and `qps` from other perf_counters easily

src/server/pegasus_server_impl.cpp
Shuo-Jia: You just get current the avaliable token, so which case need change? or why change

src/server/pegasus_server_impl.cpp
Shuo-Jia: Next pr consider it, now is ok

src/base/pegasus_const.h
```diff
@@ -68,4 +68,6 @@ extern const std::string ROCKSDB_BLOCK_CACHE_ENABLED;
 extern const std::string SPLIT_VALIDATE_PARTITION_HASH;
```
levy5307, line 68: Could you add some tests for this pull request?

Smityz: I will add some function tests for all read/write throttle tests in my next PR

levy5307 (approved):

Shuo-Jia (approved):

PR #836, opened 2021-11-08
fix: hotkey_collector bug and two function tests

cauchy1988:
fix two ut test;

(1)hotkey_collector_test.cpp 
if all the 'coarse_collector.capture_data' async logic not run, but "coarse_collector.analyse_data(result)" executed will cause unit test fail; this phenomenon has appeared many times in my laptop computer and build machine in my company;

(2)test_split.cpp
variable 'count' should be initialized, if not will cause unit test fail randomly 

##### Tests 
Unit test

##### Code changes
only unit test related code change

##### Side effects
only have impact on unit test

##### Related changes
no relate changes

### DIFF
--- a/src/server/hotkey_collector.cpp
+++ b/src/server/hotkey_collector.cpp
@@ -84,6 +84,7 @@ find_outlier_index(const std::vector<uint64_t> &captured_keys, int threshold, in
             hot_value = captured_keys[i];
         }
     }
+    std::cout << std::endl;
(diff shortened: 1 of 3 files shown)


hycdong (dismissed):
I think the title `fix: 2 ut test` is too short, I suggest updating it.

levy5307 (approved):

Smityz (approved):

hycdong (dismissed):

Smityz (approved):

hycdong (approved):

PR #857, opened 2021-12-14
fix: fix bug in split function test

levy5307:

local variable not initialized in split function test, which will produce test failure.

##### Tests 

- No code


levy5307:
Fixed at #836 . Closed here

Issue #851, opened 2021-12-07
Feature: support table online migration through backup and bulk load


hycdong:
In current implementation, table online migration is supported by duplication, this issue raise a new plan to implement it through cold backup and bulk load. In RocksDB, bulk load support a parameter called `ingest_behind`. If `ingest_behind=false`, the bulk loaded data is newest data, otherwise, the data is oldest. 
As a result, we raise backup and bulk load to implement table online migration, the migration step is:
1. client write old table and new table together
2. old table create a backup
3. convert backup into sst files
4. bulk load sst files with  `ingest_behind=true`
5. client stop write old table

The new online migration plan includes following code:
1. Add a new app_envs to control rocksdb ingest_behind option (immutable-option)
2. Add allow_ingest_behind for start bulk load request
3. Add a script to update rocksdb ingest_behind option (by close and reopen rocksdb)

PR #862, opened 2021-12-31
feat(online_migration): part3 - support ingestion_behind for pegasus

hycdong:
As apache/incubator-pegasus#851 show, we decide to support online migration.

This pull request support it for pegasus, including:
1. Parse `allow_ingest_behind` from app->envs, and set rocksdb `allow_ingest_behind`, reference function `start` and function `update_allow_ingest_behind`.
2. Support `ingest_behind` for ingestion, reference `ingest_files` functions.
3. Update start_bulk_load cmd to support `ingest_behind`
4. Add bulk_load_ingest_behind function test

### DIFF
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -1487,15 +1487,13 @@ void pegasus_server_impl::on_clear_scanner(const int64_t &args) { _context_cache
             // only be initialized with default values when calling 'LoadLatestOptions', see
             // 'rocksdb/utilities/options_util.h'.
             reset_usage_scenario_options(loaded_data_cf_opts, &tmp_data_cf_opts);
-            update_allow_ingest_behind(envs, loaded_db_opt.allow_ingest_behind);
-            _db_opts.allow_ingest_behind = _allow_ingest_behind;
+            _db_opts.allow_ingest_behind = parse_allow_ingest_behind(envs);
         }
     } else {
         // When create new DB, we have to create a new column family to store meta data (meta column
         // family).
         _db_opts.create_missing_column_families = true;
-        update_allow_ingest_behind(envs, false);
-        _db_opts.allow_ingest_behind = _allow_ingest_behind;
+        _db_opts.allow_ingest_behind = parse_allow_ingest_behind(envs);
     }
 
     std::vector<rocksdb::ColumnFamilyDescriptor> column_families(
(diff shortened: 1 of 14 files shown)


src/server/rocksdb_wrapper.cpp
```diff
@@ -180,9 +180,12 @@ int rocksdb_wrapper::write_batch_delete(int64_t decree, dsn::string_view raw_key
 
 void rocksdb_wrapper::clear_up_write_batch() { _write_batch->Clear(); }
 
-int rocksdb_wrapper::ingestion_files(int64_t decree, const std::vector<std::string> &sst_file_list)
+int rocksdb_wrapper::ingestion_files(int64_t decree,
+                                     const std::vector<std::string> &sst_file_list,
+                                     bool ingest_behind)
```
levy5307, line 185: const bool ingest_behind

src/server/rocksdb_wrapper.cpp
```diff
@@ -180,9 +180,12 @@ int rocksdb_wrapper::write_batch_delete(int64_t decree, dsn::string_view raw_key
 
 void rocksdb_wrapper::clear_up_write_batch() { _write_batch->Clear(); }
 
-int rocksdb_wrapper::ingestion_files(int64_t decree, const std::vector<std::string> &sst_file_list)
+int rocksdb_wrapper::ingestion_files(int64_t decree,
```
levy5307, line 183: What about name it `ingest_files`?

src/server/pegasus_server_impl.cpp
```diff
@@ -2695,6 +2699,22 @@ void pegasus_server_impl::update_user_specified_compaction(
     }
 }
 
+void pegasus_server_impl::update_allow_ingest_behind(const std::map<std::string, std::string> &envs,
+                                                     bool default_value)
+{
+    const auto &iter = envs.find(ROCKSDB_ALLOW_INGEST_BEHIND);
+    if (iter == envs.end()) {
+        _allow_ingest_behind = false;
```
levy5307, line 2707: I think you can set `_db_opts.allow_ingest_behind` directly

Smityz (dismissed):

src/server/pegasus_server_impl.cpp
levy5307, line 2707: Why didn't you set it to default_value?

hycdong: If `ROCKSDB_ALLOW_INGEST_BEHIND` can't be found in app_envs, it should be set as false.
`default_value` only be used when app ROCKSDB_ALLOW_INGEST_BEHIND is invalid bool, we set it as default_value. 
If the db exists before, default_value is the lastest option value, if not, default_value is false. I will update this function comment.

src/server/rocksdb_wrapper.cpp
hycdong: Done

src/server/pegasus_server_impl.cpp
hycdong: Done

levy5307 (approved):

Smityz (approved):

PR #867, opened 2022-01-06
feat(online_migration): part4 - add script to update table ingest_behind

hycdong:
As #851 show, we decide to support online migration. 

This pull request adds a script to update table envs and reopen all partitions to make rocksdb option updated.
Usage：`./scripts/pegasus_update_ingest_behind.sh <cluster-name> <cluster-meta-list> <app_name> <ingestion_behind>`
For example：
```
./scripts/pegasus_update_ingest_behind.sh onebox 127.0.0.1:34601,127.0.0.1:34602 temp false
```

### DIFF
--- a/scripts/pegasus_rebalance_cluster.sh
+++ b/scripts/pegasus_rebalance_cluster.sh
@@ -49,13 +49,6 @@ pwd="$( cd "$( dirname "$0"  )" && pwd )"
 shell_dir="$( cd $pwd/.. && pwd )"
 cd $shell_dir
 
-source ./scripts/minos_common.sh
-find_cluster $cluster
-if [ $? -ne 0 ]; then
-  echo "ERROR: cluster \"$cluster\" not found"
-  exit 1
-fi
-
 echo "UID=$UID"
 echo "PID=$PID"
 echo "Start time: `date`"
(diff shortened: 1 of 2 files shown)


levy5307 (approved):

scripts/pegasus_rebalance_cluster.sh
```diff
@@ -49,13 +49,6 @@ pwd="$( cd "$( dirname "$0"  )" && pwd )"
 shell_dir="$( cd $pwd/.. && pwd )"
 cd $shell_dir
 
-source ./scripts/minos_common.sh
-find_cluster $cluster
-if [ $? -ne 0 ]; then
-  echo "ERROR: cluster \"$cluster\" not found"
-  exit 1
-fi
-
```
Shuo-Jia, line 58: why delete？

hycdong: The rebalance script doesn't use any command of minos, only remains the cluster name check. Current scripts who call it all do cluster name check before, deleting it will not affect current scripts. 
`add_ingest_behind_script` call rebalance but won't rely on minos, so I delete it.

Shuo-Jia (approved):

Issue #900, opened 2022-01-29
Pegasus compile failed on Ubuntu after supporting build on MacOS


hycdong:
## Bug Report
Gcc 5.4.0, Cmake 3.13.2
Use Pegasus master(59dc6bd46805e18f405c3a03ebba5a8a591ce029) and rdsn master(2dd67e0885e36f964633c17afc7d91086517dfaf), complie failed, the error like following.

## Complie failed errors
```
Building...
Scanning dependencies of target pegasus_base
[  0%] Building CXX object base/CMakeFiles/pegasus_base.dir/pegasus_const.cpp.o
[  1%] Building CXX object base/CMakeFiles/pegasus_base.dir/pegasus_utils.cpp.o
[  2%] Building CXX object base/CMakeFiles/pegasus_base.dir/rrdb_types.cpp.o
[  2%] Building CXX object base/CMakeFiles/pegasus_base.dir/value_schema_manager.cpp.o
[  3%] Building CXX object base/CMakeFiles/pegasus_base.dir/value_schema_v0.cpp.o
[  4%] Building CXX object base/CMakeFiles/pegasus_base.dir/value_schema_v1.cpp.o
[  4%] Building CXX object base/CMakeFiles/pegasus_base.dir/value_schema_v2.cpp.o
Scanning dependencies of target pegasus_client_impl_objects
[  5%] Building CXX object client_lib/CMakeFiles/pegasus_client_impl_objects.dir/client_factory.cpp.o
[  5%] Building CXX object client_lib/CMakeFiles/pegasus_client_impl_objects.dir/mutation.cpp.o
[  6%] Building CXX object client_lib/CMakeFiles/pegasus_client_impl_objects.dir/pegasus_client_factory_impl.cpp.o
[  7%] Building CXX object client_lib/CMakeFiles/pegasus_client_impl_objects.dir/pegasus_client_impl.cpp.o
[  7%] Building CXX object client_lib/CMakeFiles/pegasus_client_impl_objects.dir/pegasus_scanner_impl.cpp.o
In file included from /home/heyuchen/work/pegasus_github/pegasus/DSN_ROOT/include/dsn/c/api_layer1.h:33:0,
                 from /home/heyuchen/work/pegasus_github/pegasus/DSN_ROOT/include/dsn/service_api_c.h:85,
                 from /home/heyuchen/work/pegasus_github/pegasus/DSN_ROOT/include/dsn/service_api_cpp.h:38,
[...]

PR #901, opened 2022-02-07
fix(build): Fix '__FILENAME__ was not declared' bug

acelyc111:
ref: https://github.com/apache/incubator-pegasus/issues/900


levy5307 (approved):

hycdong (approved):
Build succeed after adding this patch in my environment, LGTM

Issue #865, opened 2022-01-05
Feature: support to get/set the replication factor of each table


empiredan:
# 1. Motivation

Replication factor (RF), is the number of nodes (i.e. replica servers) where each partition of a table is replicated. In Apache Pegasus, we also call it *max replica count* or *replica count*.

Initially, the minimum RF was 3; Then, in order to reduce the cost for our users, in https://github.com/XiaoMi/rdsn/pull/932 we've supported RF=1 and RF=2, though it's not reliable enough for most distributed systems while RF is 1 or 2. 

In other words, the minimum RF has become 1. Imagine a scenario that we had 2 replica servers in a cluster where a table can only be created with RF=1 or RF=2. As the volume of data continues to grow, the capacity of  this cluster is not enough; Also, it's not scalable and reliable for the 2 replica servers.

Now we would add another 2 nodes into this cluster and increase the RF of each table to 3 to improve the reliability and scalability. Therefore, besides that RF is specified while creating table, it should be changeable after the table is created. To make RF settable dynamically, appropriate commands should be added into the management tool such as Pegasus Shell to query or update the RF.

# 2. Limitation

Previously we've restricted the RF while creating table in https://github.com/XiaoMi/rdsn/pull/963. These restriction rules will also be applied as the prerequisites before the RF could be updated successfully.

# 3. Usage

As is described above, the RF could be queried or updated by Pegasus Shell. Here is the patterns of the commands.

It should be noted that to follow the naming convention in Pegasus, `replica count` is used in the command to represent the RF.

## 3.1 Query the RF

```shell
get_replica_count <table_name>
```

After this command is executed, the RF of the specified table will be printed if it's successful; Otherwise the error will be printed.

## 3.2 Update the RF

```shell
set_replica_count <table_name> <replica_count>
```

Since this command will update the RF of a table, in case of incorrect operations a prompt will be printed immediately after the command is executed to ask if the administrator chooses to continue to update the RF. Then, it will tell the command is finished successfully, or, similarly with `get_replica_count`,  the error will be printed.

# 4. Implementation

Here we only illustrate the process of `set_replica_count`, since `get_replica_count` is quite simple.

## 4.1 Basic process

The whole process contains the following steps:
* Administrator inputs and confirms `set_replica_count` command by Pegasus Shell;
* The `set_replica_count` request is sent to Primary Meta Server;
* Meta Server checks whether the requested RF is valid, whether the table exists, etc.;
[...]

PR #914, opened 2022-02-18
feat(update_replication_factor#10): support to get/set the replication factor of each table by shell and add max_reserved_dropped_replicas to config file

empiredan:
https://github.com/apache/incubator-pegasus/issues/865


src/shell/commands/table_management.cpp
```diff
@@ -886,3 +886,108 @@ bool clear_app_envs(command_executor *e, shell_context *sc, arguments args)
...
+        "set the replica count of app({}) to {}", escaped_app_name, new_max_replica_count));
+    if (!confirm_unsafe_command(action.c_str())) {
+        return true;
+    }
+
+    auto err_resp = sc->ddl_client->set_max_replica_count(app_name, new_max_replica_count);
+    auto err = err_resp.get_error();
+    const auto &resp = err_resp.get_value();
+
+    if (err.is_ok()) {
```
levy5307, line 951: dsn_likely

src/shell/command_utils.cpp
```diff
@@ -45,3 +45,35 @@ bool validate_ip(shell_context *sc,
     err_info = fmt::format("invalid ip:port={}, can't find it in the cluster", ip_str);
     return false;
 }
+
+bool confirm_unsafe_command(const char *action)
```
levy5307, line 49: const char *action ==> const std::string &action

src/shell/command_utils.cpp
```diff
@@ -45,3 +45,35 @@ bool validate_ip(shell_context *sc,
     err_info = fmt::format("invalid ip:port={}, can't find it in the cluster", ip_str);
     return false;
 }
+
+bool confirm_unsafe_command(const std::string &action)
+{
+    const int max_attempts = 5;
+    int attempts = 0;
+    for (; attempts < max_attempts; ++attempts) {
```
levy5307, line 53: ``` for (int attempts; attempts < max_attempts; ++attempts) ```

src/shell/command_utils.cpp
```diff
@@ -45,3 +45,35 @@ bool validate_ip(shell_context *sc,
...
+        int choice = fgetc(stdin);
+        int len = 0;
+        for (int c = choice; c != '\n' && c != EOF; ++len) {
+            c = fgetc(stdin);
+        }
+        if (len != 1) {
+            continue;
+        }
+
+        if (choice == 'Y') {
```
levy5307, line 68: ```if (choice == 'Y' || choice == 'y')```

acelyc111, line 68: If 'Y' is not the default selection, use 'y'

src/shell/commands/table_management.cpp
```diff
@@ -886,3 +888,105 @@ bool clear_app_envs(command_executor *e, shell_context *sc, arguments args)
...
+        fmt::print(stdout,
+                   "the replica count of app({}) is {}\n",
+                   escaped_app_name,
+                   resp.max_replica_count);
+    } else {
+        fmt::print(stdout, "get replica count of app({}) failed: {}\n", escaped_app_name, err);
+    }
+
+    return true;
```
acelyc111, line 918: If we use the command in some 3rd party tools, we often depend on the return value and output data streams(stdout or stderr), so it's better to return false and output in stderr.

src/shell/commands/table_management.cpp
```diff
@@ -886,3 +888,105 @@ bool clear_app_envs(command_executor *e, shell_context *sc, arguments args)
...
+    if (err.is_ok()) {
+        err = dsn::error_s::make(resp.err);
+    }
+
+    std::string escaped_app_name(pegasus::utils::c_escape_string(app_name));
+    if (err.is_ok()) {
+        fmt::print(stdout,
+                   "the replica count of app({}) is {}\n",
+                   escaped_app_name,
+                   resp.max_replica_count);
```
acelyc111, line 913: It's better to output also in JSON, so we can parse easily in 3rd party tools.

src/shell/commands/table_management.cpp
```diff
@@ -886,3 +888,105 @@ bool clear_app_envs(command_executor *e, shell_context *sc, arguments args)
...
+
+bool set_max_replica_count(command_executor *e, shell_context *sc, arguments args)
+{
+    if (args.argc < 3) {
+        return false;
+    }
+
+    int new_max_replica_count;
+    if (!dsn::buf2int32(args.argv[2], new_max_replica_count)) {
+        fmt::print(stderr, "parse {} as replica count failed\n", args.argv[2]);
```
acelyc111, line 929:
```suggestion
        fmt::print(stderr, "parse '{}' as replica count failed\n", args.argv[2]);
```

src/shell/commands/table_management.cpp
```diff
@@ -886,3 +888,105 @@ bool clear_app_envs(command_executor *e, shell_context *sc, arguments args)
...
+                       resp.old_max_replica_count,
+                       new_max_replica_count,
+                       resp.hint_message);
+        } else {
+            fmt::print(stdout,
+                       "set replica count of app({}) from {} to {} successfully\n",
+                       escaped_app_name,
+                       resp.old_max_replica_count,
+                       new_max_replica_count);
+        }
```
acelyc111, line 968: Is it necesssary to split these two cases? There is little difference.

src/shell/commands/table_management.cpp
```diff
@@ -886,3 +888,105 @@ bool clear_app_envs(command_executor *e, shell_context *sc, arguments args)
...
+
+        if (resp.old_max_replica_count > 0) {
+            fmt::print(stdout,
+                       "set replica count of app({}) from {} to {} failed: {}\n",
+                       escaped_app_name,
+                       resp.old_max_replica_count,
+                       new_max_replica_count,
+                       error_message);
+        } else {
+            fmt::print(stdout,
```
acelyc111, line 984: How this case happend?

acelyc111:
Could you please add this command also in admin-cli ?

src/shell/commands/table_management.cpp
empiredan: > Is it necesssary to split these two cases? There is little difference.

OK, I can merge the two cases.

src/shell/commands/table_management.cpp
empiredan: > How this case happend?

For example, if a table does not exist, it will not has an old max_replica_count.

empiredan:
> Could you please add this command also in admin-cli ?

OK, sure.

src/shell/commands/table_management.cpp
empiredan: > It's better to output also in JSON, so we can parse easily in 3rd party tools.

OK, I'll add `-j` and `--json` for `get_replica_count`, for example: {"max_replica_count": 3}.

acelyc111: You can use `::dsn::utils::table_printer` to wrap these code.

empiredan: > You can use `::dsn::utils::table_printer` to wrap these code.

OK.

acelyc111 (approved):

src/shell/commands/table_management.cpp
```diff
@@ -886,3 +888,106 @@ bool clear_app_envs(command_executor *e, shell_context *sc, arguments args)
     }
     return true;
 }
+
+bool get_max_replica_count(command_executor *e, shell_context *sc, arguments args)
+{
```
foreverneverer, line 893: `ls` command can show the max_replica_count, here add the command has some different?

empiredan: > `ls` command can show the max_replica_count, here add the command has some different?

Actually there is no special reason for adding `get_max_replica_count` currently. This interface is designed just for a complete get/set semantics.

foreverneverer (approved):

Issue #892, opened 2022-01-24
Feature: Enhance the ease of use and efficiency of duplication


foreverneverer:
### Duplication has some shortcoming：
![选区_999(788)](https://user-images.githubusercontent.com/23136769/150751109-919b867c-3906-4d86-8def-5788ac0f002f.png)

- **It depends remote filesystem to sync the checkpoint:** 
        It leads to complexity in use, which make you have to support HDFS or FDS, and must start duplicate in three steps: 1. upload checkpoint in master 2. restore checkpoint in follower 3. start sync plog data in master.
- **The synchronization of plog data only sends a single mutation at each RPC:**
        The rate of sync is slow and cause the plog pile up in sometimes.

### The issue request will hope enhance it:
![选区_999(789)](https://user-images.githubusercontent.com/23136769/150752425-c7fd1db1-baea-4a79-8989-f35d86ae794f.png)

- Use `learn` to replace `backup-restore` to sync checkpoint
- Use `batch-send` to replace `single-send` to sync plog
- Connect the above two processes to automate them, and the establishment of duplication can be completed with one command

Shuo-Jia:
### The flow chart is as follow: 
![选区_999(791)](https://user-images.githubusercontent.com/23136769/150759571-0323797c-5110-4432-ba51-8a5bdba37b70.png)

1. master meta: receive create dup table request and mark current `max_plog_commit_decree` as start plog point
2. master meta: init the dup status as `DS_PREPARE`,  which means master replica need prepare latest_checkpoint  for duplicating
3. master replica: generate latest checkpoint , which require `laster_durable_decee >= max_plog_commit_decree`
4. master replica: update all checkpoint has been prepared and response master meta
5. master meta: change to `DS_APP`, which means all checkpoints have prepared and follower can duplicate it
6. master meta: send create table request to follower meta, and require the table init with `master replica checkpoint`(like restore progress which init table with `backup replica checkpoint`)
7. follower meta: create dup table with dup info
8. follower replica: start open or init replica 
9. follower replica: duplicate checkpoint from master replica before init replica, and load replica using duplicated data
10. master meta: check the follower table if create completed periodically
11. master meta: if completed, change to `DS_LOG`, which means master replica can send plog data to follower table
12. master replica: start pipeline to batch send plog data.

neverchanje:
The motivation is fair, lgtm.

cauchy1988:
[...]

PR #919, opened 2022-03-02
feat(dup_enhancement#18):  support batch send log on master side and handle it on follower side

Shuo-Jia:
# Ref-Issue

# Change
- support batch send plog
- new config
```diff
+ duplicate_log_batch_megabytes = 4
```


src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -42,6 +42,11 @@ namespace replication {
 namespace pegasus {
 namespace server {
 
+DSN_DEFINE_int32("pegasus",
```
levy5307, line 45: DSN_DEFINE_uint32

hycdong (approved):

Issue #887, opened 2022-01-20
code refactor


levy5307:
## Feature Request
 
Code refactor to make code cleaner

PR #921, opened 2022-03-03
refactor: make ctor/dtor function of singleton classes to be private

levy5307:

make ctor functions of singleton to be private, to avoid creating more than one object for singleton classes

https://github.com/apache/incubator-pegasus/issues/887


hycdong (dismissed):

src/test/bench_test/config.h
```diff
@@ -50,6 +48,12 @@ struct config : public ::dsn::utils::singleton<config>
     uint64_t seed;
     // Default environment suitable for the current operating system
     rocksdb::Env *env;
+
+private:
+    config();
+    ~config() = default;
```
Smityz, line 54: should we expose a `delete` interface?

levy5307: Sorry I have no opinion about what you mean

hycdong (approved):

Shuo-Jia (approved):

PR #940, opened 2022-04-08
feat(dup_enhancement#25): add `-s | --sst` argument to support ignoring checkpoint when duplicate

foreverneverer:
# Ref-Issue

### DIFF
--- a/src/shell/commands/duplication.cpp
+++ b/src/shell/commands/duplication.cpp
@@ -39,6 +39,13 @@ bool add_dup(command_executor *e, shell_context *sc, arguments args)
         return false;
     }
 
+    for (const auto &flag : cmd.flags()) {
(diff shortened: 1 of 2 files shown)


hycdong (dismissed):

src/shell/commands/duplication.cpp
```diff
@@ -39,6 +39,13 @@ bool add_dup(command_executor *e, shell_context *sc, arguments args)
         return false;
     }
 
+    for (const auto &flag : cmd.flags()) {
+        if (flag != "s" && flag != "sst") {
```
levy5307, line 43: dsn_unlikely

levy5307 (approved):

hycdong (approved):

Issue #962, opened 2022-04-29
Use admin-cli and pegic to replace origin c++ shell


foreverneverer:

neverchanje:
Please change the privilege settings of the doc and allow it to be publicly readable. :)

foreverneverer:
> Please change the privilege settings of the doc and allow it to be publicly readable. :)

try again?

acelyc111:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-df7f3f23-7fff-8d3c-1c3b-acf62d836802"><div dir="ltr" style="margin-left:0pt;" align="left">

命令 | C++ shell | AdminShell(AdminCli) | UserShell(Pegic) | 备注
-- | -- | -- | -- | --
帮助 | help | help | help |   |   |   |  
版本 | version | version | - |   |   |   |  
退出 | exit | exit | - |   |   |   |  
  |   |   |   |   |   |   |  
更换连接集群 | cc | 不再支持 | 不再支持 |   |   |   |  
指定表 | use | use | use |   |   |   |  
指定超时时间 | timeout | 不再支持 | 不再支持 |   |   |   |  
  |   |   |   |   |   |   |  
集群信息 | cluster_info | cluster-info | - |   |   |   |  
节点信息 | nodes | nodes | - |   |   |   |  
服务器信息 | server_info | server-info | - |   |   |   |  
服务器指标 | server_stat | node-stat | - |   |   |   |  
远程命令 | remote_command | remote-command |   |   |   |   |  
  |   |   |   |   |   |   |  
列出所有表 | ls | ls |   |   |   |   |  
表分片信息 | app | table-partition |   |   |   |   |  
表指标信息 | app_stat | table-stat |   |   |   |   |  
强制刷LOG | flush_log | 待讨论 |   |   |   |   |  
创建表 | create | create |   |   |   |   |  
删除表 | drop | drop |   |   |   |   |  
恢复表 | recall | recall |   |   |   |   |  
获取环境变量 | get_app_env | table-env |   |   |   |   |  
设置环境变量 | set_app_env |   |   |   |   |  
删除环境变量 | del_app_env |   |   |   |   |  
清空环境变量 | clear_app_env |   |   |   |   |  
  |   |   |   |   |   |   |  
设置集群状态 | set_meta_level | meta-level |   |   |   |   |  
查看表状态 | get_meta_level |   |   |   |   |  
[...]

PR #969, opened 2022-05-17
feat(admin-cli): support nodes capacity balance using admin-cli

foreverneverer:
# Related-Issue
https://github.com/apache/incubator-pegasus/issues/962

# Problem
By default, the node capacity of the server needs to be updated every 10 minutes. Therefore, after a partition is migrated completed, the tool cannot immediately obtain the latest capacity distribution. Please adjust the node capacity update interval of the server to speed up the equalization speed. Relevant configurations are as follows:
```diff
- disk_stat_interval_seconds = 600
+ - disk_stat_interval_seconds = 60 # or less

- gc_memory_replica_interval_ms = 600000
+ gc_memory_replica_interval_ms = 60000 # or less
```


hycdong (dismissed):

admin-cli/executor/toolkits/nodesbalancer/migrator.go
```diff
@@ -0,0 +1,253 @@
...
+ */
+
+package nodesbalancer
+
+import (
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/XiaoMi/pegasus-go-client/idl/base"
```
acelyc111, line 27: need update

admin-cli/executor/toolkits/nodesbalancer/balancer.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return err
+	}
+	toolkits.LogInfo("set meta.lb.only_primary_balancer true")
+	err = executor.RemoteCommand(client, session.NodeTypeMeta, "", "meta.lb.only_primary_balancer", []string{"true"})
+	if err != nil {
+		return err
+	}
+	// reset garbage replica clear interval
+	toolkits.LogInfo("set gc_disk_error_replica_interval_seconds 10")
```
acelyc111, line 100: Should we restore these configs after this rebalance operation?

admin-cli/executor/toolkits/diskbalancer/migrator.go
```diff
@@ -68,7 +68,7 @@ func changeDiskCleanerInterval(client *executor.Client, replicaServer string, cl
 }
 
 func getNextMigrateAction(client *executor.Client, replicaServer string, minSize int64) (*MigrateAction, error) {
-	disks, totalUsage, totalCapacity, err := queryDiskCapacityInfo(client, replicaServer)
+	disks, totalUsage, totalCapacity, err := QueryDiskCapacityInfo(client, replicaServer)
```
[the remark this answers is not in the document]
foreverneverer: uppercase is means `public` in go, which can be used for other file

admin-cli/executor/toolkits/nodesbalancer/balancer.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/XiaoMi/pegasus-go-client/session"
+	"github.com/apache/incubator-pegasus/admin-cli/executor"
+	"github.com/apache/incubator-pegasus/admin-cli/executor/toolkits"
+)
+
+// By default, the node capacity of the server needs to be updated every 10 minutes.
```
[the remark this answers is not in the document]
foreverneverer: The config hasn't be supported query, so we just add hint for user

admin-cli/executor/toolkits/nodesbalancer/balancer.go
foreverneverer: Yeah, I will recover it

acelyc111 (approved):

hycdong (approved):

PR #973, opened 2022-05-18
feat: add app drop interface to the java client

GehaFearless:
As https://github.com/apache/incubator-pegasus/issues/972，I add dropApp interface to the java client.


foreverneverer (approved):

java-client/src/test/java/org/apache/pegasus/client/TestAdminClient.java
```diff
@@ -27,7 +27,11 @@
 
 public class TestAdminClient {
   PegasusAdminClientInterface toolsClient;
+  PegasusClientInterface pClient;
```
acelyc111, line 30: pClient is only used in testDropApp, it's not needed to make it as member.

java-client/src/test/java/org/apache/pegasus/client/TestAdminClient.java
```diff
@@ -88,11 +94,36 @@ public void testIsAppHealthyIfTableNotExists() throws PException {
...
+    Assert.assertTrue(isAppHealthy);
+
+    toolsClient.dropApp(appName, tableOpTimeoutMs);
+    PegasusClientInterface pClient = PegasusClientFactory.createClient(this.clientOptions);
+
+    try {
+      pClient.openTable(appName);
+    } catch (PException e) {
+      String msg = e.getMessage();
+      Assert.assertTrue(msg.contains("ERR_OBJECT_NOT_FOUND") && msg.contains("No such table"));
```
acelyc111, line 125: Check that this exception must be threwed.

foreverneverer (approved):

Issue #998, opened 2022-06-07
Bug: different judgement to empty hashkey 


GehaFearless:
### What appeared
When I use multi_set from c++ client, I found that hashkey couldn't be empty.  However, My table data have existed a group data (hashkey="", sortkey="", value=""). 

### What problem 
I find that's different judgement to empty hashkey from client interfaces.
It's "PERR_INVALID_HASH_KEY" for "multi" prefix interfaces when hashkey is empty, but single kv operations is ok. 

### How to solve
I think empty hashkey also should be allowed on "multi" prefix interfaces.

PR #1001, opened 2022-06-13
fix(c-client): allow empty hashkey use multi data operation

GehaFearless:
https://github.com/apache/incubator-pegasus/issues/998


acelyc111:
Add some unit test to check empty hashkey can be set, and can be get then.

PR #1007, opened 2022-06-17
feat(dup_enhancement#26): update duplicate rpc thrift in go-client

foreverneverer:
# Ref-Issue


go-client/session/radmin_rpc_types.go
```diff
@@ -15,7 +15,7 @@
 * KIND, either express or implied.  See the License for the
 * specific language governing permissions and limitations
 * under the License.
-*/
+ */
```
acelyc111, line 18: revert it plz

go-client/session/admin_rpc_types.go
```diff
@@ -15,7 +15,7 @@
 * KIND, either express or implied.  See the License for the
 * specific language governing permissions and limitations
 * under the License.
-*/
+ */
```
acelyc111, line 18: revert it plz

go-client/session/radmin_rpc_types.go
foreverneverer: done

acelyc111 (approved):

acelyc111:
We should unify the IDL *.thrift files, now they are distributed every where, we can do it after rdsn has been merged to Pegasus.

foreverneverer:
> We should unify the IDL *.thrift files, now they are distributed everywhere, we can do it after rdsn has been merged to Pegasus.

Yeah, I have the same plan

hycdong (approved):

Issue #903, opened 2022-02-08
common problem fix


Shuo-Jia:
## Bug Report
This issue is used for tracking some little bug

PR #1017, opened 2022-06-24
build(thirdparty): add thirdparty for centos6 build env 

foreverneverer:
# Ref-Issue
https://github.com/apache/incubator-pegasus/issues/903

https://github.com/XiaoMi/rdsn/pull/979

# Change
https://github.com/XiaoMi/rdsn/pull/979 remove `krb5` and `cyrus-sasl`,  but it's necessary for `CentOS6` when compile `zookeeper`. 

In pack_server.sh, it also need be update

# Note
- if you find `typedef 'zoo_sasl_client_t' re-defination` , that indicates that it conflicts with the system library. Be sure to uninstall the system `cyrus-sasl`
- if you find ` undefined reference to 'zookeeper_init_sasl'`, it because that `zookeeper` compile before `cyrus-sasl`, please remove all `zookeeper` compile result and re-run build


hycdong (approved):

hycdong (dismissed):

acelyc111 (approved):

rdsn/thirdparty/CMakeLists.txt
```diff
@@ -156,6 +156,33 @@ ExternalProject_Add(thrift
         DEPENDS boost
         )
 
+# kerberos
+ExternalProject_Add(krb5
```
acelyc111, line 160: Can't the two libs be installed directly on centos6?

acelyc111 (changes requested):

rdsn/thirdparty/CMakeLists.txt
acelyc111: Why not add centos6 Dockerfile if you insist to support centos6?

foreverneverer: centos6 has stopped supporting and I haven't find the open software yum-repo.  I have build centos6-docker-env in our company internal

foreverneverer: > Can't the two libs be installed directly on centos6?

the version is too old to be compatibility with `zookeeper`

acelyc111: Have you tried https://www.getpagespeed.com/server-setup/how-to-fix-yum-after-centos-6-went-eol? I've test that the two libs can be installed.
But I think there must be some other problems to build on centos6, it's better to create a centos6 Dockerfile to resolve all the problems.
And, for the non-centos6 platforms, we'd better skip build the two packages since we have installed the packages.

foreverneverer: I think if one package has compatibility problem in different env, the best solution is specify unified version(if we can find the version) in thirdparty, but not use different version in different env.

So I think the two package actually should be included in thirdparty though in centos7, unless one day we no longer support centos6

foreverneverer: > Have you tried https://www.getpagespeed.com/server-setup/how-to-fix-yum-after-centos-6-went-eol? I've tested that the two libs can be installed.
> But I think there must be some other problems to build on centos6, it's better to create a centos6 Dockerfile to resolve all the problems.
> And, for the non-centos6 platforms, we'd better skip build the two packages since we have installed the packages.

Great, I will try it

acelyc111: > I think if one package has compatibility problem in different env, the best solution is specify unified version(if we can find the version) in thirdparty, but not use different version in different env.

Agree. But only limit to the platforms we plan to support, centos 6 is EOL on 2020-12, I don't think it's worthy to support it.

foreverneverer: Yeah, I actually don't want support the old version, but our company need it, and maybe need some time to upgrade the centos7.

foreverneverer: > > Have you tried https://www.getpagespeed.com/server-setup/how-to-fix-yum-after-centos-6-went-eol? I've tested that the two libs can be installed.
> > But I think there must be some other problems to build on centos6, it's better to create a centos6 Dockerfile to resolve all the problems.
> 
> Great, I will try it

I have test and the repo can be used

acelyc111 (approved):

acelyc111 (changes requested):
.

acelyc111 (approved):

hycdong (approved):

acelyc111 (approved):

acelyc111 (approved):

Issue #1018, opened 2022-06-24
Feature: PegasusScanner adds hasNext method


WHBANG:
scan is a common interface for storage components, we hope PegasusScanner has two methods, next() and hasNext() like most iterators.

PR #1019, opened 2022-06-24
feat(java-client): Add hasNext() for PegasusScannerInterface

WHBANG:
https://github.com/apache/incubator-pegasus/issues/1018

Added a hasNext interface to the pegasusScanner class.

##### Tests 

- Unit test
Added hasNext unit test to TestScan.java file


java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
```diff
@@ -88,9 +88,27 @@ public PegasusScanner(
     _needCheckHash = needCheckHash;
     _incomplete = false;
     _fullScan = fullScan;
+    _nextItem = null;
```
foreverneverer, line 91: The global variable will cause `thread-safe` problem,  in origin code, I remember it will `lock` before fetch `next`: https://github.com/WHBANG/incubator-pegasus/blob/bd099a9c963de8de4cd8ad4c80ca6bb9653d24cd/java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java#L125

Please check and make sure the `refactor` is `thread-safe`

WHBANG: OK, i will add  _nextItemLock to make _nextItem thread-safe

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
```diff
@@ -88,9 +88,31 @@ public PegasusScanner(
...
+      if (_incomplete) {
+        return false;
+      }
+      if (_rpcRunning && !_encounterError) {
+        return !_promises.isEmpty();
+      } else {
+        _nextItem = next();
+        return _nextItem != null;
+      }
+    }
```
foreverneverer, line 105: I think you don't have to make such complex, just like follow seem to be ok:
```java
public boolean hasNext() throws PException {
        _nextItem = next();
        return _nextItem != null;
}
```

`next()` method include enough logic to judge current status and make sure `thread-safe`. 

That is to say `_incomplete`, `_rpcRunning && !_encounterError` should throw exception but return fasle

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
```diff
@@ -88,9 +88,31 @@ public PegasusScanner(
...
   }
 
   public Pair<Pair<byte[], byte[]>, byte[]> next() throws PException {
+    synchronized (_nextItemLock) {
```
foreverneverer, line 109: shouldn't use new `lock`, you should use `_promisesLock`

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
WHBANG: Yes, but if hasNext is called many times individually,  it will cause the iterator to move,  which is not desirable.

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
WHBANG: good idea. i agree

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
```diff
@@ -88,9 +88,31 @@ public PegasusScanner(
...
+    _nextItem = null;
+  }
+
+  public boolean hasNext() throws PException {
+    synchronized (_nextItemLock) {
+      if (_incomplete) {
+        return false;
+      }
+      if (_rpcRunning && !_encounterError) {
+        return !_promises.isEmpty();
```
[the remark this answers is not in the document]
WHBANG: when the rpc compeleted, the scan is not compeleted, but the key-value is cached on the _kvs and promise queues, obtained in turn through next(), and _promises.removeFirst()

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
foreverneverer: > but if hasNext is called many times individually, it will cause the iterator to move, which is not desirable
 This case is right, so you may need update it like:
```java
public boolean hasNext() throws PException {
          synchronized (_nextItemLock) {
            if (_nextItem != null) {
                 return true;
            }
        }
        _nextItem = next();
        return _nextItem != null;
}
```

foreverneverer (approved):

acelyc111 (approved):

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
```diff
@@ -88,9 +88,27 @@ public PegasusScanner(
...
+    _nextItem = null;
+  }
+
+  public boolean hasNext() throws PException {
+    synchronized (_promisesLock) {
+      if (_nextItem != null) {
+        return true;
+      }
+    }
+    _nextItem = next();
```
acelyc111, line 100: It's not thread safe in some cases since `_nextItem` will be overwritten.

```
time  thread1         thread2
  t1  hasNext         hasNext
  t2  _nextItem = a
  t3                  _nextItem = b
  t4  next -> b
  t5                  next -> nullptr
```

acelyc111 (changes requested):

acelyc111 (approved):

foreverneverer (approved):

acelyc111 (approved):

PR #1035, opened 2022-07-06
fix(update_replication_factor#12): update the mutation_2pc_min_replica_count base the max count of table

foreverneverer:
# Ref-Issue
https://github.com/apache/incubator-pegasus/issues/865

https://github.com/apache/incubator-pegasus/issues/865 support different replica factor, but `mutation_2pc_min_replica_count` still keep `2`, if one table update the factor to `2`, when one replica is down(for example, offline node case),  the table would keep `write reject` status.

So `mutation_2pc_min_replica_count` should be updated base the table factor, this pr should set it `replica_max_count /2  + 1`,  but consider that sometimes user need set the `replica_max_count > 3`, I think the config `mutation_2pc_min_replica_count` is still valid:
1. if user set the `mutation_2pc_min_replica_count > 0`,  the final `2pc_min_replica_count = mutation_2pc_min_replica_count`
2. otherwise, `2pc_min_replica_count = app_max_replica_count <= 2 ? 1 : app_max_replica_count / 2 + 1` which can be make sure the table is alive though more node crush

# Config-Change
if you want use value of `config.ini`, you don't need update it, otherwise:
```diff
- mutation_2pc_min_replica_count = 2
+ mutation_2pc_min_replica_count = 0
```

##### Tests 

- Unit test
```shell
1. start server in test cluster
4. create table with 3 replica count
5. start read/set with 10k qps
6. set replica count = 2
7. kill one node
8. watch the write status
```

##### Related changes


rdsn/src/replica/replica.h
```diff
@@ -497,6 +497,8 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
 
     void update_app_max_replica_count(int32_t max_replica_count);
 
+    int32_t mutation_2pc_min_replica_count();
```
acelyc111, line 500:
```suggestion
    int32_t mutation_2pc_min_replica_count() const;
```

rdsn/src/meta/meta_service.h
```diff
@@ -174,6 +174,8 @@ class meta_service : public serverlet<meta_service>
         return metas.substr(0, metas.length() - 1);
     }
 
+    int32_t mutation_2pc_min_replica_count(int32_t app_max_replica_count);
```
acelyc111, line 177:
```suggestion
    int32_t mutation_2pc_min_replica_count(int32_t app_max_replica_count) const;
```

rdsn/src/common/replication_common.h
```diff
@@ -117,6 +117,11 @@ class replication_options
     ~replication_options();
 
     void initialize();
+
+    int32_t min_2pc_replica_count(int32_t app_max_replica_count)
```
acelyc111, line 121:
```suggestion
    int32_t min_2pc_replica_count(int32_t app_max_replica_count) const
```

rdsn/src/common/replication_common.h
```diff
@@ -117,6 +117,11 @@ class replication_options
...
+    {
+        return std::min(mutation_2pc_min_replica_count, app_max_replica_count) - 1;
```
acelyc111, line 123: Should be `return std::min(mutation_2pc_min_replica_count, app_max_replica_count - 1);` ?

acelyc111, line 123: Should be `return std::min(mutation_2pc_min_replica_count, app_max_replica_count - 1);` ?

acelyc111, line 123: Should be `return std::min(mutation_2pc_min_replica_count, app_max_replica_count - 1);` ?

foreverneverer: yeah, the latest may be not pushed...

foreverneverer: Oh, the function should be removed...

rdsn/src/replica/replica.cpp
```diff
@@ -581,5 +581,10 @@ error_code replica::store_app_info(app_info &info, const std::string &path)
     return err;
 }
 
+int32_t replica::mutation_2pc_min_replica_count()
+{
+    return std::min(_options->mutation_2pc_min_replica_count, _app_info.max_replica_count - 1);
```
empiredan, line 586: Should this be compared as follows ?
```c++
return std::min({_options->mutation_2pc_min_replica_count, _app_info.max_replica_count - 1, 1});
```

Since for single replica `max_replica_count` is 1, then `std::min(_options->mutation_2pc_min_replica_count, _app_info.max_replica_count - 1)` will be 0.

foreverneverer: I retain _options->mutation_2pc_min_replica_count means that in the case: mutation_2pc_min_replica_count = 2, max_replica_count = 5, but the mim_2pc should equal with 2 but not 4, if you add `1` and then the mutation_2pc_min_replica_count is invalid.

But your case is right, so the function should:
```
assert(app_info.max_replica_count > 0);
return app_info.max_replica_count <=2 ? 1 : std::min(_options->mutation_2pc_min_replica_count, _app_info.max_replica_count - 1);
```

Do you think it is ok？

empiredan: Yeah, this made sense. However, another case is that the replica count is increased from 1 to 3: since `_options->mutation_2pc_min_replica_count` is originally set to 1, `std::min(_options->mutation_2pc_min_replica_count, _app_info.max_replica_count - 1)` will still return 1. Does this means we have to modify `_options->mutation_2pc_min_replica_count` ?

foreverneverer: Yeah, retain `_options->mutation_2pc_min_replica_count` is for user custom requirements, if user mark it `1` that means user hope just 1 replica is ok

foreverneverer: I suggest set `mutation_2pc_min_replica_count = 2` by default

empiredan: OK, I understand. I think in most cases setting `mutation_2pc_min_replica_count` to 2 is ok.

Is there a scenario that user sets `mutation_2pc_min_replica_count` to 3 while `max_replica_count` is 3 ? In this case the real `mutation_2pc_min_replica_count` will return 2. Should we handle this scenario ?

foreverneverer: Oh，great! Your suggestion is important.

So I modify like:
```c++
assert(app_info.max_replica_count > 0);
if (mutation_2pc_min_replica_count > 0) {//  >0 means use the user config
    return mutation_2pc_min_replica_count;
} else {// otherwise use the table max_replica_count
    return app_info.max_replica_count <=2 ? 1 :  _app_info.max_replica_count - 1
}

Comment (empiredan) [reply rdsn/src/replica/replica.cpp:586]: I think both methods can work well with most cases. Maybe it's better to reserve `mutation_2pc_min_replica_count` for special cases, so we can adopt first method ? And while `mutation_2pc_min_replica_count` is not enabled, use N/2+1 instead (no special reason, since PacificA doesn't need quorum write) ?
```
assert(_app_info.max_replica_count > 0);
if (mutation_2pc_min_replica_count > 0) {//  >0 means use the user config
    return mutation_2pc_min_replica_count;
} else {// otherwise use the table max_replica_count
    return app_info.max_replica_count <=2 ? 1 :  _app_info.max_replica_count / 2 + 1;
}
```

Comment (foreverneverer) [reply rdsn/src/replica/replica.cpp:586]: OK, I agree!

File: rdsn/src/meta/meta_service.cpp [hunk 6]
```diff
@@ -1306,5 +1306,15 @@ void meta_service::on_set_max_replica_count(configuration_set_max_replica_count_
...
 }
 
+int32_t meta_service::mutation_2pc_min_replica_count(int32_t app_max_replica_count) const
+{
+    dassert_f(_app_info.max_replica_count > 0);
+if (_opts.mutation_2pc_min_replica_count > 0) {//  >0 means use the user config
+    return _opts.mutation_2pc_min_replica_count;
+} else {// otherwise based on the table max_replica_count
+    return app_max_replica_count <=2 ? 1 :  app_max_replica_count / 2 + 1;
+}
```
empiredan, line 1316:
```suggestion
    if (_opts.mutation_2pc_min_replica_count > 0) { //  > 0 means use the user config
        return _opts.mutation_2pc_min_replica_count;
    } else { // otherwise based on the table max_replica_count
        return app_max_replica_count <= 2 ? 1 :  app_max_replica_count / 2 + 1;
    }
```

rdsn/src/replica/replica.cpp
```diff
@@ -581,5 +581,15 @@ error_code replica::store_app_info(app_info &info, const std::string &path)
...
 }
 
+int32_t replica::mutation_2pc_min_replica_count() const
+{
+    dassert_f(_app_info.max_replica_count > 0);
+if (_options->mutation_2pc_min_replica_count > 0) {//  >0 means use the user config
+    return _options->mutation_2pc_min_replica_count;
+} else {// otherwise based on the table max_replica_count
+    return _app_info.max_replica_count <=2 ? 1 :  _app_info.max_replica_count / 2 + 1;
+}
```
empiredan, line 591:
```suggestion
    if (_options->mutation_2pc_min_replica_count > 0) { // > 0 means use the user config
        return _options->mutation_2pc_min_replica_count;
    } else { // otherwise based on the table max_replica_count
        return _app_info.max_replica_count <= 2 ? 1 : _app_info.max_replica_count / 2 + 1;
    }
```

foreverneverer: format done

rdsn/src/meta/meta_service.cpp
```diff
@@ -1306,5 +1306,15 @@ void meta_service::on_set_max_replica_count(configuration_set_max_replica_count_
                      server_state::sStateHash);
 }
 
+int32_t meta_service::mutation_2pc_min_replica_count(int32_t app_max_replica_count) const
+{
+    dassert_f(app_max_replica_count > 0, "max_replica_count > 0");
```
empiredan, line 1311: dcheck_gt(app_max_replica_count, 0);

rdsn/src/meta/meta_service.cpp
```diff
@@ -1306,5 +1306,15 @@ void meta_service::on_set_max_replica_count(configuration_set_max_replica_count_
...
+    if (_opts.mutation_2pc_min_replica_count > 0) { //  >0 means use the user config
+        return _opts.mutation_2pc_min_replica_count;
+    } else { // otherwise, the value based on the table max_replica_count
+        return app_max_replica_count <= 2 ? 1 : app_max_replica_count / 2 + 1;
+    }
+}
+
```
empiredan, line 1318: I think we can encapsulate this logic into a function `int32_t mutation_2pc_min_replica_count(int32_t 2pc_min_replica_count, int32_t app_max_replica_count)`, and put in `rdsn/src/common/replication_common.h` ?

Therefore both functions can just be implemented like this:
```C++
int32_t meta_service::mutation_2pc_min_replica_count(int32_t app_max_replica_count) const
{
    return mutation_2pc_min_replica_count(_opts.mutation_2pc_min_replica_count, app_max_replica_count);
}

int32_t replica::mutation_2pc_min_replica_count() const
{
    return mutation_2pc_min_replica_count(_options->mutation_2pc_min_replica_count, _app_info.max_replica_count);
}
```

empiredan:
Should we also add new `mutation_2pc_min_replica_count()` into `partition_health_status()` ? Since it also uses `mutation_2pc_min_replica_count` to decide if a partition is unwritable:

```c++
inline health_status partition_health_status(const partition_configuration &pc,
                                             int mutation_2pc_min_replica_count)
{
    if (pc.primary.is_invalid()) {
        if (pc.secondaries.empty())
            return HS_DEAD;
        else
            return HS_UNREADABLE;
    } else { // !pc.primary.is_invalid()
        int n = pc.secondaries.size() + 1;
        if (n < mutation_2pc_min_replica_count)
            return HS_UNWRITABLE;
        else if (n < pc.max_replica_count)
            return HS_WRITABLE_ILL;
        else
            return HS_HEALTHY;
    }
}
```

empiredan (dismissed):

rdsn/src/common/replication_common.cpp
```diff
@@ -230,7 +230,8 @@ void replication_options::initialize()
         "replication",
         "mutation_2pc_min_replica_count",
         mutation_2pc_min_replica_count,
-        "minimum number of alive replicas under which write is allowed");
+        "minimum number of alive replicas under which write is allowed. it's valid if larger than "
+        "0, otherwise, the final value based app max replica count");
```
acelyc111, line 234:
```suggestion
        "0, otherwise, the final value is based on app_max_replica_count");
```

acelyc111 (approved):

empiredan (approved):

Issue #1038, opened 2022-07-07
a specific version thrift tool used by java-client, cann't download now


shenxingwuying:
wget http://mirrors.tuna.tsinghua.edu.cn/apache/thrift/0.11.0/thrift-0.11.0.tar.gz

--2022-07-07 23:32:07--  http://mirrors.tuna.tsinghua.edu.cn/apache/thrift/0.11.0/thrift-0.11.0.tar.gz
Resolving mirrors.tuna.tsinghua.edu.cn (mirrors.tuna.tsinghua.edu.cn)... 101.6.15.130, 2402:f000:1:400::2
Connecting to mirrors.tuna.tsinghua.edu.cn (mirrors.tuna.tsinghua.edu.cn)|101.6.15.130|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2022-07-07 23:32:07 ERROR 404: Not Found.

PR #1045, opened 2022-07-11
chore(build): remove the auto-generated Java code by thrift compiler

shenxingwuying:

Many auto codes by thrift generated by thrift, that's not necessary.
Byt It depends a higher thrift version, it's download url has invalid, an issue :
https://github.com/apache/incubator-pegasus/issues/1038


java-client/idl/recompile_thrift.sh
```diff
@@ -19,21 +19,28 @@
...
-# Note: by default the thrift will be installed in /usr/local. If you'd like to install it somewhere else,
-# you'd better provide install path with option `--prefix=xxx'.
- 
+function GenThriftTool() {
+  wget http://mirrors.tuna.tsinghua.edu.cn/apache/thrift/0.11.0/thrift-0.11.0.tar.gz
+  tar xvf thrift-0.11.0.tar.gz
+  cd thrift-0.11.0
+  ./configure --with-csharp=no --with-java=no --with-python=no --with-erlang=no --with-perl=no \
+      --with-php=no --with-ruby=no --with-haskell=no --with-php_extension=no --with-rs=no --with-go=no
+  make
```
acelyc111, line 28: Better to add ` -j $(($(nproc)/2+1))` to make faster.

java-client/idl/recompile_thrift.sh
```diff
@@ -19,21 +19,28 @@
...
+  sudo make install
```
acelyc111, line 29: It's suggest to install packages in Dockerfile.

java-client/idl/recompile_thrift.sh
acelyc111: leave some CPU resource for other purpose

java-client/idl/recompile_thrift.sh
```diff
@@ -19,21 +19,33 @@
...
-#       --with-php=no --with-ruby=no --with-haskell=no --with-php_extension=no --with-rs=no --with-go=no
-#   make
-#   sudo make install
-# 
-# Note: by default the thrift will be installed in /usr/local. If you'd like to install it somewhere else,
-# you'd better provide install path with option `--prefix=xxx'.
- 
+function GenThriftTool() {
+    set -e
+    wget --progress=dot:giga https://github.com/apache/thrift/archive/refs/tags/0.11.0.tar.gz -O thrift-0.11.0.tar.gz
```
acelyc111, line 24: Why we use thrift-0.11.0 in Java and 0.9.3 in all the other places, there must be some historical reason.
I can remember a little there is a compatiable reason in Java client that we shaded thrift 0.11.0, but not sure.
Any body could please clarify it? Many thanks!

java-client/idl/recompile_thrift.sh
```diff
@@ -19,21 +19,33 @@
...
+    tar xzf thrift-0.11.0.tar.gz
+    pushd thrift-0.11.0
+    ./bootstrap.sh
+    ./configure --with-csharp=no --with-java=no --with-python=no --with-erlang=no --with-perl=no \
+        --with-php=no --with-ruby=no --with-haskell=no --with-php_extension=no --with-rs=no --with-go=no
```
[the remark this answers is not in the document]
shenxingwuying: > We only need the execute binary? You can set `--enable_libs=no` instead.

The command copy from history' comments, as left side.
I try it.

java-client/build.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
shenxingwuying: > Add license header

ok

java-client/idl/recompile_thrift.sh
shenxingwuying: > Why we use thrift-0.11.0 in Java and 0.9.3 in all the other places, there must be some historical reason. I can remember a little there is a compatiable reason in Java client that we shaded thrift 0.11.0, but not sure. Any body could please clarify it? Many thanks!

I don't know the history's problems., and the problems not just thrift version. One more thing,
in fact, I think the thrift proto files (for interface) should be one shared by cpp and java and others. But the proto files has very large diffenences between cpp and java.

java-client/build.sh
```diff
@@ -0,0 +1,50 @@
...
```
acelyc111, line 1: There are too many shell scripts in the project, and we are going to clean them up, it's not suggest to add more shell scripts now. You can describe the build/install/test commands in README instead.

shenxingwuying: ok

acelyc111 (dismissed):
Great!

acelyc111 (approved):

docker/pegasus-build-env/centos6/Dockerfile
```diff
@@ -20,16 +20,16 @@ FROM centos:6.10
 LABEL maintainer=jiashuo
 
 # defaut repo url hasn't been supported, use thirdparty repo url
-RUN curl https://www.getpagespeed.com/files/centos6-eol.repo --output /etc/yum.repos.d/CentOS-Base.repo
-RUN curl https://www.getpagespeed.com/files/centos6-epel-eol.repo --output /etc/yum.repos.d/epel.repo
-RUN curl https://www.getpagespeed.com/files/centos6-scl-eol.repo --output /etc/yum.repos.d/CentOS-SCLo-scl.repo
-RUN curl https://www.getpagespeed.com/files/centos6-scl-rh-eol.repo --output /etc/yum.repos.d/CentOS-SCLo-scl-rh.repo
+RUN wget --progress=dot:giga https://www.getpagespeed.com/files/centos6-eol.repo -O /etc/yum.repos.d/CentOS-Base.repo
```
foreverneverer, line 23: it seems not have `wget` tools for the new centos env

java-client/src/main/java/org/apache/pegasus/apps/batch_get_request.java
```diff
@@ -1,469 +0,0 @@
-/*
```
foreverneverer, line 1: These auto-generated file should add into `.gitignore`, other same

acelyc111: +1

docker/pegasus-build-env/centos6/Dockerfile
shenxingwuying: 1. Docker lint will warn both curl and wget used together. So avoid this should use only one.
2. curl get the file thrift-0.11.0.tgz, I failed. So I fix curl before to wget.
3. If wget is not exist. Keey curl and ignore the warnings in docker lint?

shenxingwuying: -:81 DL4001 warning: Either use Wget or Curl but not both

java-client/src/main/java/org/apache/pegasus/apps/batch_get_request.java
shenxingwuying: done

docker/pegasus-build-env/centos6/Dockerfile
shenxingwuying: DONE

foreverneverer: > 2\. curl get the file thrift-0.11.0.tgz, I failed. So I fix curl before to wget.

thrift can be fetched by `wget`, but the centos6-epel-eol.repo...., should use `curl`

Or you can test, whether `wget` exist or not for the new centos6 env. I remember it don't have

Issue #1047, opened 2022-07-12
use official rocksdb library


acelyc111:
For some historical reasons, we use the Xiaomi modified rocksdb library, since Pegasus 2.1.0, we have removed all modifications on rocksdb, only left some code to keep compatiable [1], now before 2.4.0 releasing, it's a chance to switch to the official version.

1. https://github.com/XiaoMi/pegasus-rocksdb/commits/v6.6.4-compatible

PR #1048, opened 2022-07-12
feat(thirdparty): use official rocksdb

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1047

For some historical reasons, we use the Xiaomi modified rocksdb library. Since Pegasus 2.1.0, we have removed all modifications on rocksdb, only left some code to keep compatiable [1], now we are planning to switch to the official version.

It's expect that `thirdparty/rocksdb_fix_atomic_flush_0879c240.patch` must be corresponding to this patch [2].

**ATTENTION**: There is a limit that the server must be upgraded from 2.1, 2.2, 2.3 or 2.4, to ensure that there is no Pegasus introduced tags which the official rocksdb doesn't recognize left in MANIFEST file.

1. https://github.com/XiaoMi/pegasus-rocksdb/commits/v6.6.4-compatible
2. https://github.com/XiaoMi/pegasus-rocksdb/commit/ef29819c7a1ea9334ae170f506b653757f517a52

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -87,6 +87,7 @@ header:
     - 'thirdparty/fix_libevent_for_macos.patch'
     - 'thirdparty/fix_s2_for_aarch64.patch'
     - 'thirdparty/fix_thrift_for_cpp11.patch'
+    - 'thirdparty/rocksdb_fix_atomic_flush_0879c240.patch'
     # should be empty, or ignore all comment lines
     - 'src/utils/test/config-empty.ini'
     # The MIT License (MIT), Copyright (c) 2015 Microsoft Corporation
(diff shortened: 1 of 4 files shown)


rdsn/thirdparty/CMakeLists.txt
```diff
@@ -378,9 +378,9 @@ ExternalProject_Add(jemalloc
 option(ROCKSDB_PORTABLE "build a portable binary" OFF)
 
 ExternalProject_Add(rocksdb
-        URL ${OSS_URL_PREFIX}/pegasus-rocksdb-ef29819c7a1ea9334ae170f506b653757f517a52.zip
-        https://github.com/XiaoMi/pegasus-rocksdb/archive/ef29819c7a1ea9334ae170f506b653757f517a52.zip
-        URL_MD5 2a6488cd87c37e0c2e42a5ca74bebc87
+        https://github.com/facebook/rocksdb/archive/refs/tags/v6.6.4.tar.gz
```
acelyc111, line 381:
```suggestion
        URL https://github.com/facebook/rocksdb/archive/refs/tags/v6.6.4.tar.gz
```

levy5307 (dismissed):

foreverneverer (changes requested):
hope block merge before release 2.4

levy5307:
> hope block merge before release 2.4

Better to convert it to draft :)

acelyc111:
> hope block merge before release 2.4

@foreverneverer Since we have created an separete branch v2.4 for releasing 2.4.0, so could we continue this work?

foreverneverer (approved):

empiredan (approved):

Issue #819, opened 2021-09-27
support jemalloc


levy5307:
## Feature Request

make pegasus to support jemalloc

### Related pull requests

https://github.com/XiaoMi/rdsn/pull/910

empiredan:
## Support to set/get dirty_decay_ms and muzzy_decay_ms dynamically

Since we've supported building rDSN with jemalloc in https://github.com/XiaoMi/rdsn/pull/910, it's necessary for us to  supporting setting/getting dirty_decay_ms and muzzy_decay_ms dynamically. 

For details please see https://github.com/XiaoMi/rdsn/pull/928

PR #1050, opened 2022-07-15
feat: support to use jemalloc while building pegasus

empiredan:
This PR is for https://github.com/apache/incubator-pegasus/issues/819.

Previously we've supported jemalloc while building rDSN (https://github.com/XiaoMi/rdsn/pull/910). However, building pegasus with jemalloc enabled will lead to the following problem: 
image

The reason can be described like this: cmake file for RocksDB had been generated to enable jemalloc (JeMalloc::JeMalloc);
on the other hand, however, rDSN was not told that jemalloc should be enabled, since pegasus didn't pass `USE_JEMALLOC` to rDSN.

Enabling jemalloc while building pegasus can be implemented in this way: since `dsn_common_setup()` are called in `src/CMakeLists.txt` before building executables, using jemalloc for pegasus can be supported by passing the option `USE_JEMALLOC` to `rdsn/bin/dsn.cmake`.


foreverneverer:
Great！

foreverneverer (approved):

src/CMakeLists.txt
```diff
@@ -25,6 +25,11 @@ endif()
 
 dsn_common_setup()
 
+message(STATUS "USE_JEMALLOC = ${USE_JEMALLOC}")
+if(USE_JEMALLOC)
+    message(STATUS "jemalloc is enabled")
+endif()
+
```
acelyc111, line 32: Move to ./CMakeLists.txt

acelyc111 (approved):

foreverneverer (approved):

Issue #1064, opened 2022-07-19
Fix：fix verification bug for recovery function


hycdong:
When zk is not useful, pegasus provides the function to recovery data from replica server to zk, meta server will collect app information from replica servers and verify their consistence. However, in release 2.1.1, 2.2.x, 2.3.x, the `duplicating` filed of `app_info` are different between primary and secondaries, which will lead to verification failed.

PR #1065, opened 2022-07-19
fix: fix recovery verification bug

hycdong:
As https://github.com/apache/incubator-pegasus/issues/1064 shows, this pr:
- fix recorery verification bug
- refactor meta recovery code by using rpc_holder structure
- update recovery function test to avoid unnecessary failure

### DIFF
--- a/src/rdsn/src/meta/server_state.h
+++ b/src/rdsn/src/meta/server_state.h
@@ -353,6 +353,41 @@ private:
                                                   const std::vector<std::string> &keys,
                                                   const std::vector<std::string> &values);
 
+    bool app_info_compatible_equal(const app_info &l, const app_info &r) const
+    {
+        if (l.status != r.status) {
+            return false;
+        }
+        if (l.app_type != r.app_type) {
+            return false;
+        }
+        if (l.app_name != r.app_name) {
+            return false;
+        }
+        if (l.app_id != r.app_id) {
+            return false;
+        }
+        if (l.partition_count != r.partition_count) {
+            return false;
+        }
+        if (l.is_stateful != r.is_stateful) {
+            return false;
+        }
+        if (l.max_replica_count != r.max_replica_count) {
+            return false;
+        }
+        if (l.expire_second != r.expire_second) {
+            return false;
+        }
+        if (l.create_second != r.create_second) {
+            return false;
+        }
+        if (l.drop_second != r.drop_second) {
+            return false;
+        }
+        return true;
+    }
+
 private:
     friend class bulk_load_service;
     friend class bulk_load_service_test;
--- a/src/rdsn/src/meta/server_state.h
+++ b/src/rdsn/src/meta/server_state.h
@@ -355,34 +355,11 @@ private:
 
     bool app_info_compatible_equal(const app_info &l, const app_info &r) const
     {
-        if (l.status != r.status) {
-            return false;
-        }
-        if (l.app_type != r.app_type) {
-            return false;
-        }
-        if (l.app_name != r.app_name) {
-            return false;
-        }
-        if (l.app_id != r.app_id) {
-            return false;
-        }
-        if (l.partition_count != r.partition_count) {
-            return false;
-        }
-        if (l.is_stateful != r.is_stateful) {
-            return false;
-        }
-        if (l.max_replica_count != r.max_replica_count) {
-            return false;
-        }
-        if (l.expire_second != r.expire_second) {
-            return false;
-        }
-        if (l.create_second != r.create_second) {
-            return false;
-        }
-        if (l.drop_second != r.drop_second) {
+        if (l.status != r.status || l.app_type != r.app_type || l.app_name != r.app_name ||
+            l.app_id != r.app_id || l.partition_count != r.partition_count ||
+            l.is_stateful != r.is_stateful || l.max_replica_count != r.max_replica_count ||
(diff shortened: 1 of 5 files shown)


src/rdsn/src/meta/server_state.h
```diff
@@ -353,6 +353,41 @@ class server_state
...
+        }
+        if (l.expire_second != r.expire_second) {
+            return false;
+        }
+        if (l.create_second != r.create_second) {
+            return false;
+        }
+        if (l.drop_second != r.drop_second) {
+            return false;
+        }
```
foreverneverer, line 387:
```suggestion
        if (l.status != r.status || l.app_type != r.app_type || ... || ... ...) {
            return false;
        }
```

src/rdsn/src/meta/server_state.h
```diff
@@ -353,6 +353,18 @@ class server_state
                                                   const std::vector<std::string> &keys,
                                                   const std::vector<std::string> &values);
 
+    bool app_info_compatible_equal(const app_info &l, const app_info &r) const
+    {
+        if (l.status != r.status || l.app_type != r.app_type || l.app_name != r.app_name ||
```
acelyc111, line 358: Better to add some comments to explain why do this partial fields check.

acelyc111:
I found before `duplicating` field is, before:
```
12:optional bool duplicating;
```

after:
```
12:optional bool duplicating = false;
```

Because it's `optional`, the `operator ==` of `app_info` will check whether this filed is set or not at first, like:
```
    if (__isset.duplicating != rhs.__isset.duplicating)
      return false;
    else if (__isset.duplicating && !(duplicating == rhs.duplicating))
      return false;
```

That is to say, if `duplicating` is not set, the default value will not take effect, so how this bug happend?

hycdong:
> I found before `duplicating` field is, before:
> 
> ```
> 12:optional bool duplicating;
> ```
> 
> after:
> 
> ```
> 12:optional bool duplicating = false;
> ```
> 
> Because it's `optional`, the `operator ==` of `app_info` will check whether this filed is set or not at first, like:
> 
> ```
>     if (__isset.duplicating != rhs.__isset.duplicating)
>       return false;
>     else if (__isset.duplicating && !(duplicating == rhs.duplicating))
>       return false;
> ```
> 
> That is to say, if `duplicating` is not set, the default value will not take effect, so how does this bug happend?


For Pegasus release 2.1.x, 2.2.x and 2.3.x, the duplicating field of `app_info` is like: 
> 12:optional bool duplicating;

It means that when a table is not executing duplication, `app_info.__isset.duplicating = false`. 
However, in the releases version above, `duplicating` field will be set by function `on_config_proposal`:
https://github.com/apache/incubator-pegasus/blob/42f4eee5b41192145aec1a9b41d9d0f32c7115ce/src/rdsn/src/replica/replica_config.cpp#L95

It will lead to primary and secondaries have inconsistent duplicating value when table is not executing duplication. 
For primary replica -> app_info.__isset.duplicating = false,
For secondary replica -> app_info.__isset.duplicating = true, app_info.duplicating = false.

As result, tables created in 2.1.x, 2.2.x, 2.3.x and not duplicating, they will have inconsistent `duplicating` field between replicas, which will lead to the bug.

src/rdsn/src/meta/server_state.h
hycdong: > // compatible for app.duplicating different between primary and secondaries in 2.1.x, 2.2.x and 2.3.x release

The comment is added for its caller, it will only be used for this special compatible case.

acelyc111:
> > I found before `duplicating` field is, before:
> > ```
> > 12:optional bool duplicating;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > after:
> > ```
> > 12:optional bool duplicating = false;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Because it's `optional`, the `operator ==` of `app_info` will check whether this filed is set or not at first, like:
> > ```
> >     if (__isset.duplicating != rhs.__isset.duplicating)
> >       return false;
> >     else if (__isset.duplicating && !(duplicating == rhs.duplicating))
> >       return false;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That is to say, if `duplicating` is not set, the default value will not take effect, so how does this bug happend?
> 
> For Pegasus release 2.1.x, 2.2.x and 2.3.x, the duplicating field of `app_info` is like:
> 
> > 12:optional bool duplicating;
> 
> It means that when a table is not executing duplication, `app_info.__isset.duplicating = false`. However, in the releases version above, `duplicating` field will be set by function `on_config_proposal`:
> 
> https://github.com/apache/incubator-pegasus/blob/42f4eee5b41192145aec1a9b41d9d0f32c7115ce/src/rdsn/src/replica/replica_config.cpp#L95
> 
> It will lead to primary and secondaries have inconsistent duplicating value when table is not executing duplication. For primary replica -> app_info.__isset.duplicating = false, For secondary replica -> app_info.__isset.duplicating = true, app_info.duplicating = false.
> 
> As result, tables created in 2.1.x, 2.2.x, 2.3.x and not duplicating, they will have inconsistent `duplicating` field between replicas, which will lead to the bug.

I see, thanks for your clarify.

acelyc111 (approved):

Issue #1066, opened 2022-07-19
Still-running timer will lead to heap-use-after-free error after percentile is destructed


empiredan:
The timer will call `cancel()` while percentile is being destructed, however it won't wait until the callback function is finished. This will lead to heap-use-after-free error as follows (https://github.com/apache/incubator-pegasus/runs/7388468584?check_suite_focus=true):
```
[----------] 11 tests from metrics_test
[ RUN      ] metrics_test.create_entity
[       OK ] metrics_test.create_entity (2 ms)
[ RUN      ] metrics_test.recreate_entity
[       OK ] metrics_test.recreate_entity (0 ms)
[ RUN      ] metrics_test.create_metric
[       OK ] metrics_test.create_metric (0 ms)
[ RUN      ] metrics_test.recreate_metric
[       OK ] metrics_test.recreate_metric (0 ms)
[ RUN      ] metrics_test.gauge_int64
[       OK ] metrics_test.gauge_int64 (0 ms)
[ RUN      ] metrics_test.gauge_double
[       OK ] metrics_test.gauge_double (0 ms)
[ RUN      ] metrics_test.gauge_increment
[       OK ] metrics_test.gauge_increment (2698 ms)
[ RUN      ] metrics_test.counter
[       OK ] metrics_test.counter (3526 ms)
[ RUN      ] metrics_test.volatile_counter
[       OK ] metrics_test.volatile_counter (11942 ms)
[ RUN      ] metrics_test.percentile_int64
[       OK ] metrics_test.percentile_int64 (5458 ms)
[ RUN      ] metrics_test.percentile_double
[       OK ] metrics_test.percentile_double (5451 ms)
[----------] 11 tests from metrics_test (29077 ms total)
[----------] 4 tests from nth_element_test
[ RUN      ] nth_element_test.basic_int64
[       OK ] nth_element_test.basic_int64 (0 ms)
[ RUN      ] nth_element_test.generated_int64
[       OK ] nth_element_test.generated_int64 (5 ms)
[ RUN      ] nth_element_test.basic_double
[       OK ] nth_element_test.basic_double (0 ms)
[ RUN      ] nth_element_test.generated_double
[       OK ] nth_element_test.generated_double (8 ms)
[----------] 4 tests from nth_element_test (13 ms total)
[----------] 4 tests from table_printer_test
[ RUN      ] table_printer_test.empty_content_test
[       OK ] table_printer_test.empty_content_test (0 ms)
[...]

PR #1074, opened 2022-07-24
feat: support to close percentile to prevent from heap-use-after-free error

empiredan:
This PR is to resolve the issue https://github.com/apache/incubator-pegasus/issues/1066.

The root cause of issue https://github.com/apache/incubator-pegasus/issues/1066 is that the percentile is freed during the destruction of the metric registry, after which sometimes the handler for asynchronous wait operation of timer will be still executed, where the freed percentile is certainly referenced. This will lead to heap-use-after-free error.

The solution for this problem will be extend the lifetime of percentile: during the construction of the percentile, it can increment its ref count (its base class is `ref_counter`) since it is referenced by timer. This effectively extends the lifetime of percentile: even if both of its metric entity and registry are destructed, it can still survive in the memory for the reason that its ref count is 1 at least.

Once the percentile should be released when it will retire from metric entity or the metric registry (singleton) will be destructed, the `close()` method for percentile will be invoked. `close()` will change the state of timer from *running* to *closing*, and also call `cancel()` for the timer. Then, whether the timer is cancelled successfully or not, the timer will finally be stopped: if successful, the callback for cancel will be executed immediately; otherwise, the *closing* state will tell the timer to stop finally, albeit sometimes after a while.


src/rdsn/include/dsn/utility/metrics.h
```diff
@@ -168,6 +171,16 @@ class metric_entity : public ref_counter
...
 
+    // Close all "closeable" metrics owned by this entity.
+    //
+    // `async` is used to control if the close operations are asynchronous or not. It is set to
+    // true by default, which means all metrics owned by this entity will be closed asynchronously
+    // without waiting for the close operations to be finished.
+    //
+    // Otherwise, once `async` is set to false, close() will be blocked until the close operations
+    // are finished.
+    void close(bool async = true);
```
acelyc111, line 182: use 'wait' maybe more accurate, and it's suggest to use enum class instead of boolean, true/false is meanless when we see the invoker.

src/rdsn/src/utils/metrics.cpp
```diff
@@ -132,27 +185,78 @@ uint64_t percentile_timer::generate_initial_delay_ms(uint64_t interval_ms)
...
+        on_close();
+
         return;
     }
 
-    _exec();
+    TRY_PROCESS_TIMER_CLOSING();
+    _on_exec();
 
+    TRY_PROCESS_TIMER_CLOSING();
```
acelyc111, line 259: better to add `#undef TRY_PROCESS_TIMER_CLOSING` when you'll never use it.

src/rdsn/include/dsn/utility/metrics.h
empiredan: OK, we can use enum class instead.

src/rdsn/src/utils/metrics.cpp
empiredan: Sure.

acelyc111 (approved):

foreverneverer (approved):

Issue #1053, opened 2022-07-17
Merge rdsn and Pegasus


acelyc111:
Although rdsn and Pegasus have been merged to one repo, there are still many work left to do:

- [x] unify th clang-format tools
- [x] remove run.sh in rdsn
- [x] unify the workflows
- [x] unify the CMakelists.txt
- [x] Unify build and test shell scripts to run.sh
- [x] Move some common files out of rdsn directory
- [x] reduce rebuilding time cost
- [x] Unify the cmake of rdsn and pegasus
- [x] move files in 'include' dir of rdsn to the src pf rdsn
- [x] refactor(rdsn): flatten source code directory by moving files from 'src/rdsn/src/*' to 'src/*'

acelyc111:
# Refactor directories

After the [xiaomi/rdsn](https://github.com/XiaoMi/rdsn) project has been merged into the Pegasus project, the rdsn source code directory structure is kept as before in the original repository. But it's just a temporary status, we are planning to integrate the two projects closer in the long-term plan.

The following things I'm going to do is moving source code directories from `src/rdsn/src/*` to `src/*`, for example, `src/rdsn/src/runtime/api_task.h` will be moved to `src/runtime/api_task.h`, this change will make the source structure more simple and easier to understand and accept by more new comer of Pegasus.

Of course, it's just one step of the whole refactory plan, the `src/rdsn/include/*` files have been moved to the places where they are more related to by pull request [1]. And there will be more refactor pull requests to modularize the source files in the future.

However, this change will influence hundreds of files, their paths will be changed, even though there are no functional changes. This change is presented by this [2] pull request.

Directories in `src` before this refactor:
```
.
├── base
│   └── test
├── client_lib
├── geo
│   ├── bench
│   ├── lib
│   └── test
├── include
│   ├── pegasus
│   └── rrdb
├── rdsn
│   └── src
│       ├── aio
│       │   └── test
│       ├── block_service
│       │   ├── fds
│       │   ├── hdfs
│       │   ├── local
│       │   └── test
│       ├── client
│       ├── common
│       │   ├── serialization_helper
│       │   └── test
│       ├── failure_detector
│       │   └── test
│       ├── http
│       │   └── test
│       ├── meta
│       │   ├── duplication
│       │   └── test
│       │       ├── balancer_simulator
│       │       ├── meta_state
│       │       └── misc
│       ├── nfs
│       │   └── test
│       ├── perf_counter
│       │   └── test
[...]

Issue #1060, opened 2022-07-18
Core dumped for unit test meta_bulk_load_http_test.start_compaction_test


empiredan:
While executing `meta_bulk_load_http_test.start_compaction_test` core dumped as follows:
```
[----------] 2 tests from meta_http_service_test
[ RUN      ] meta_http_service_test.get_app_from_primary
[       OK ] meta_http_service_test.get_app_from_primary (671 ms)
[ RUN      ] meta_http_service_test.get_app_envs
[       OK ] meta_http_service_test.get_app_envs (668 ms)
[----------] 2 tests from meta_http_service_test (1339 ms total)
[----------] 1 test from meta_backup_test_base
[ RUN      ] meta_backup_test_base.get_backup_policy
[       OK ] meta_backup_test_base.get_backup_policy (1669 ms)
[----------] 1 test from meta_backup_test_base (1669 ms total)
[----------] 4 tests from meta_bulk_load_http_test
[ RUN      ] meta_bulk_load_http_test.start_bulk_load_request
[       OK ] meta_bulk_load_http_test.start_bulk_load_request (1174 ms)
[ RUN      ] meta_bulk_load_http_test.query_bulk_load_request
E2022-07-18 09:50:29.361 (1658137829361131403 9503) test_meta.THREAD_POOL_DEFAULT0.0000250600010001: meta_bulk_load_service.cpp:1537:on_query_bulk_load_status(): app(app_not_exist) is not existed or not available
[       OK ] meta_bulk_load_http_test.query_bulk_load_request (2172 ms)
[ RUN      ] meta_bulk_load_http_test.start_compaction_test
got signal id: 11
Segmentation fault (core dumped)
ERROR: run dsn.meta.test failed, return_code = 139
Error: Process completed with exit code 1.
```

For details please see https://github.com/apache/incubator-pegasus/runs/7386305679?check_suite_focus=true.

acelyc111:
Maybe also related to `meta_bulk_load_http_test.update_scenario_test` test Segmentation fault.
```
[----------] 4 tests from meta_bulk_load_http_test
[ RUN      ] meta_bulk_load_http_test.start_bulk_load_request
[       OK ] meta_bulk_load_http_test.start_bulk_load_request (1195 ms)
[ RUN      ] meta_bulk_load_http_test.query_bulk_load_request
[...]

PR #1079, opened 2022-07-25
refactor: Move some common files out of rdsn directory

acelyc111:
- move rdsn docs to ./docs
- move common *.thrift to ./idl
- move thrift compilation scripts to ./scripts


foreverneverer (approved):

src/rdsn/src/runtime/security/security.thrift
```diff
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-include "../../dsn.thrift"
+include "../../../../../idl/dsn.thrift"
```
neverchanje, line 18: dsn.thrift in rdsn/src/ looks fine to me, why did you move it to an outer directory?

neverchanje (approved):
rest lgtm

neverchanje:
```
[ RUN      ] meta_bulk_load_http_test.start_compaction_test
got signal id: 11
Segmentation fault (core dumped)
```

Seems bulkload has some bugs.

src/rdsn/src/runtime/security/security.thrift
acelyc111: dsn.thrift contains some common thrift structures which will be used for not only rdsn, but also client libs. Its name 'dsn.thrift' is not accurate, maybe 'base.thrift' would be better, like it named in client lib directories(e.g. https://github.com/apache/incubator-pegasus/blob/master/java-client/idl/base.thrift).

acelyc111:
> start_compaction_test

not related to this pr, we tracked it by issue, https://github.com/apache/incubator-pegasus/issues/1060

src/rdsn/src/runtime/security/security.thrift
acelyc111: Of course, we will refactor the code and only leave one unique base.thrift in Pegasus later.

PR #1082, opened 2022-07-27
refactor(build): skip re-generating thrift and skip re-cmaking

acelyc111:

this patch includes:
1. skip re-generating thrift and re-cmaking if it's not the first build, or doesn't specify `-c`. It will save much time if you only modify a little code.
2. add `set -e` for most of the shell scripts
3. use python instead of shell to get local ip
4. fix an error path to `$ROOT/src/builder/src/server/test/config.ini` in run.sh


neverchanje:
scripts/collector_table_counter_gen.py Remove this file if it is useless.

neverchanje (approved):

empiredan (approved):

acelyc111:
> scripts/collector_table_counter_gen.py Remove this file if it is useless.

OK, we'll do it later.

PR #1087, opened 2022-07-29
refactor(java-client): optimize code style

totalo:
[#887](https://github.com/apache/incubator-pegasus/issues/887)


java-client/src/main/java/org/apache/pegasus/client/ClientOptions.java
```diff
@@ -227,6 +227,11 @@ public boolean equals(Object options) {
     return false;
   }
 
+  @Override
+  public int hashCode() {
```
Apache9, line 231: Usually this is not a good practise to just call super.hashCode...

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
```diff
@@ -188,6 +193,16 @@ private void asyncStartScan() {
         new Table.ClientOPCallback() {
           @Override
           public void onCompletion(client_operator clientOP) throws Throwable {
+            if (!(clientOP instanceof rrdb_get_scanner_operator)) {
+              logger.error(
+                  "scan rpc callback, encounter logic error, we just abandon this scan, "
+                      + "tableName({}), appId({})",
+                  _table.getTableName(),
```
Apache9, line 200: These are all C style naming? Can we remove the leading '_'?

java-client/src/main/java/org/apache/pegasus/client/PegasusTable.java
```diff
@@ -98,6 +128,10 @@ public Future<Long> asyncSortKeyCount(byte[] hashKey, int timeout) {
         new Table.ClientOPCallback() {
           @Override
           public void onCompletion(client_operator clientOP) {
+            if (!(clientOP instanceof rrdb_sortkey_count_operator)) {
+              promise.setFailure(new PException("Unexpected client operator"));
```
Apache9, line 132: Better to not expose netty's promise and future. We can use CompletableFuture in JDK.  Can be a separated issue, and this will be an incompatible change...

Apache9 (commented):
Left some minor nits. PTAL.

java-client/src/main/java/org/apache/pegasus/client/ClientOptions.java
totalo: Got it, i will fix it.

java-client/src/main/java/org/apache/pegasus/client/PegasusScanner.java
totalo: Yes, but in this project, many of the formats are C-grammatical formats, and the current way of writing this is entirely because the attribute names are like this. I want to handle them uniformly in subsequent PRs.

java-client/src/main/java/org/apache/pegasus/client/PegasusTable.java
totalo: Consistent with the above c-format problem, the current usage is like this, I think the follow-up pr can be optimized.

acelyc111 (approved):
Thanks for your contribution to improve the Java code!

totalo:
> Thanks for your contribution to improve the Java code!

Happy to do this!

empiredan (approved):

Issue #1032, opened 2022-07-05
Release 2.4.0


foreverneverer:
# New Module
From this version, more project module will join Apache Pegasus Project. In this version, the following projects are included:
- [RDSN](https://github.com/XiaoMi/rdsn): In the past, `rdsn` exists as a sub-project in current repository, in this release, it officially become the core module of the project.
- PegasusClient: Pegasus support multi-language client. In the past, it existed different repository, in this release, the following clients are included in the repository:
  - [Pegasus-Java-Client](https://github.com/XiaoMi/pegasus-java-client)
  - [Pegasus-Scala-Client](https://github.com/xiaomi/pegasus-scala-client)
  - [Pegasus-Golang-Client](https://github.com/XiaoMi/pegasus-go-client)
  - [Pegasus-Python-Client](https://github.com/XiaoMi/pegasus-python-client)
  - [Pegasus-NodeJS-Client](https://github.com/XiaoMi/pegasus-nodejs-client)
- [PegasusDocker](https://github.com/xiaomi/pegasus-scala-client): Pegasus support docker-build, now it is included the latest version
- PegasusShell: In the past, Pegasus use c++ shell to manage the cluster, now in the latest version, we bring new shell tools: [AdminCli](https://github.com/pegasus-kv/admin-cli) and [Pegic](https://github.com/pegasus-kv/pegic)

# New architecture
In this version, we remove the shared log to enhance the pegasus performance,  Related pull request as follow:
- https://github.com/XiaoMi/rdsn/pull/993
- https://github.com/XiaoMi/rdsn/pull/994
- https://github.com/XiaoMi/rdsn/pull/999
- https://github.com/XiaoMi/rdsn/pull/1019
- https://github.com/XiaoMi/rdsn/pull/1022
- https://github.com/XiaoMi/rdsn/pull/1048
- https://github.com/XiaoMi/rdsn/pull/1028
- https://github.com/apache/incubator-pegasus/pull/890

# New Feature
## Replica-factor update
Supporting flexible replica count. In the past, the replica factor was Immutable once one table was created. In current version, user can dynamically adjust the factor of specified table. Related pull request as follow:

- https://github.com/XiaoMi/rdsn/pull/1061
- https://github.com/XiaoMi/rdsn/pull/1072
- https://github.com/XiaoMi/rdsn/pull/1077
- https://github.com/XiaoMi/rdsn/pull/1087
- https://github.com/XiaoMi/rdsn/pull/1109
[...]

PR #1088, opened 2022-08-01
fix: revert zk c client to 3.4.10 for compatibility

foreverneverer:
Related issue: https://github.com/apache/incubator-pegasus/issues/1092

This pr revert zk c client to 3.4.10, this need merge into Pegasus 2.4:  https://github.com/apache/incubator-pegasus/issues/1032

##### Tests 

- Unit test
- Integration test

##### Related changes

### DIFF
--- a/scripts/clear_zk.sh
+++ b/scripts/clear_zk.sh
@@ -34,7 +34,7 @@ then
 fi
 
 cd $INSTALL_DIR
-ZOOKEEPER_HOME=`pwd`/apache-zookeeper-3.7.0-bin
+ZOOKEEPER_HOME=`pwd`/apache-zookeeper-3.4.10
 
 if [ -d "$ZOOKEEPER_HOME" ]
 then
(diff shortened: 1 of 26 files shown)


acelyc111 (changes requested):
Let’s disccus clearly before merging, thanks!

foreverneverer:
> Let’s disccus clearly before merging, thanks!

OK, I hope it can be determined as soon as possible, because the release of 2.4 has been postponed for a long time. If it can't be solved, I hope this problem can be reconsidered after 2.4

acelyc111:
> > Let’s disccus clearly before merging, thanks!
> 
> OK, I hope it can be determined as soon as possible, because the release of 2.4 has been postponed for a long time. If it can't be solved, I hope this problem can be reconsidered after 2.4

How about release 2.4 without this patch, and you can release an internal release or a other type of spcial release with this patch. In SensorsData we have tested with zk client 3.7.0 with zk server kerberos enabled and disabled, all works well as expected. And you guys also tested that in some product envs, I guess there may be some DNS issue in you mentioned "tjwq-test" env :)

acelyc111 (approved):

acelyc111:
LGTM to merge it to branch 2.4.0

hycdong (approved):

Issue #1090, opened 2022-08-01
improve performance of count_data


GehaFearless:
## Feature Request

**Is your feature request related to a problem? Please describe:**
When we precisely count data for a large table, it's unnecessarily return key-values from server to client. 

**Describe the feature you'd like:**

I try to add optional section "only_return_count" on get_scanner_request & scan_request, also add optional  section kv_count on scan_response.
Different versions of the software will not interact with each other because of optional  section.

PR #728, opened 2021-04-28
improve performance of count_data

ZhongChaoqiang:
### What problem does this PR solve? 
When we precisely count data for a large table, it will cost minutes or hours.

### What is changed and how does it work?
Actually,we just need the count of data.So we just need transfer the count of data from server to client,  but not the detailed data.
In our test, it will 10x faster than before.

##### Tests 
- Unit test
- Manual test (add detailed scripts or steps below)
1. create a table with millions of data.
2. use "count_data -c -f"

##### Related changes
- Need to update the documentation
- Need to be included in the release note

PR #1091, opened 2022-08-01
feat: improve performance of count_data

GehaFearless:

issue: https://github.com/apache/incubator-pegasus/issues/1090
same job as before: https://github.com/apache/incubator-pegasus/pull/728

When we precisely count data for a large table, it will cost minutes or hours. 
However, it's unnecessarily return key-values from server to client. 

Actually, we just need the count of data. So we just need transfer the count of data from server to client, but not the detailed data.

**In my test, it will 2x on onebox faster than before.**
for my bench on onebox：
1> before
wecom-temp-d7e2a1e011357a0ca4c3ad44d910e69e

2> this commit
image

##### Tests 

- Unit test

##### Related changes

### DIFF
--- a/idl/rrdb.thrift
+++ b/idl/rrdb.thrift
@@ -295,7 +295,7 @@ struct scan_response
     4:i32           app_id;
     5:i32           partition_index;
     6:string        server;
-    7:optional i32  kv_count = -1;
+    7:optional i32  kv_count;
 }
 
 struct duplicate_request
--- a/idl/rrdb.thrift
+++ b/idl/rrdb.thrift
@@ -285,7 +285,6 @@ struct get_scanner_request
 struct scan_request
 {
     1:i64           context_id;
-    2:optional bool only_return_count;
 }
 
 struct scan_response
--- a/idl/rrdb.thrift
+++ b/idl/rrdb.thrift
@@ -279,7 +279,7 @@ struct get_scanner_request
     11:optional bool    validate_partition_hash;
     12:optional bool    return_expire_ts;
     13:optional bool full_scan; // true means client want to build 'full scan' context with the server side, false otherwise
-    14:optional bool only_return_count;
+    14:optional bool only_return_count = false;
 }
 
 struct scan_request
--- a/idl/rrdb.thrift
+++ b/idl/rrdb.thrift
@@ -279,11 +279,13 @@ struct get_scanner_request
     11:optional bool    validate_partition_hash;
     12:optional bool    return_expire_ts;
     13:optional bool full_scan; // true means client want to build 'full scan' context with the server side, false otherwise
+    14:optional bool only_return_count;
 }
 
 struct scan_request
 {
     1:i64           context_id;
+    2:optional bool only_return_count;
 }
 
 struct scan_response
@@ -294,6 +296,7 @@ struct scan_response
     4:i32           app_id;
     5:i32           partition_index;
     6:string        server;
+    7:optional i32  kv_count = -1;
 }
 
 struct duplicate_request
--- a/src/client_lib/pegasus_client_impl.h
+++ b/src/client_lib/pegasus_client_impl.h
@@ -247,6 +247,13 @@ public:
 
     static void init_error();
 
+    enum class async_scan_type : char
+    {
+        normal,
+        count_only,
+        count_only_finished
+    };
+
     class pegasus_scanner_impl : public pegasus_scanner
     {
     public:
@@ -296,7 +303,7 @@ public:
         volatile bool _rpc_started;
         bool _validate_partition_hash;
         bool _full_scan;
-        bool _already_add_count;
+        async_scan_type _type;
 
         void _async_next_internal();
         void _start_scan();
--- a/src/client_lib/pegasus_client_impl.h
+++ b/src/client_lib/pegasus_client_impl.h
@@ -296,7 +296,6 @@ public:
         volatile bool _rpc_started;
         bool _validate_partition_hash;
         bool _full_scan;
-        bool _only_calculate_count;
         bool _already_add_count;
 
         void _async_next_internal();
--- a/src/client_lib/pegasus_client_impl.h
+++ b/src/client_lib/pegasus_client_impl.h
@@ -249,9 +249,9 @@ public:
 
     enum class async_scan_type : char
     {
-        normal,
-        count_only,
-        count_only_finished
+        NORMAL,
+        COUNT_ONLY,
+        COUNT_ONLY_FINISHED
     };
 
     class pegasus_scanner_impl : public pegasus_scanner
--- a/src/client_lib/pegasus_client_impl.h
+++ b/src/client_lib/pegasus_client_impl.h
@@ -247,21 +247,15 @@ public:
 
     static void init_error();
 
-    enum class async_scan_type : char
-    {
-        NORMAL,
-        COUNT_ONLY,
-        COUNT_ONLY_FINISHED
-    };
-
     class pegasus_scanner_impl : public pegasus_scanner
     {
     public:
         int next(std::string &hashkey,
                  std::string &sortkey,
                  std::string &value,
-                 internal_info *info = nullptr,
-                 int32_t *count = nullptr) override;
+                 internal_info *info = nullptr) override;
+
+        int next(int32_t &count, internal_info *info = nullptr) override;
 
         void async_next(async_scan_next_callback_t &&) override;
 
@@ -285,6 +279,13 @@ public:
                              bool full_scan);
 
     private:
+        enum class async_scan_type : char
+        {
+            NORMAL,
+            COUNT_ONLY,
+            COUNT_ONLY_FINISHED
+        };
+
         ::dsn::apps::rrdb_client *_client;
         ::dsn::blob _start_key;
         ::dsn::blob _stop_key;
@@ -330,13 +331,17 @@ private:
 
         void async_next(async_scan_next_callback_t &&callback) override;
 
+        int next(int32_t &count, internal_info *info = nullptr) override
+        {
+            return _p->next(count, info);
+        }
+
         int next(std::string &hashkey,
                  std::string &sortkey,
                  std::string &value,
-                 internal_info *info,
-                 int32_t *count = nullptr) override
+                 internal_info *info) override
         {
-            return _p->next(hashkey, sortkey, value, info, count);
+            return _p->next(hashkey, sortkey, value, info);
         }
     };
 
--- a/src/client_lib/pegasus_client_impl.h
+++ b/src/client_lib/pegasus_client_impl.h
@@ -253,12 +253,8 @@ public:
         int next(std::string &hashkey,
                  std::string &sortkey,
                  std::string &value,
-                 internal_info *info = nullptr) override;
-
-        int next(std::string &hashkey,
-                 std::string &sortkey,
-                 std::string &value,
-                 int32_t &count_number) override;
+                 internal_info *info = nullptr,
+                 int32_t *count = nullptr) override;
 
         void async_next(async_scan_next_callback_t &&) override;
 
@@ -329,17 +325,10 @@ private:
         int next(std::string &hashkey,
                  std::string &sortkey,
                  std::string &value,
-                 int32_t &count_number) override
+                 internal_info *info,
+                 int32_t *count = nullptr) override
         {
-            return _p->next(hashkey, sortkey, value, count_number);
-        }
-
-        int next(std::string &hashkey,
-                 std::string &sortkey,
-                 std::string &value,
-                 internal_info *info) override
-        {
-            return _p->next(hashkey, sortkey, value, info);
+            return _p->next(hashkey, sortkey, value, info, count);
         }
     };
 
--- a/src/client_lib/pegasus_client_impl.h
+++ b/src/client_lib/pegasus_client_impl.h
@@ -296,6 +296,8 @@ public:
         volatile bool _rpc_started;
         bool _validate_partition_hash;
         bool _full_scan;
+        bool _only_calculate_count;
+        bool _already_add_count;
 
         void _async_next_internal();
         void _start_scan();
--- a/src/client_lib/pegasus_client_impl.h
+++ b/src/client_lib/pegasus_client_impl.h
@@ -255,6 +255,11 @@ public:
                  std::string &value,
                  internal_info *info = nullptr) override;
 
+        int next(std::string &hashkey,
+                 std::string &sortkey,
+                 std::string &value,
+                 int32_t &count_number) override;
+
         void async_next(async_scan_next_callback_t &&) override;
 
         bool safe_destructible() const override;
@@ -287,6 +292,7 @@ public:
         std::vector<::dsn::apps::key_value> _kvs;
         internal_info _info;
         int32_t _p;
+        int32_t _kv_count;
 
         int64_t _context;
         mutable ::dsn::zlock _lock;
@@ -319,6 +325,14 @@ private:
         pegasus_scanner_impl_wrapper(pegasus_scanner *p) : _p(p) {}
 
         void async_next(async_scan_next_callback_t &&callback) override;
+
+        int next(std::string &hashkey,
+                 std::string &sortkey,
+                 std::string &value,
+                 int32_t &count_number) override
+        {
+            return _p->next(hashkey, sortkey, value, count_number);
+        }
 
         int next(std::string &hashkey,
                  std::string &sortkey,
--- a/src/client_lib/pegasus_scanner_impl.cpp
+++ b/src/client_lib/pegasus_scanner_impl.cpp
@@ -55,40 +55,14 @@ pegasus_client_impl::pegasus_scanner_impl::pegasus_scanner_impl(::dsn::apps::rrd
       _rpc_started(false),
       _validate_partition_hash(validate_partition_hash),
       _full_scan(full_scan)
-
 {
 }
 
 int pegasus_client_impl::pegasus_scanner_impl::next(std::string &hashkey,
                                                     std::string &sortkey,
                                                     std::string &value,
-                                                    int32_t &count_number)
-{
-    ::dsn::utils::notify_event op_completed;
-    int ret = -1;
-    auto callback = [&](int err,
-                        std::string &&hash,
-                        std::string &&sort,
-                        std::string &&val,
-                        internal_info &&ii,
-                        uint32_t expire_ts_seconds,
-                        int32_t kv_count) {
-        ret = err;
-        hashkey = std::move(hash);
-        sortkey = std::move(sort);
-        value = std::move(val);
-        count_number = kv_count;
-        op_completed.notify();
-    };
-    async_next(std::move(callback));
-    op_completed.wait();
-    return ret;
-}
-
-int pegasus_client_impl::pegasus_scanner_impl::next(std::string &hashkey,
-                                                    std::string &sortkey,
-                                                    std::string &value,
-                                                    internal_info *info)
+                                                    internal_info *info,
+                                                    int32_t *count)
 {
     ::dsn::utils::notify_event op_completed;
     int ret = -1;
@@ -105,6 +79,9 @@ int pegasus_client_impl::pegasus_scanner_impl::next(std::string &hashkey,
         value = std::move(val);
         if (info) {
             (*info) = std::move(ii);
+        }
+        if (count) {
+            (*count) = kv_count;
         }
         op_completed.notify();
     };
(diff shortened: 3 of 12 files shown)


src/shell/command_helper.h
```diff
@@ -499,36 +501,40 @@ inline void scan_data_next(scan_data_context *context)
...
-                                static_cast<uint32_t>(histogram_type::VALUE_SIZE), value_size);
-
-                            long row_size = hash_key_size + sort_key_size + value_size;
-                            context->statistics->measureTime(
-                                static_cast<uint32_t>(histogram_type::ROW_SIZE), row_size);
-
-                            if (context->top_count > 0) {
-                                context->top_rows.push(
-                                    std::move(hash_key), std::move(sort_key), row_size);
+                        if (kv_count == -1) {
```
empiredan, line 504: Does `kv_count == -1` means `only_return_count` ? Can `only_return_count` be used as the condition to be checked ?

src/shell/command_helper.h
```diff
@@ -499,36 +501,40 @@ inline void scan_data_next(scan_data_context *context)
...
-                            if (hash_key != context->last_hash_key) {
-                                context->split_hash_key_count++;
-                                context->last_hash_key = std::move(hash_key);
+                            if (context->count_hash_key) {
+                                if (hash_key != context->last_hash_key) {
+                                    context->split_hash_key_count++;
+                                    context->last_hash_key = std::move(hash_key);
+                                }
                             }
+                        } else {
```
empiredan, line 536:
```c++
if (...) {
    context->split_rows += kv_count;
    scan_data_next(context);
    break;
}
```

idl/rrdb.thrift
```diff
@@ -279,11 +279,15 @@ struct get_scanner_request
     11:optional bool    validate_partition_hash;
     12:optional bool    return_expire_ts;
     13:optional bool full_scan; // true means client want to build 'full scan' context with the server side, false otherwise
+    14:optional filter_type value_filter_type;
+    15:optional dsn.blob    value_filter_pattern;
```
acelyc111, line 283: Are the two parameters related to this patch?

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -55,7 +55,35 @@ pegasus_client_impl::pegasus_scanner_impl::pegasus_scanner_impl(::dsn::apps::rrd
       _rpc_started(false),
       _validate_partition_hash(validate_partition_hash),
       _full_scan(full_scan)
+
```
acelyc111, line 58: remove the blank line

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -55,7 +55,35 @@ pegasus_client_impl::pegasus_scanner_impl::pegasus_scanner_impl(::dsn::apps::rrd
...
+{
+}
+
+int pegasus_client_impl::pegasus_scanner_impl::next(std::string &hashkey,
```
acelyc111, line 62: why not reuse the exist `pegasus_client_impl::pegasus_scanner_impl::next` ?

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -162,13 +194,15 @@ void pegasus_client_impl::pegasus_scanner_impl::_async_next_internal()
...
 
         // valid data got
-        std::string hash_key, sort_key;
-        pegasus_restore_key(_kvs[_p].key, hash_key, sort_key);
-        std::string value(_kvs[_p].value.data(), _kvs[_p].value.length());
-        uint32_t expire_ts_seconds = _kvs[_p].__isset.expire_ts_seconds
-                                         ? static_cast<uint32_t>(_kvs[_p].expire_ts_seconds)
-                                         : 0;
-
+        std::string hash_key, sort_key, value = "";
```
acelyc111, line 197:
```suggestion
        std::string hash_key, sort_key, value;
```

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -162,13 +194,15 @@ void pegasus_client_impl::pegasus_scanner_impl::_async_next_internal()
...
+        uint32_t expire_ts_seconds = 0;
+        if (_kv_count == -1) {
```
acelyc111, line 199: it suggest to add some comment to explain what dose `_kv_count == -1` mean

idl/rrdb.thrift
```diff
@@ -279,11 +279,15 @@ struct get_scanner_request
...
+    16:optional bool    only_return_count = false;
 }
 
 struct scan_request
 {
     1:i64           context_id;
+    2:optional bool only_return_count;
```
acelyc111, line 290: the `get_scanner_request ` has a `only_return_count` field, why add another one to `scan_request`?

src/server/pegasus_server_impl.cpp
```diff
@@ -1389,6 +1419,11 @@ void pegasus_server_impl::on_scan(scan_rpc rpc)
             it->Next();
         }
 
+        if (request.only_return_count) {
```
acelyc111, line 1422: you can use the flag in pegasus_scan_context, right?

src/shell/commands/data_operations.cpp
```diff
@@ -2397,6 +2401,15 @@ bool count_data(command_executor *e, shell_context *sc, arguments args)
         options.no_value = false;
     else
         options.no_value = true;
+
+    if (diff_hash_key || stat_size || value_filter_type != pegasus::pegasus_client::FT_NO_FILTER ||
```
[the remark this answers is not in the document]
GehaFearless: I will make another one pull request laster.

src/server/pegasus_server_impl.cpp
```diff
@@ -2316,24 +2332,28 @@ pegasus_server_impl::append_key_value_for_scan(std::vector<::dsn::apps::key_valu
             return range_iteration_state::kFiltered;
         }
     }
-    std::shared_ptr<char> key_buf(::dsn::utils::make_shared_array<char>(raw_key.length()));
-    ::memcpy(key_buf.get(), raw_key.data(), raw_key.length());
-    kv.key.assign(std::move(key_buf), 0, raw_key.length());
+    if (fill_value) {
```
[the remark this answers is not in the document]
GehaFearless: `return range_iteration_state::kNormal` All branches use it, So I think is unnecessary

src/shell/command_helper.h
GehaFearless: In order to cope with the different versions of the match, I think this is the best

src/server/pegasus_server_impl.cpp
```diff
@@ -1449,7 +1466,10 @@ void pegasus_server_impl::on_scan(scan_rpc rpc)
         resp.error = rocksdb::Status::Code::kNotFound;
     }
 
-    _cu_calculator->add_scan_cu(req, resp.error, resp.kvs);
+    // abandon calculate capacity unit
+    if (only_return_count) {
```
WHBANG, line 1470: if (!only_return_count)

src/server/pegasus_server_impl.cpp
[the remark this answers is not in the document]
acelyc111: Sorry, I didn't get your point..
I think short-circut return is often a recommand code style too.

src/server/pegasus_server_impl.cpp
```diff
@@ -2274,7 +2294,8 @@ pegasus_server_impl::append_key_value_for_scan(std::vector<::dsn::apps::key_valu
                                                uint32_t epoch_now,
                                                bool no_value,
                                                bool request_validate_hash,
-                                               bool request_expire_ts)
+                                               bool request_expire_ts,
+                                               bool fill_value)
```
acelyc111, line 2298: The function name `append_key_value_for_scan` is not suitable now, because the 'append' operation will not happen when fill_value is false.
You can do a refactor to separete then key-value valicate part and append part into two functions.

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -162,13 +168,16 @@ void pegasus_client_impl::pegasus_scanner_impl::_async_next_internal()
...
-        pegasus_restore_key(_kvs[_p].key, hash_key, sort_key);
-        std::string value(_kvs[_p].value.data(), _kvs[_p].value.length());
-        uint32_t expire_ts_seconds = _kvs[_p].__isset.expire_ts_seconds
-                                         ? static_cast<uint32_t>(_kvs[_p].expire_ts_seconds)
-                                         : 0;
-
+        std::string hash_key, sort_key, value;
+        uint32_t expire_ts_seconds = 0;
+        // _kv_count == -1 means req just want to get data counts, not include data value
+        if (_kv_count == -1) {
```
acelyc111, line 174: should be _kv_count != -1 ?

src/server/pegasus_server_impl.cpp
```diff
@@ -1389,6 +1401,11 @@ void pegasus_server_impl::on_scan(scan_rpc rpc)
             it->Next();
         }
 
+        if (!only_return_count) {
+            resp.kvs.emplace_back(::dsn::apps::key_value());
```
acelyc111, line 1405: Same

src/server/pegasus_server_impl.cpp
```diff
@@ -1297,7 +1303,10 @@ void pegasus_server_impl::on_get_scanner(get_scanner_rpc rpc)
         _pfc_recent_filter_count->add(filter_count);
     }
 
-    _cu_calculator->add_scan_cu(req, resp.error, resp.kvs);
+    // abandon calculate capacity unit
+    if (!request.only_return_count) {
```
acelyc111, line 1307: We should also calculate CU, and you have to update add_scan_cu, add some small CU at least.

src/server/pegasus_server_impl.cpp
```diff
@@ -1389,6 +1401,11 @@ void pegasus_server_impl::on_scan(scan_rpc rpc)
             it->Next();
         }
 
+        if (!only_return_count) {
```
acelyc111, line 1404: should be `if (only_return_count)` ?

src/server/pegasus_server_impl.cpp
```diff
@@ -1198,7 +1198,8 @@ void pegasus_server_impl::on_get_scanner(get_scanner_rpc rpc)
             epoch_now,
             request.no_value,
             request.__isset.validate_partition_hash ? request.validate_partition_hash : true,
-            return_expire_ts);
+            return_expire_ts,
+            !request.only_return_count);
```
acelyc111, line 1202: I found you reverse the value every calls of append_key_value_for_scan, why not use it directly?

acelyc111:
Could you give some performance comparation with the old version ?

src/server/pegasus_server_impl.cpp
```diff
@@ -1187,21 +1187,23 @@ void pegasus_server_impl::on_get_scanner(get_scanner_rpc rpc)
...
             epoch_now,
-            request.no_value,
-            request.__isset.validate_partition_hash ? request.validate_partition_hash : true,
-            return_expire_ts);
+            request.__isset.validate_partition_hash ? request.validate_partition_hash : true);
+
         switch (state) {
         case range_iteration_state::kNormal:
             count++;
+            if (!request.only_return_count) {
```
acelyc111, line 1203: Should judge wether request.__isset.only_return_count is set at first, client
maybe old version, it have to keep compatitable.

src/client_lib/pegasus_scanner_impl.cpp
acelyc111: You should explain the case of `_kv_count == -1` and all other cases, `_kv_count > -1` is only part of case you have to explain.

acelyc111: How about add another variable to indicate directly what you want to do?

src/server/pegasus_server_impl.cpp
```diff
@@ -1221,6 +1222,10 @@ void pegasus_server_impl::on_get_scanner(get_scanner_rpc rpc)
 
         it->Next();
     }
+    if (request.only_return_count) {
+        resp.kvs.emplace_back(::dsn::apps::key_value());
```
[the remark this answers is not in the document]
acelyc111: It's too tricky...
How much work if refactor it?

GehaFearless:
> Could you give some performance comparation with the old version ?
for my bench on onebox：
1> before
wecom-temp-d7e2a1e011357a0ca4c3ad44d910e69e

2> this commit
image

src/server/pegasus_server_impl.cpp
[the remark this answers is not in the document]
GehaFearless: I did it by adding a member variable _already_add_count

src/server/pegasus_server_impl.cpp
```diff
@@ -1312,6 +1319,7 @@ void pegasus_server_impl::on_scan(scan_rpc rpc)
     resp.app_id = _gpid.get_app_id();
     resp.partition_index = _gpid.get_partition_index();
     resp.server = _primary_address;
+    bool only_return_count = false;
```
acelyc111, line 1322: Define it in `if (context) {...}` scope is enough.

src/test/function_test/test_scan.cpp
```diff
@@ -168,6 +168,34 @@ class scan : public testing::Test
...
+        ASSERT_NE(nullptr, scanner);
+        int32_t kv_count;
+        while (PERR_OK == (ret = (scanner->next(hash_key, sort_key, value, nullptr, &kv_count)))) {
+            data_counts += kv_count;
+        }
+        ASSERT_EQ(PERR_SCAN_COMPLETE, ret) << "Error occurred when scan. error="
+                                           << client->get_error_string(ret);
+        delete scanner;
+    }
+    ASSERT_EQ(10990, data_counts);
```
acelyc111, line 196: where dose 10990 come from? it would be better to use a variable from previous variables, do not use a magic number.

src/shell/commands/data_operations.cpp
```diff
@@ -2397,6 +2398,15 @@ bool count_data(command_executor *e, shell_context *sc, arguments args)
...
+        sort_key_filter_type == pegasus::pegasus_client::FT_MATCH_EXACT) {
+        options.only_return_count = false;
```
acelyc111, line 2404: The jugement is complex, add comments to explain why.

idl/rrdb.thrift
```diff
@@ -279,6 +279,7 @@ struct get_scanner_request
     11:optional bool    validate_partition_hash;
     12:optional bool    return_expire_ts;
     13:optional bool full_scan; // true means client want to build 'full scan' context with the server side, false otherwise
+    14:optional bool only_return_count;
```
acelyc111, line 282: set the default value to false?

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -261,6 +278,18 @@ void pegasus_client_impl::pegasus_scanner_impl::_on_scan_response(::dsn::error_c
...
             _p = -1;
             _context = response.context_id;
+            // 1. kv_count exist on response means server is newer version (added counting size only
+            // implementation)
+            //   1> kv_count == -1 indicates response still have key and value data
+            //   2> kv_count > -1 indicates response only have kv size count, but not key && value
+            // 2. kv_count is not existed means server is older version
+            if (response.__isset.kv_count) {
+                if (response.kv_count != -1) {
+                    _only_calculate_count = true;
```
acelyc111, line 288: Request is sent from client, we can init _only_calculate_count when send request, right?

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -51,17 +51,20 @@ pegasus_client_impl::pegasus_scanner_impl::pegasus_scanner_impl(::dsn::apps::rrd
       _options(options),
       _splits_hash(std::move(hash)),
       _p(-1),
+      _kv_count(-1),
       _context(SCAN_CONTEXT_ID_COMPLETED),
       _rpc_started(false),
       _validate_partition_hash(validate_partition_hash),
-      _full_scan(full_scan)
+      _full_scan(full_scan),
+      _already_add_count(true)
```
acelyc111, line 59: why the default value is true?

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -261,6 +277,17 @@ void pegasus_client_impl::pegasus_scanner_impl::_on_scan_response(::dsn::error_c
...
             _p = -1;
             _context = response.context_id;
+            // 1. kv_count exist on response means server is newer version (added counting size only
+            // implementation)
+            //   1> kv_count == -1 indicates response still have key and value data
+            //   2> kv_count > -1 indicates response only have kv size count, but not key && value
+            // 2. kv_count is not existed means server is older version
+            if (response.__isset.kv_count) {
+                if (response.kv_count != -1) {
+                    _already_add_count = false;
```
acelyc111, line 287: I don't understand what does `_already_add_count` mean, why set it to false here

src/shell/commands/data_operations.cpp
```diff
@@ -2333,6 +2333,7 @@ bool count_data(command_executor *e, shell_context *sc, arguments args)
         tp.output(std::cout, tp_output_format::kTabular);
         return true;
     }
+    options.only_return_count = true;
```
acelyc111, line 2336: why set it here? only need to set it when `-c` flag is set, right?

src/test/function_test/test_scan.cpp
```diff
@@ -168,6 +168,38 @@ class scan : public testing::Test
...
+        ASSERT_NE(nullptr, scanner);
+        int32_t kv_count;
+        while (PERR_OK == (ret = (scanner->next(hash_key, sort_key, value, nullptr, &kv_count)))) {
+            data_counts += kv_count;
+        }
+        ASSERT_EQ(PERR_SCAN_COMPLETE, ret) << "Error occurred when scan. error="
+                                           << client->get_error_string(ret);
+        delete scanner;
+    }
+    int data_sizes = 0;
```
acelyc111, line 196: better naming as xxx_count

src/client_lib/pegasus_scanner_impl.cpp
```diff
@@ -261,6 +278,17 @@ void pegasus_client_impl::pegasus_scanner_impl::_on_scan_response(::dsn::error_c
             _kvs = std::move(response.kvs);
             _p = -1;
             _context = response.context_id;
+            // 1. kv_count exist on response means server is newer version (added counting size only
+            // implementation)
+            //   1> kv_count == -1 indicates response still have key and value data
+            //   2> kv_count > -1 indicates response only have kv size count, but not key && value
+            // 2. kv_count is not existed means server is older version
+            if (response.__isset.kv_count) {
+                if (response.kv_count != -1) {
```
acelyc111, line 287: in which case that response.__isset.kv_count is set but response.kv_count == -1?

GehaFearless: case value_filter be used

src/client_lib/pegasus_client_impl.h
```diff
@@ -247,13 +247,21 @@ class pegasus_client_impl : public pegasus_client
 
     static void init_error();
 
+    enum class async_scan_type : char
```
acelyc111, line 250: Do not expose it to public in header, make it as a private of `class pegasus_scanner_impl` is enough?

src/client_lib/pegasus_client_impl.h
```diff
@@ -247,13 +247,21 @@ class pegasus_client_impl : public pegasus_client
...
+
     class pegasus_scanner_impl : public pegasus_scanner
     {
     public:
         int next(std::string &hashkey,
                  std::string &sortkey,
                  std::string &value,
-                 internal_info *info = nullptr) override;
+                 internal_info *info = nullptr,
+                 int32_t *count = nullptr) override;
```
acelyc111, line 264: hashkey-sortkey-value and count will not be returned at the same time? If not, separete it into two functions would be better. one is:
```
        int next(std::string &hashkey,
                 std::string &sortkey,
                 std::string &value,
                 internal_info *info = nullptr)
```
the other is 
```
        int next(int32_t &count,
                 internal_info *info = nullptr)
```

acelyc111 (dismissed):

src/shell/commands/data_operations.cpp
```diff
@@ -2397,6 +2397,17 @@ bool count_data(command_executor *e, shell_context *sc, arguments args)
         options.no_value = false;
     else
         options.no_value = true;
+
+    // cuz option only_return_count not return data(hashkey&sortkey&value) to client
+    // all of this options need check data on client side
```
empiredan, line 2402: I think maybe it's better to explain each of conditions here ?
```suggestion 
    // Decide whether real data should be returned to client. Once the real data is
    // decided not to be returned to client side: option `only_return_count` will be
    // used.
```

acelyc111 (approved):

empiredan (approved):

Issue #1095, opened 2022-08-02
Pegasus unit test failed for download hadoop failed


WHBANG:
## Bug Report
https://github.com/apache/incubator-pegasus/runs/7605983359?check_suite_focus=true

6650K .......... .......... .......... .......... ..........  2%  143K 9m22s
  6700K .......... .......... .......... .......... ..........  2%  137K 9m30s
  6750K .......... .......... .......... .......... ..........  2%  198K 9m35s
  6800K .......... .......... .......... .......... ..........  2%  189K 9m39s
  6850K .......... .......... .......... .......... ..........  2%  190K 9m44s
  6900K .......... .......... .......... .......... ..........  2%  189K 9m49s
  6950K .......... .......... .......... .......... ..........  2%  186K 9m53s
  7000K .......... .......... .......... .......... ..........  2%  121K 10m3s
  7050K .......... .......... .......... .......... ..........  2%  180K 10m7s
  7100K .......... .......... .......... .......... ..........  2%  180K 10m12s
  7[150](https://github.com/apache/incubator-pegasus/runs/7605983261?check_suite_focus=true#step:10:151)K .......... .......... .......... .......... ..........  2%  180K 10m17s
  7200K .......... .......... .......... .......... ..........  3% 88.3K 10m30s
  7250K .......... .......... .......... .......... ..........  3%  159K 10m36s
  7300K .......... .......... .......... .......... .....       3% 71.8K=21s

2022-08-01 07:38:16 (357 KB/s) - Read error at byte 7521757/246123562 (Success). Giving up.

ERROR: download hadoop failed
Error: Process completed with exit code 1.

WHBANG:
https://github.com/apache/incubator-pegasus/runs/7626009179?check_suite_focus=true
Another download failure problem：

Run export LD_LIBRARY_PATH=`pwd`/thirdparty/output/lib:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server
Test start time: Tue Aug  2 07:20:24 Asia 2022
[...]

PR #1099, opened 2022-08-03
refactor(build): remove 'rdsn' from build and test commands

acelyc111:

This patch is planning to remove 'rdsn' from build and test commands. After doing this, the CI will last longer, so I have to the following work to reduce the effect:
1. Refactor Cpp CI yaml:
  a. shorten names, it will be easier to read on Web
  b. split 'test' into 'build' and 'test'
  c. remove some self-hosted runner configs
  d. use `actions/cache@v3` to speed up build staged
  e. remove changes check on rdsn directory
  f. make -j parameter with N cpus instead N/2
  g. use `actions/upload-artifact@v3` and `actions/download-artifact@v3` to pass test artifacts from build job to test job
  h. use `matrix` to run test parallely to reduce test last time, and make re-run cheaper (and also add some missing test cases in run.sh)
  i. remove LSAN since it's included by ASAN
2. Cmake changes:
  a. unify rdsn and pegasus cmakelists
  b. add ccache detection and configs in cmake
  c. adjust includes and links
  d. make `pegasus_geo_test` can be run by shell which is the unify way
  e. remove src/rdsn/CMakeLists.txt
3. shell scripts update
  a. remove 'rdsn' from run.sh for both build and test functions
  b. re-org thrift generating and git_commit.h generating
  c. use different build dirs for different build types, then you can switch build types with lower cost (NOTE: you have to remove `src/builder` directory after this patch)
  d. make `wget` more robust (by `wget -T 10 -t 5`) https://github.com/apache/incubator-pegasus/issues/1095
4. unit test fixes after this patch because it will run more tests
  a. move templates to header files (`src/rdsn/src/meta/meta_service.h`)
  b. other cmake and cpp files

image


hycdong (dismissed):

hycdong:
LGTM~
Besides, I have following suggestions:
1. I recommend checking if `pack_server.sh`, `pack_client.sh`, `pack_tools.sh` can work well. In our previous build refactor, those script may have problems after build refactor.
2. I recommend refactoring pegasus_function_test in future. Restrict with test time, some long-time function tests are not running if `on_travis` is set as true.

acelyc111:
> * I recommend checking if `pack_server.sh`, `pack_client.sh`, `pack_tools.sh` can work well. In our previous build refactor, those script may have problems after build refactor.
> * I recommend refactoring pegasus_function_test in future. Restrict with test time, some long-time function tests are not running if `on_travis` is set as true.

1. Thanks for your reminding, I missed that, `pack_server.sh` and `pack_tools.sh` has been added back, `pack_client.sh` is not running before so I didn't add it, I will do it in another patch.
2. Yes, I'll do it in another patch.

hycdong (approved):

levy5307 (approved):

Issue #1025, opened 2022-06-30
BUG: Rocksdb options not changed even if update in Pegasus config file


acelyc111:
Because we will load rocksdb options from disk when bootstrap, even if we update configs in Pegasus' config file, it seems will not take effect on these options.

WHBANG:
i mark

acelyc111:
> 

Thanks!

acelyc111:
> i mark

@WHBANG Thanks!

PR #1108, opened 2022-08-09
fix: rocksdb options not changed even if update in Pegasus config file

WHBANG:
https://github.com/apache/incubator-pegasus/issues/1025

 The specific reason is that if you open an existing rocksdb, the usage scenario will be set to normal first, and then set to the corresponding mode according to the actual usage scenario. This process will change the relevant options, causing rocksdb to execute flush and compact, so in order to avoid this situation (https://github.com/apache/incubator-pegasus/pull/587), rocksdb::LoadLatestOptions is called, and these parameters are initialized with the last value, so the last thing you see is that if the configuration file is modified, and without changing the usage scenario mode, these modified configurations will not take effect after restart.

Options involved:
```
level0_file_num_compaction_trigger
level0_slowdown_writes_trigger
level0_stop_writes_trigger
soft_pending_compaction_bytes_limit
hard_pending_compaction_bytes_limit
disable_auto_compactions
max_compaction_bytes
write_buffer_size
max_write_buffer_number
```
Solution: 
1. After opening rocksdb, calculate the value of the corresponding options according to the usage scenario mode in the environment variable, compare whether there is a change, and call setOpion() if there is a change to set the newly calculated value. 
3. If the usage scenario mode has been changed during this process, the options will be the correct value, and nothing needs to be done.

##### Tests 

- Unit test

-

### DIFF
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -3013,6 +3013,10 @@ void pegasus_server_impl::reset_usage_scenario_options(
 
 void pegasus_server_impl::recalculate_usage_scenario(const rocksdb::ColumnFamilyOptions &cur_opts)
 {
+#define UPDATE_OPTION_IF_NEEDED(option, value)                                                     \
+    if ((value) != cur_opts.option) {                                                              \
+        options["#option"] = std::to_string((value));                                              \
+    }                                                                                              \
     std::unordered_map<std::string, std::string> new_options;
     if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario ||
         ROCKSDB_ENV_USAGE_SCENARIO_PREFER_WRITE == _usage_scenario) {
@@ -3022,11 +3026,8 @@ void pegasus_server_impl::recalculate_usage_scenario(const rocksdb::ColumnFamily
                 new_options["write_buffer_size"] =
                     std::to_string(get_random_nearby(_data_cf_opts.write_buffer_size));
             }
-            if (_data_cf_opts.level0_file_num_compaction_trigger !=
-                cur_opts.level0_file_num_compaction_trigger) {
-                new_options["level0_file_num_compaction_trigger"] =
-                    std::to_string(_data_cf_opts.level0_file_num_compaction_trigger);
-            }
+            UPDATE_OPTION_IF_NEEDED(level0_file_num_compaction_trigger,
+                                    _data_cf_opts.level0_file_num_compaction_trigger);
         } else {
             uint64_t buffer_size = dsn::rand::next_u64(_data_cf_opts.write_buffer_size,
                                                        _data_cf_opts.write_buffer_size * 2);
@@ -3043,63 +3044,26 @@ void pegasus_server_impl::recalculate_usage_scenario(const rocksdb::ColumnFamily
                     std::to_string(std::max<uint64_t>(4UL, max_size / buffer_size));
             }
         }
-        if (_data_cf_opts.level0_slowdown_writes_trigger !=
-            cur_opts.level0_slowdown_writes_trigger) {
-            new_options["level0_slowdown_writes_trigger"] =
-                std::to_string(_data_cf_opts.level0_slowdown_writes_trigger);
-        }
-        if (_data_cf_opts.level0_stop_writes_trigger != cur_opts.level0_stop_writes_trigger) {
-            new_options["level0_stop_writes_trigger"] =
-                std::to_string(_data_cf_opts.level0_stop_writes_trigger);
-        }
-        if (_data_cf_opts.soft_pending_compaction_bytes_limit !=
-            cur_opts.soft_pending_compaction_bytes_limit) {
-            new_options["soft_pending_compaction_bytes_limit"] =
-                std::to_string(_data_cf_opts.soft_pending_compaction_bytes_limit);
-        }
-        if (_data_cf_opts.hard_pending_compaction_bytes_limit !=
-            cur_opts.hard_pending_compaction_bytes_limit) {
-            new_options["hard_pending_compaction_bytes_limit"] =
-                std::to_string(_data_cf_opts.hard_pending_compaction_bytes_limit);
-        }
-        if (_data_cf_opts.disable_auto_compactions != cur_opts.disable_auto_compactions) {
-            new_options["disable_auto_compactions"] = "false";
-        }
-        if (_data_cf_opts.max_compaction_bytes != cur_opts.max_compaction_bytes) {
-            new_options["max_compaction_bytes"] =
-                std::to_string(_data_cf_opts.max_compaction_bytes);
-        }
-        if (_data_cf_opts.max_write_buffer_number != cur_opts.max_write_buffer_number) {
-            new_options["max_write_buffer_number"] =
-                std::to_string(_data_cf_opts.max_write_buffer_number);
-        }
+        UPDATE_OPTION_IF_NEEDED(level0_slowdown_writes_trigger,
+                                _data_cf_opts.level0_slowdown_writes_trigger);
+        UPDATE_OPTION_IF_NEEDED(level0_stop_writes_trigger,
+                                _data_cf_opts.level0_stop_writes_trigger);
+        UPDATE_OPTION_IF_NEEDED(soft_pending_compaction_bytes_limit,
+                                _data_cf_opts.soft_pending_compaction_bytes_limit);
+        UPDATE_OPTION_IF_NEEDED(hard_pending_compaction_bytes_limit,
+                                _data_cf_opts.hard_pending_compaction_bytes_limit);
+        UPDATE_OPTION_IF_NEEDED(disable_auto_compactions, "false");
+        UPDATE_OPTION_IF_NEEDED(max_compaction_bytes, _data_cf_opts.max_compaction_bytes);
+        UPDATE_OPTION_IF_NEEDED(max_write_buffer_number, _data_cf_opts.max_write_buffer_number);
     } else {
         // ROCKSDB_ENV_USAGE_SCENARIO_BULK_LOAD
-        if (_data_cf_opts.level0_file_num_compaction_trigger !=
-            cur_opts.level0_file_num_compaction_trigger) {
-            new_options["level0_file_num_compaction_trigger"] = "1000000000";
-        }
-        if (_data_cf_opts.level0_slowdown_writes_trigger !=
-            cur_opts.level0_slowdown_writes_trigger) {
-            new_options["level0_slowdown_writes_trigger"] = "1000000000";
-        }
-        if (_data_cf_opts.level0_stop_writes_trigger != cur_opts.level0_stop_writes_trigger) {
-            new_options["level0_stop_writes_trigger"] = "1000000000";
-        }
-        if (_data_cf_opts.soft_pending_compaction_bytes_limit !=
-            cur_opts.soft_pending_compaction_bytes_limit) {
-            new_options["soft_pending_compaction_bytes_limit"] = "0";
-        }
-        if (_data_cf_opts.hard_pending_compaction_bytes_limit !=
-            cur_opts.hard_pending_compaction_bytes_limit) {
-            new_options["hard_pending_compaction_bytes_limit"] = "0";
-        }
-        if (_data_cf_opts.disable_auto_compactions != cur_opts.disable_auto_compactions) {
-            new_options["disable_auto_compactions"] = "true";
-        }
-        if (_data_cf_opts.max_compaction_bytes != cur_opts.max_compaction_bytes) {
-            new_options["max_compaction_bytes"] = std::to_string(static_cast<uint64_t>(1) << 60);
-        }
+        UPDATE_OPTION_IF_NEEDED(level0_file_num_compaction_trigger, 1000000000);
+        UPDATE_OPTION_IF_NEEDED(level0_slowdown_writes_trigger, 1000000000);
+        UPDATE_OPTION_IF_NEEDED(level0_stop_writes_trigger, 1000000000);
+        UPDATE_OPTION_IF_NEEDED(soft_pending_compaction_bytes_limit, 0);
+        UPDATE_OPTION_IF_NEEDED(hard_pending_compaction_bytes_limit, 0);
+        UPDATE_OPTION_IF_NEEDED(disable_auto_compactions, "true");
+        UPDATE_OPTION_IF_NEEDED(max_compaction_bytes, static_cast<uint64_t>(1) << 60);
         if (!check_value_if_nearby(_data_cf_opts.write_buffer_size, cur_opts.write_buffer_size)) {
             new_options["write_buffer_size"] =
                 std::to_string(get_random_nearby(_data_cf_opts.write_buffer_size * 4));
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -3011,7 +3011,7 @@ void pegasus_server_impl::reset_usage_scenario_options(
     target_opts->max_write_buffer_number = base_opts.max_write_buffer_number;
 }
 
-bool pegasus_server_impl::recalculate_usage_scenario(const rocksdb::ColumnFamilyOptions &cur_opts)
+void pegasus_server_impl::recalculate_usage_scenario(const rocksdb::ColumnFamilyOptions &cur_opts)
 {
     std::unordered_map<std::string, std::string> new_options;
     if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario ||
--- a/src/server/pegasus_server_impl.cpp
+++ b/src/server/pegasus_server_impl.cpp
@@ -3051,7 +3051,7 @@ void pegasus_server_impl::recalculate_data_cf_options(
     if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario ||
         ROCKSDB_ENV_USAGE_SCENARIO_PREFER_WRITE == _usage_scenario) {
         if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario) {
-            UPDATE_OPTION_IF_NOT_NEARBY(write_buffer_size, data_cf_opts.write_buffer_size);
+            UPDATE_OPTION_IF_NOT_NEARBY(write_buffer_size, _data_cf_opts.write_buffer_size);
             UPDATE_OPTION_IF_NEEDED(level0_file_num_compaction_trigger);
         } else {
             uint64_t buffer_size = dsn::rand::next_u64(_data_cf_opts.write_buffer_size,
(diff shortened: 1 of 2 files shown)


src/server/pegasus_server_impl.h
```diff
@@ -325,6 +328,15 @@ class pegasus_server_impl : public pegasus_read_service
         return dsn::rand::next_u64(base_value - gap, base_value + gap);
     }
 
+    // return true if value in range of [0.75, 1.25] * base_value
+    bool check_value_if_nearby(uint64_t base_value, uint64_t check_value)
+    {
+        uint64_t gap = base_value / 4;
+        uint64_t actual_gap =
+            (base_value < check_value) ? check_value - base_value : base_value - check_value;
```
acelyc111, line 336: std::abs(check_value - base_value);

src/server/pegasus_server_impl.cpp
```diff
@@ -2634,6 +2634,10 @@ void pegasus_server_impl::update_usage_scenario(const std::map<std::string, std:
                            old_usage_scenario,
                            new_usage_scenario);
         }
+    } else {
+        // When an old db is opened and the conf is changed, the options related to usage scenario
+        // need to be recalculated with new values.
+        recalculate_usage_scenario(_tmp_data_cf_opts);
```
acelyc111, line 2640: Usage scenario is not changed, maybe naming as recalculate_cf_options?

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3011,112 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+{
+    std::unordered_map<std::string, std::string> new_options;
+    if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario ||
+        ROCKSDB_ENV_USAGE_SCENARIO_PREFER_WRITE == _usage_scenario) {
+        if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario) {
+            if (!check_value_if_nearby(_data_cf_opts.write_buffer_size,
+                                       cur_opts.write_buffer_size)) {
+                new_options["write_buffer_size"] =
+                    std::to_string(get_random_nearby(_data_cf_opts.write_buffer_size));
+            }
```
acelyc111, line 3024: There are many duplicate code, use a macro here will be easier to read.
```
#define UPDATE_OPTION_IF_NEEDED(option, new_value) \
  ...
```
`#` and `##` can be used in this macro.

src/server/pegasus_server_impl.h
WHBANG: in this case, the following error will be reported:
```
error: call of overloaded ‘abs(uint64_t)’ is ambiguous
```

src/server/pegasus_server_impl.cpp
```diff
@@ -1560,9 +1560,10 @@ dsn::error_code pegasus_server_impl::start(int argc, char **argv)
 
     ddebug_replica("start to open rocksDB's rdb({})", rdb_path);
 
-    // Here we create a `tmp_data_cf_opts` because we don't want to modify `_data_cf_opts`, which
+    // Here we create a `_tmp_data_cf_opts` because we don't want to modify `_data_cf_opts`, which
     // will be used elsewhere.
-    rocksdb::ColumnFamilyOptions tmp_data_cf_opts = _data_cf_opts;
+    _tmp_data_cf_opts = _data_cf_opts;
```
acelyc111, line 1565: 'tmp' is not good name for a member variable.
hosw about rename '_data_cf_opts' to '_base_data_cf_opts', and rename '_tmp_data_cf_opts' to '_table_data_cf_opts' ?

src/server/pegasus_server_impl.cpp
```diff
@@ -2634,6 +2635,10 @@ void pegasus_server_impl::update_usage_scenario(const std::map<std::string, std:
                            old_usage_scenario,
                            new_usage_scenario);
         }
+    } else {
+        // When an old db is opened and the conf is changed, the options related to usage scenario
+        // need to be recalculated with new values.
+        recalculate_data_cf_options(_tmp_data_cf_opts);
```
acelyc111, line 2641: If I didn't miss any thing, I found this function will not be called when open an exist db, right?

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,97 @@ void pegasus_server_impl::reset_usage_scenario_options(
     target_opts->max_write_buffer_number = base_opts.max_write_buffer_number;
 }
 
+void pegasus_server_impl::recalculate_data_cf_options(
+    const rocksdb::ColumnFamilyOptions &cur_data_cf_opts)
+{
+#define UPDATE_OPTION_IF_NEEDED(option, value, str_value)                                          \
+    if ((value) != cur_data_cf_opts.option) {                                                      \
+        new_options[#option] = (str_value);                                                        \
+    }
```
acelyc111, line 3022: How about:
```
#define UPDATE_OPTION_IF_NEEDED(option, new_value)             \
    if (_data_cf_opts.option != cur_data_cf_opts.option) {     \
        new_options[#option] = #new_value;                     \
    }
```

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,97 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+
+    if (!_is_need_update_data_cf_opts)
```
acelyc111, line 3024: What's the aim of this variable?

src/server/pegasus_server_impl.cpp
WHBANG: No, no matter what the last `usage scenario` is, it will change from 'normal' to the corresponding mode, so it will be executed once.
here is a cyclic task, it will get the current `usage scenario`, in order to only execute once when the db is opened, so `is_need_update_data_cf_opts` variable is added to mark whether `recalculate_data_cf_options` has been called

src/server/pegasus_server_impl.cpp
WHBANG: `if (_data_cf_opts.option != cur_data_cf_opts.option) { ` is wrong
use level0_file_num_compaction_trigger as an example: 
1. start_onebox, do not change the config template
```
cur_data_cf_opts.level0_file_num_compaction_trigger = 4
_data_cf_opts.level0_file_num_compaction_trigger = 4
```
2. set `usage scenario`is `bulk load`
```
cur_data_cf_opts.level0_file_num_compaction_trigger = 1000000000
_data_cf_opts.level0_file_num_compaction_trigger = 4
```
3. restart the db, will call LoadLatestOptions() 
```
cur_data_cf_opts.level0_file_num_compaction_trigger = 1000000000
_data_cf_opts.level0_file_num_compaction_trigger = 4
```
At this time `_data_cf_opts.option != cur_data_cf_opts.option`, but in fact the value of option is correct, nothing needs to be done, so need `(value) != cur_data_cf_opts.option) `

src/server/pegasus_server_impl.cpp
WHBANG: Avoid `recalculate_data_cf_options` being called all the time

src/server/pegasus_server_impl.cpp
WHBANG: `_data_cf_opts` involves too many files, not necessary.
i will rename `_tmp_data_cf_opts ` to `_table_data_cf_opts `

src/server/pegasus_server_impl.cpp
acelyc111: I meant value and str_value are always the same, you can use one instead.

WHBANG: when `new_options[#option] = #new_value;`run  the following statement:
```
UPDATE_OPTION_IF_NEEDED(level0_slowdown_writes_trigger,_data_cf_opts.level0_slowdown_writes_trigger)
```
result is:
```
new_options["level0_slowdown_writes_trigger"] = "_data_cf_opts.level0_slowdown_writes_trigger"
```

src/server/pegasus_server_impl.cpp
```diff
@@ -2634,6 +2635,10 @@ void pegasus_server_impl::update_usage_scenario(const std::map<std::string, std:
                            old_usage_scenario,
                            new_usage_scenario);
         }
+    } else {
+        // When an old db is opened and the conf is changed, the options related to usage scenario
```
acelyc111, line 2639: How about change 'the conf is changed' to 'the rocksDB related configs in server config.ini has been changed' ?

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,93 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+            }
+        }
+        UPDATE_NUMBER_OPTION_IF_NEEDED(level0_slowdown_writes_trigger,
+                                       _data_cf_opts.level0_slowdown_writes_trigger);
+        UPDATE_NUMBER_OPTION_IF_NEEDED(level0_stop_writes_trigger,
+                                       _data_cf_opts.level0_stop_writes_trigger);
+        UPDATE_NUMBER_OPTION_IF_NEEDED(soft_pending_compaction_bytes_limit,
+                                       _data_cf_opts.soft_pending_compaction_bytes_limit);
+        UPDATE_NUMBER_OPTION_IF_NEEDED(hard_pending_compaction_bytes_limit,
+                                       _data_cf_opts.hard_pending_compaction_bytes_limit);
```
acelyc111, line 3067: how about define another macro, to leave only one parameter, change "XXX(xxx, _data_cf_opts.xxx)" to "XXX(xxx)"?

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,88 @@ void pegasus_server_impl::reset_usage_scenario_options(
     target_opts->max_write_buffer_number = base_opts.max_write_buffer_number;
 }
 
+void pegasus_server_impl::recalculate_data_cf_options(
+    const rocksdb::ColumnFamilyOptions &cur_data_cf_opts)
+{
+#define UPDATE_NUMBER_OPTION_IF_NEEDED(option, value)                                              \
+    if ((value) != cur_data_cf_opts.option) {                                                      \
+        new_options[#option] = std::to_string((value));                                            \
+    }
```
empiredan, line 3022: better to use `do { ... } while (0)` ?

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,88 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+    if (new_options.size() > 0) {
+        if (set_options(new_options)) {
+            ddebug_replica(
+                "{}: recalculate the value of the options related to usage scenario \"{}\"",
+                replica_name(),
+                _usage_scenario);
+        }
+    }
+    _table_data_cf_opts_recalculated = false;
```
empiredan, line 3095:
```c++
#undef UPDATE_OPTION_IF_NEEDED
#undef UPDATE_BOOL_OPTION_IF_NEEDED
#undef UPDATE_NUMBER_OPTION_IF_NEEDED
```

empiredan (approved):

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,97 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+    do {                                                                                           \
+        if ((value) != cur_data_cf_opts.option) {                                                  \
+            new_options[#option] = std::to_string((value));                                        \
+        }                                                                                          \
+    } while (0)
+
+#define UPDATE_BOOL_OPTION_IF_NEEDED(option, value)                                                \
+    do {                                                                                           \
+        if ((value) != cur_data_cf_opts.option) {                                                  \
+            if ((value))                                                                           \
```
acelyc111, line 3029: it's recommend to keep `{}` for `if..else..` even though there is only one line.
some other places are the same.

src/server/pegasus_server_impl.h
```diff
@@ -413,11 +425,18 @@ class pegasus_server_impl : public pegasus_read_service
     std::shared_ptr<KeyWithTTLCompactionFilterFactory> _key_ttl_compaction_filter_factory;
     std::shared_ptr<rocksdb::Statistics> _statistics;
     rocksdb::DBOptions _db_opts;
+    // The value of option in data_cf according to conf template file config.ini
     rocksdb::ColumnFamilyOptions _data_cf_opts;
+    // Dynamically calculate the value of current data_cf option according to the conf module file
+    // and usage mode
```
acelyc111, line 431:
```suggestion
    // and usage scenario
```

src/server/pegasus_server_impl.h
```diff
@@ -413,11 +425,18 @@ class pegasus_server_impl : public pegasus_read_service
...
+    rocksdb::ColumnFamilyOptions _table_data_cf_opts;
     rocksdb::ColumnFamilyOptions _meta_cf_opts;
     rocksdb::ReadOptions _data_cf_rd_opts;
     std::string _usage_scenario;
     std::string _user_specified_compaction;
+    // Whether it is necessary to update the current data_cf, it is required when opening the db,
+    // but not later
```
acelyc111, line 438:
```suggestion
    // Whether it is necessary to update the current data_cf, it is required when opening the db at the first time,
    // but not later
```

src/server/pegasus_server_impl.h
```diff
@@ -413,11 +425,18 @@ class pegasus_server_impl : public pegasus_read_service
...
+    bool _table_data_cf_opts_recalculated;
```
acelyc111, line 439: the name `_table_data_cf_opts_recalculated` means option has been **recalculated**, it's different with the meaning you defined before, that is to say, true and flase should reverse.

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,98 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+        if (!check_value_if_nearby(_data_cf_opts.write_buffer_size * 4,
+                                   cur_data_cf_opts.write_buffer_size)) {
+            new_options["write_buffer_size"] =
+                std::to_string(get_random_nearby(_data_cf_opts.write_buffer_size * 4));
+        }
+        if (cur_data_cf_opts.max_write_buffer_number !=
+            std::max(_data_cf_opts.max_write_buffer_number, 6)) {
+            new_options["max_write_buffer_number"] =
+                std::to_string(std::max(_data_cf_opts.max_write_buffer_number, 6));
+        }
```
acelyc111, line 3092: Can be replaced by `UPDATE_NUMBER_OPTION_IF_NEEDED(max_write_buffer_number, std::max(_data_cf_opts.max_write_buffer_number, 6))` ?

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,98 @@ void pegasus_server_impl::reset_usage_scenario_options(
     target_opts->max_write_buffer_number = base_opts.max_write_buffer_number;
 }
 
+void pegasus_server_impl::recalculate_data_cf_options(
+    const rocksdb::ColumnFamilyOptions &cur_data_cf_opts)
+{
+#define UPDATE_NUMBER_OPTION_IF_NEEDED(option, value)                                              \
+    do {                                                                                           \
+        if ((value) != cur_data_cf_opts.option) {                                                  \
```
acelyc111, line 3021: better to add a line:
```
auto _v = (value);
```
then use `_v` instead of `value`.

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,98 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+        return;
+    std::unordered_map<std::string, std::string> new_options;
+    if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario ||
+        ROCKSDB_ENV_USAGE_SCENARIO_PREFER_WRITE == _usage_scenario) {
+        if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario) {
+            if (!check_value_if_nearby(_data_cf_opts.write_buffer_size,
+                                       cur_data_cf_opts.write_buffer_size)) {
+                new_options["write_buffer_size"] =
+                    std::to_string(get_random_nearby(_data_cf_opts.write_buffer_size));
+            }
```
acelyc111, line 3049: how about add another mnacro like `UPDATE_OPTION_IF_NOT_NEARBY`?

acelyc111 (approved):

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,96 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+    do {                                                                                           \
+        auto _v = (value);                                                                         \
+        if (_v != cur_data_cf_opts.option) {                                                       \
+            new_options[#option] = std::to_string(_v);                                             \
+        }                                                                                          \
+    } while (0)
+
+#define UPDATE_BOOL_OPTION_IF_NEEDED(option, value)                                                \
+    do {                                                                                           \
+        if ((value) != cur_data_cf_opts.option) {                                                  \
```
empiredan, line 3029: Should `(value)` be assigned to a variable (e.g. `auto _v = (value);`), just like what has been done in `UPDATE_NUMBER_OPTION_IF_NEEDED`, to avoid evaluating 2 times ?

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,96 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+        auto _v = (value);                                                                         \
+        if (!check_value_if_nearby(_v, cur_data_cf_opts.option)) {                                 \
+            new_options[#option] = std::to_string(get_random_nearby(_v));                          \
+        }                                                                                          \
+    } while (0)
+
+#define UPDATE_OPTION_IF_NEEDED(option) UPDATE_NUMBER_OPTION_IF_NEEDED(option, _data_cf_opts.option)
+
+    if (_table_data_cf_opts_recalculated)
+        return;
```
empiredan, line 3049:
```c++
if (_table_data_cf_opts_recalculated) {
    return;
}
```

src/server/pegasus_server_impl.cpp
```diff
@@ -3007,6 +3013,96 @@ void pegasus_server_impl::reset_usage_scenario_options(
...
+    if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario ||
+        ROCKSDB_ENV_USAGE_SCENARIO_PREFER_WRITE == _usage_scenario) {
+        if (ROCKSDB_ENV_USAGE_SCENARIO_NORMAL == _usage_scenario) {
+            UPDATE_OPTION_IF_NOT_NEARBY(write_buffer_size, _data_cf_opts.write_buffer_size);
+            UPDATE_OPTION_IF_NEEDED(level0_file_num_compaction_trigger);
+        } else {
+            uint64_t buffer_size = dsn::rand::next_u64(_data_cf_opts.write_buffer_size,
+                                                       _data_cf_opts.write_buffer_size * 2);
+            if (!(cur_data_cf_opts.write_buffer_size >= _data_cf_opts.write_buffer_size &&
+                  cur_data_cf_opts.write_buffer_size <= _data_cf_opts.write_buffer_size * 2)) {
```
empiredan, line 3060:
```c++ 
if (cur_data_cf_opts.write_buffer_size < _data_cf_opts.write_buffer_size ||
    cur_data_cf_opts.write_buffer_size > _data_cf_opts.write_buffer_size * 2)
```

empiredan (approved):

acelyc111 (approved):

PR #1102, opened 2022-08-04
feat(backup): 2. update and refactor meta backup engine class

hycdong:
https://github.com/apache/incubator-pegasus/issues/1081
As the issue shows, meta_backup_engine will  intertact with replica server and contorl backup lifecycle and status switch. This pr is the first part about it, including:
- update `init_backup` and `start` functions
- update `backup_status` and `backup_item` structures
- update varieties of `meta_backup_engine.h`
- rename `backup_engine` into `meta_backup_engine`
- refactor `meta_backup_engine.h` including public/private, function definition order etc.

Issue #1081, opened 2022-07-27
Feature: enhance cold backup and restore function


hycdong:
# Background
Pegasus currently supports cold backup and restore functions, but both of them have some disadvantages. 

For cold backup, pegasus supports periodic backup through policy. Users can create a policy with backup related parameters such as provider, interval time, and apply this policy to sereval tables. Besides, pegasus also supports onetime backup since release 2.3.0. 
However, backup function has following disadvantages:
- Periodic backup can not start accurately by start time
- When Periodic backup interval time is less than 1 day, periodic backup will be triggered unexpectedly.
- User defined provider path is not supported for periodic backup.
- Once backup is started, it can not be canceled. When backup failed, it will continue to retry until succeed, even restart meta server.
- Current backup will cost heavy I/O during creating checkpoint.
- The path on provider is hard to find one table's backup.
- Backup code is not firendly to read and maintain.

For restore, pegasus supports two data_version. Tables created in release 1.x is V0, and tables created in release 2.x is V1. Restore process will create an empty table, then apply the backup checkpoint. There will be a compatible problem that release 2.x table can not apply V0 checkpoint, which will lead to coredump making cluster useless. As a result, restore need to check table data_version to make it robust.

# New backup design
The enhance version of backup, simplify backup v2, will solve all probelms above, providing a simple backup function.

## Components
Meta backup function is consist of three parts:
- Backup engine - intertact with replica server
- Periodic backup context - manage table periodic backup policy and backups
  - meta server will have a timer to check whether periodic backup should be triggered
  - for first triggered backup, server will check it by start_time whose format is like "15:00"
  - for not-first backup, server  will compare last backup start time and periodic backup interval
  - periodic backup is not allowed to be modified, but can be deleted and recreated
- Backup service - manage cluster all tables backup, including onetime backup and periodic backup. Besides, it also expose the rpc interface to admin-cli and shell
  - add table periodic backup policy
  - query periodic backup policy
  - disable/enable periodic backup policy
  - delete periodic backup policy
  - start onetime backup
  - query backup (onetime and periodic)
  - cancel backup (onetime and periodic)
 
## Main flow
![image](https://user-images.githubusercontent.com/17868458/181185836-6eea1905-eb29-4557-bb1b-421c2b12d3a4.png)
- when receving start backup, engine will turn its backup status into `checkpointing` and send request to replica servers
- replica will turn its state into `checkpointing`, and turn to `checkpointed` after generating checkpoint succeed
[...]

PR #1112, opened 2022-08-12
feat(backup): 4. meta server send backup_request to replica

hycdong:
https://github.com/apache/incubator-pegasus/issues/1081

This pull request implement that meta server send backup_request to replica server:
1. meta server create `backup_engine` instance by `init_backup`, already in [#1102](https://github.com/apache/incubator-pegasus/pull/1102)
2. write `app_metadata` file into remote file system by calling `write_app_info`
3. update app backup_status from `UNINITIALIZED` into `CHECKPOINTING` by calling `update_backup_item_on_remote_storage`
4. send backup_request to all primary partition by calling `backup_app_partition`
5. handle response and resend request by calling `on_backup_reply`, will be implemented in further pr

Besides, this pr also update following related content:
- update `backup_request` and `backup_response` in thrift
- implement some common functions in `backup_restore_common`
- rename `cold_backup_constant` into `backup_constant`
- update `cold_backup_root` config into gflags
- add related unit test


src/rdsn/src/common/backup.thrift
```diff
@@ -64,24 +64,28 @@ struct configuration_restore_request
...
     1:dsn.error_code    err;
     2:dsn.gpid          pid;
-    3:i32               progress;  // the progress of the cold_backup
-    4:string            policy_name;
-    5:i64               backup_id;
-    6:i64               checkpoint_total_size;
+    3:i64               backup_id;
+    4:backup_status     status;
+    5:optional dsn.error_code   checkpoint_err;
+    6:optional dsn.error_code   upload_err;
```
foreverneverer, line 86: why need two error_code? using one-error_code + hint is not ok?

src/rdsn/src/common/backup_restore_common.cpp
```diff
@@ -36,5 +44,41 @@ const std::string backup_restore_constant::BACKUP_ID("restore.backup_id");
 const std::string backup_restore_constant::SKIP_BAD_PARTITION("restore.skip_bad_partition");
 const std::string backup_restore_constant::RESTORE_PATH("restore.restore_path");
 
+std::string get_backup_root(const std::string &backup_root,
+                            const std::string &user_defined_root_path)
+{
+    if (user_defined_root_path.empty()) {
+        return backup_root;
+    }
+    return utils::filesystem::path_combine(user_defined_root_path, backup_root);
```
foreverneverer, line 53: if user specify root path, we should return the value, but not combine `system` + `user` path

src/rdsn/src/common/backup.thrift
```diff
@@ -64,24 +64,28 @@ struct configuration_restore_request
     9:optional string   restore_path;
 }
 
+// meta -> replica
 struct backup_request
```
acelyc111, line 68: Will you use another RPC code? Is there any problem if a user use old version shell-tool  attempt to control the cluster?

src/rdsn/src/common/backup_restore_common.cpp
hycdong: If user specify root_path, we will return `root_path/backup_root`, otherwise return backup_root, backup_root is actually the cluster name.

src/rdsn/src/common/backup.thrift
hycdong: This rpc is not used from shell-tool to meta server, is meta server to replica server, won't trigger the control problem.

src/rdsn/src/common/backup.thrift
hycdong: Good idea, meta server can distinguish error with its backup_status.

foreverneverer: Ok, what the different between  `1:dsn.error_code    err` and `5:optional dsn.error_code   checkpoint_upload_err;`?

hycdong: `err` is used for backup_request rpc, such as ERR_OK, ERR_INVALID_STATE. `checkpoint_upload_err` is used for backup checkpoint  and upload. If backup checkpoint failed, checkpoint_upload_err won't be error_ok.

foreverneverer: So I think that just `err` is still OK. In other rpc, I remember that we have always defined it using one `err`.

hycdong: No, error is not always OK. For example, current partition is not primary, backup should be failed, err is ERR_INVALID_STATE.

foreverneverer: It's doesn't matter. If all passed, you return `ERR_OK`, otherwise, return `ERR_XXX`, you don't need return the case : `err = ERR_OK`, but `checkpoint=ERR_CHECKPOINT_FAILED`, just return `ERR_CHECKPOINT_FAILED` is ok.

foreverneverer: > No, error is not always OK

I mean that just need one variable `err` is enough

hycdong: I got your point. You think checkpoint_upload_err is unecessary, all errors can be shown by err. I think it is not clear if we combine two different error into one. `checkpoint_upload_err` is only used to show if checkpoint or upload meet error during backup. `err` is used for common error for backup request.

foreverneverer: If your error definition is appropriate enough, for example, I don't think a single err `ERR_ CHEKCPOINT_ FAILED` has different from  two err `err = ERR_ OK but err_checnpoint=ERR_ CHEKCPOINT_ FAILED`.

foreverneverer: In addition, hint_ message can also explain the error code, so there is no need to distinguish `common` and `backup`

hycdong: I don't agree with that one error code should identify all error cases. `checkpoint_upload_err` is only valid in backup checkpoint and upload stage, other stage it should be set. `err=ok` and `checkpoint_upload_err=ok` means backup not meet error during checkpoint and upload. It is clear to have two errors.
Besides, this rpc is sent from meta server and replica server periodically during backup, meta server will do different actions according to the err and checkpoint_upload_err, not only show the hint_message.
I still inisist on my previous design, seperate another error code.

foreverneverer: > meta server will do different actions

One err_code can't meet this requirement?

hycdong: In fact, only one error code can meet the requirement, but as I mentioned before I thought it was not clear enough, they are different error, should not be combined.

foreverneverer: I still think `checkpoint_upload_err` is redundant. if `ERR_CHECKPOINT_FAILED` can be clearly indicates that an error occurred in the checkpoint stage , why define `check_point_err` and set it value is `ERR_CHECKPOINT_FAILED` to explain?

In the past rpc, we haven't this design. Include bulkload, duplication, for example, we don't need add `ingest_err` to show the error of ingesting, or add `download_err` to show the err of downloading and so on, which is redundant, bu just use `ERR_INGEST_FAILD`, `ERR_DOWNLOAD` is clear enough.

foreverneverer: With your design, the sender needs to judge the response as follows:
```c++
if (rpc == ok) {
     if ( resp.err == ok) {
         if (resp.checkpoint_err == ok) {
             /** if you define `more clear err code as you say`, maybe need:**/
              /** if (err_a == ...) {
                        if (err_b == ...)
}
}
}
}
```
This logic is also redundant. don't suggest this design, you can refer the origin rpc define, I think the origin is elegant

hycdong: Actually, bulk load still seperete rpc_error, download_error and ingestion_err, you can reference sturcture `partition_bulk_load_state`.

And in my current design, won't have the too many if condition in your example. In my design, it just like
```
if (rpc != ok) { // handle this condition }
if (resp.err != ok) { // handle this condition }
if (resp.checkpoint_upload_err != ok) { // handle this condition }
```
If I combine error and checkpoint_upload_err into one, the code will be like:
```
if (rpc != ok) { // handle this condition }
if (resp.err != ok) { 
    if (original resp.err != ok){
        // handle this condition
    }
    if (original resp.checkpoint_upload_err != ok ) {
        // handle this condition
    }
}
```

I think it is okay to have two errors to distinguish different errors, if there is only one field redundant, but can make structure clear. In my previous design, checkpoint error and upload error also should be seperated, I have already compromised my logic :-)

foreverneverer: Well, wait other's opinion, if they have no have no objections, I will be willing think it is a reasonable design

src/rdsn/src/common/backup.thrift
acelyc111: OK, thanks.
I meant how to keep compatablity, if the meta servers are in new version, and the replica server are in old version, what will happen if we ask the cluster to do backup? Is it neccessary to add new rpc code for the new implemention.

src/rdsn/src/common/backup_restore_common.h
```diff
@@ -52,5 +57,69 @@ class backup_restore_constant
...
+
+// TODO(heyuchen): add other common functions
+// get_compatible_backup_root is only used for restore compatible backup
+
+// The backup root path on block service
+// if user_defined_root_path is not empty
+// - return <user_defined_root_path>/<backup_root>
+// else
+// - return <backup_root>
+std::string get_backup_root(const std::string &backup_root,
```
acelyc111, line 97: where will this function been called? if only called in src/rdsn/src/meta/meta_backup_engine.cpp, how about define it as a local funciton there?

src/rdsn/src/common/backup_restore_common.cpp
```diff
@@ -36,5 +44,41 @@ const std::string backup_restore_constant::BACKUP_ID("restore.backup_id");
...
+                            const bool is_compatible)
+{
+    std::string str_app = app_name + "_" + std::to_string(app_id);
+    std::stringstream ss;
+    if (!is_compatible) {
+        ss << root << "/" << str_app << "/" << backup_id;
+    } else {
+        ss << root << "/" << backup_id << "/" << str_app;
+    }
+    return ss.str();
```
acelyc111, line 69: how about use fmt::format instead?

src/rdsn/src/common/backup_restore_common.cpp
```diff
@@ -36,5 +44,41 @@ const std::string backup_restore_constant::BACKUP_ID("restore.backup_id");
...
+}
+
+std::string get_backup_meta_path(const std::string &root,
+                                 const std::string &app_name,
+                                 const int32_t app_id,
+                                 const int64_t backup_id,
+                                 const bool is_compatible)
+{
+    std::stringstream ss;
+    ss << get_backup_path(root, app_name, app_id, backup_id, is_compatible) << "/meta";
```
acelyc111, line 79: how about use fmt::format instead?

src/rdsn/src/common/backup_restore_common.h
```diff
@@ -52,5 +57,69 @@ class backup_restore_constant
...
+                            const std::string &user_defined_root_path);
+
+// This backup path on block service
+// if is_compatible = false (root is the return value of get_backup_root function)
+// - return <root>/<app_name>_<app_id>/<backup_id>
+// else (only used for restore compatible backup, root is the return value of
+// get_compatible_backup_root function)
+// - return <root>/<backup_id>/<app_name>_<app_id>
+std::string get_backup_path(const std::string &root,
```
acelyc111, line 106: if it is only used in backup_restore_common.cpp, how about define it as a local anonymous function there?

src/rdsn/src/meta/meta_backup_engine.cpp
```diff
@@ -86,87 +86,57 @@ void meta_backup_engine::start()
...
+        app_info_buffer = json::json_forwarder<app_info>::encode(tmp);
     }
 
-    std::string backup_root =
-        dsn::utils::filesystem::path_combine(_backup_path, _backup_service->backup_root());
-    // TODO(heyuchen): refactor and update it in future
-    std::string file_name = "todo";
-    return write_backup_file(file_name, app_info_buffer);
+    std::string remote_meta_dir =
+        get_backup_meta_path(get_backup_root(FLAGS_cold_backup_root, _cur_backup.backup_path),
```
acelyc111, line 103: can we move `get_backup_root` into `get_backup_meta_path`, and only left `_cur_backup.backup_path` here?

src/rdsn/src/common/backup.thrift
hycdong: If meta server is in new version, replica server in old version. I don't consider this condition compatible, I think it is a dangerous case for meta server and replica server has different version, because new version especially a feature version, meta server will provide mamy new rpc which replica server can not recognize, and it is not necessary to add new rpc code for compatible only when new meta and old replica case.

src/rdsn/src/common/backup.thrift
hycdong: @acelyc111 Please give us your suggestion about this comment~ @foreverneverer has different opion with me, and we can not persude each other.

src/rdsn/src/meta/meta_backup_engine.h
```diff
@@ -40,6 +40,19 @@ struct app_backup_info
...
+/// backup path on remote storage
+///
+/// Onetime backup:
+/// <cluster_root>/backup/<app_id>/once/<backup_id>/<backup_item>
+///
+/// Periodic backup:
+/// <cluster_root>/backup/<app_id>/periodic/<periodic_backup_policy>
+/// <cluster_root>/backup/<app_id>/periodic/<backup_id>/<backup_item>
+///
+static const std::string ONETIME_PATH = "once";
```
acelyc111, line 53: Will these variables be used in some other files? if not, better to hide in cpp.

src/rdsn/src/meta/meta_backup_engine.h
```diff
@@ -92,12 +111,28 @@ class meta_backup_engine
...
+        return meta_options::concat_path_unix_style(_meta_svc->get_cluster_root(), "backup");
+    }
+
+    std::string get_remote_backup_path() const
+    {
+        auto type_path = _is_periodic_backup ? PERIODIC_PATH : ONETIME_PATH;
+        std::stringstream ss;
+        ss << get_remote_storage_root() << "/" << _cur_backup.app_id << "/" << type_path << "/"
+           << _cur_backup.backup_id;
+        return ss.str();
```
acelyc111, line 133: how about use `fmt::format`?

src/rdsn/src/meta/test/meta_backup_engine_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "meta_test_base.h"
+#include "meta/meta_backup_engine.h"
+
+namespace dsn {
+namespace replication {
+
+class meta_backup_engine_test : public meta_test_base
+{
+public:
+    void SetUp()
```
acelyc111, line 31: better to add `override`

src/rdsn/src/meta/test/meta_backup_engine_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {
+        meta_test_base::SetUp();
+        create_app(APP_NAME, PARTITION_COUNT);
+        std::shared_ptr<app_state> app = find_app(APP_NAME);
+        _app_id = app->app_id;
+        _backup_engine = std::make_shared<meta_backup_engine>(_ms.get(), false);
+        fail::setup();
+    }
+
+    void TearDown()
```
acelyc111, line 41: same

src/rdsn/src/common/backup.thrift
acelyc111: Rolling update is a common case which many user have to face. We will not ensure every feature should work well at this case, but at least avoid the server crash.
If we rewrite the RPC message, the related RPC code would better to add a new one, and left the old one as deprecated.

src/rdsn/src/common/backup.thrift
acelyc111: If I didn't miss anything, I didn't find where `checkpoint_upload_err` is used currently. We keep this point in mind and see how will it be used later, if found it's not so elegant, we can point it out then.

src/rdsn/src/common/backup_restore_common.h
hycdong: There are many logic will call this function, including replica and meta.

src/rdsn/src/common/backup.thrift
hycdong: Rolling update is a common case, but it is not recommended to update meta firstly, we recommendly update replica server firstly, so it is a seldom case that meta is new version but replica is old version.

src/rdsn/src/common/backup_restore_common.h
hycdong: All functions in `backup_restore_common.h` will be used in meta and replica files, backup and restore logic, so I move them into `backup_restore_common.h`. This file is added in my last round refactor.

src/rdsn/src/meta/meta_backup_engine.h
hycdong: Yes, it will be used in other cpp files.

src/rdsn/src/meta/test/meta_backup_engine_test.cpp
hycdong: Done

src/rdsn/src/common/backup_restore_common.cpp
hycdong: Done, and I resolve the following similar conversation.

src/rdsn/src/common/backup.thrift
acelyc111: No matter which one is new version. The new replica server will still crash if receive a old backup request from old versio meta server?

acelyc111 (approved):
LGTM
But still doubt if there is any compatablity problem.

foreverneverer (approved):

hycdong:
> LGTM But still doubt if there is any compatablity problem.

Okay, I will add compatiable explaination after all code merged.

Issue #1122, opened 2022-08-16
Fix the conditions for file changes that trigger workflows for github


empiredan:
While only doc files such as `LICENSE` or `NOTICE` are changed and committed, cpp workflows for github will report error as follows (see https://github.com/apache/incubator-pegasus/runs/7854426641?check_suite_focus=true):
```
Run rm -rf thirdparty
  rm -rf thirdparty
  tar -zcvhf release__builder.tar DSN_ROOT/ src/builder/bin src/builder/src/server/test/config.ini  --exclude='*CMakeFiles*'
  shell: sh -e {0}
tar: DSN_ROOT: Cannot stat: No such file or directory
tar: src/builder/bin: Cannot stat: No such file or directory
tar: src/builder/src/server/test/config.ini: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
Error: Process completed with exit code [2](https://github.com/apache/incubator-pegasus/runs/7854426137?check_suite_focus=true#step:11:2).
```

The binaries are not built since there is not any source file changed. However, since the binaries are used to be build packages which will be uploaded as artifacts, it will report "No such file or directory".

PR #1123, opened 2022-08-16
fix(github): fix the conditions for file changes that trigger workflows

empiredan:
This PR is to solve https://github.com/apache/incubator-pegasus/issues/1122.

To filter all the file changes that should trigger actions for github, at beginning of CI all of the path patterns are listed, which means workflows for build and test will never be run for the files that are irrelevant.


acelyc111 (approved):

acelyc111:
@foreverneverer This patch need to be cherry-picked to branch 2.4

empiredan:
> @foreverneverer This patch need to be cherry-picked to branch 2.4

OK.

foreverneverer (approved):

PR #1129, opened 2022-08-19
feat(backup): 6. meta handle backup_response during checkpointing

hycdong:
https://github.com/apache/incubator-pegasus/issues/1081
This pr is about the meta server receive `backup_response` with checkpoint status:
- if one partition failed during checkpoint, whole backup process will be failed
- if all partitions finish checkpoint, turn app backup_status into uploading
- resend backup_request to replica if backup is not finished
Besides, this pr adds related unit tests.

### DIFF
--- a/src/rdsn/src/meta/meta_backup_engine.cpp
+++ b/src/rdsn/src/meta/meta_backup_engine.cpp
@@ -191,6 +191,8 @@ void meta_backup_engine::on_backup_reply(const error_code err,
             derror_f("partition[{}] handle backup failed", pid);
             return;
         }
+
+        // TODO(heyuchen): check if backup canceled
     }
 
     auto rep_error = err == ERR_OK ? response.err : err;
--- a/src/rdsn/src/meta/meta_backup_engine.cpp
+++ b/src/rdsn/src/meta/meta_backup_engine.cpp
@@ -180,6 +180,84 @@ void meta_backup_engine::backup_app_partition(const gpid &pid)
 }
 
 // ThreadPool: THREAD_POOL_DEFAULT
+void meta_backup_engine::on_backup_reply(const error_code err,
+                                         const backup_response &response,
+                                         const gpid &pid,
+                                         const rpc_address &primary)
+{
+    {
+        zauto_read_lock l(_lock);
+        if (_is_backup_failed) {
+            derror_f("partition[{}] handle backup failed", pid);
+            return;
+        }
+    }
+
+    auto rep_error = err == ERR_OK ? response.err : err;
+    if (rep_error != ERR_OK) {
+        derror_f(
+            "backup_id({}): receive backup response for partition {} from server {}, error = {}",
+            _cur_backup.backup_id,
+            pid.to_string(),
+            primary.to_string(),
+            rep_error);
+        handle_replica_backup_failed(pid.get_app_id());
+        return;
+    }
+
+    if (response.backup_id != _cur_backup.backup_id) {
+        dwarn_f("backup_id({}): receive outdated backup response(backup_id={}) for partition {} "
+                "from server {}, ignore it",
+                _cur_backup.backup_id,
+                response.backup_id,
+                pid.to_string(),
+                primary.to_string());
+        retry_backup(pid);
+        return;
+    }
+
+    if (response.__isset.checkpoint_upload_err) {
+        auto type = response.status == backup_status::UPLOADING ? "upload" : "checkpoint";
+        derror_f("backup_id({}): receive backup response for partition {} from server {}, meet {} "
+                 "error = {}",
+                 _cur_backup.backup_id,
+                 pid.to_string(),
+                 primary.to_string(),
+                 type,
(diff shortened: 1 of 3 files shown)


src/rdsn/src/meta/meta_backup_engine.cpp
```diff
@@ -179,6 +179,86 @@ void meta_backup_engine::backup_app_partition(const gpid &pid)
...
 
+// ThreadPool: THREAD_POOL_DEFAULT
+void meta_backup_engine::on_backup_reply(const error_code err,
+                                         const backup_response &response,
+                                         const gpid &pid,
+                                         const rpc_address &primary)
+{
+    {
+        zauto_read_lock l(_lock);
+        if (_is_backup_failed) {
```
foreverneverer, line 190: Is there a status of BACKUP_FAILED?  Why do we still need a flag variable?

src/rdsn/src/meta/meta_backup_engine.cpp
```diff
@@ -179,6 +179,86 @@ void meta_backup_engine::backup_app_partition(const gpid &pid)
...
+                "from server {}, ignore it",
+                _cur_backup.backup_id,
+                response.backup_id,
+                pid.to_string(),
+                primary.to_string());
+        retry_backup(pid);
+        return;
+    }
+
+    if (response.__isset.checkpoint_upload_err) {
```
foreverneverer, line 221: is `response.__isset.checkpoint_upload_err` always  `true`

src/rdsn/src/meta/meta_backup_engine.cpp
```diff
@@ -179,6 +179,86 @@ void meta_backup_engine::backup_app_partition(const gpid &pid)
...
+        auto type = response.status == backup_status::UPLOADING ? "upload" : "checkpoint";
```
foreverneverer, line 222: `enum` seem can be converted to `string`?

src/rdsn/src/meta/meta_backup_engine.cpp
```diff
@@ -179,6 +179,86 @@ void meta_backup_engine::backup_app_partition(const gpid &pid)
...
+    if (response.status == backup_status::CHECKPOINTED) {
+        ddebug_f("backup_id({}): backup for partition {} from server {} finish checkpoint",
+                 _cur_backup.backup_id,
+                 pid.to_string(),
+                 primary.to_string());
+        {
+            zauto_write_lock l(_lock);
+            _backup_status[pid.get_partition_index()] = backup_status::UPLOADING;
+        }
+        if (check_partition_backup_status(backup_status::UPLOADING)) {
```
foreverneverer, line 244: so just `uploading` status storage zk?

src/rdsn/src/meta/meta_backup_engine.h
```diff
@@ -119,6 +124,17 @@ class meta_backup_engine
 
     void update_backup_item_on_remote_storage(backup_status::type new_status, int64_t end_time = 0);
 
+    bool check_partition_backup_status(backup_status::type expected_status) const
+    {
```
foreverneverer, line 128: consider use `std::all_of`?

src/rdsn/src/meta/meta_backup_engine.cpp
hycdong: `on_backup_reply` only handle one partition backup, `_is_backup_failed` will set true when one partition backup failed, all partitions needs to be failed, you can reference function `handle_replica_backup_failed`.

src/rdsn/src/meta/meta_backup_engine.cpp
hycdong: No.

src/rdsn/src/meta/meta_backup_engine.cpp
hycdong: Yes, enum_to_string(status) can make status into string, but it is too long, such as replication::backup_status::uploading

src/rdsn/src/meta/meta_backup_engine.cpp
hycdong: I don't get your question meaning.
When one partition checkpoint succeed, its backup_status will be uploading, if all partitions' status are uploading, app backup status will be uploading on zk and meta locally. `update_backup_item_on_remote_storage` has alreay implemented in previous pr.

foreverneverer: I mean why only `backup_status::UPLOADING` allow to be sync to zk?

hycdong: Not only `uploading` should be sync to zk, other state also need, just not in this pr. 
In previous pr, when backup start, its status is checkpoint, which also need to be sync to zk, you can reference function `meta_backup_engine::start`. 
In future pr, when backup failed or succeed, its status also need to be sync to zk.

foreverneverer (approved):

empiredan (approved):

PR #1096, opened 2022-08-02
refactor(java-client): remove duplicate variable

acelyc111:
https://github.com/apache/incubator-pegasus/issues/887

No functional changes.

PR #1130, opened 2022-08-19
refactor(java-client): optimize the code style of PegasusScanner

totalo:
For https://github.com/apache/incubator-pegasus/issues/887

No function change


java-client/src/main/java/org/apache/pegasus/client/ClientOptions.java
```diff
@@ -233,7 +233,6 @@ public int hashCode() {
     result = 31 * result + Objects.hashCode(falconPushInterval.toMillis());
     result = 31 * result + (enableWriteLimit ? 1 : 0);
     result = 31 * result + Objects.hashCode(metaQueryTimeout.toMillis());
-    result = 31 * result + authProtocol.hashCode();
```
foreverneverer, line 236: `authProtocol` is not existed, I'm confused why ci can pass in https://github.com/apache/incubator-pegasus/pull/1087.

do we update action for java client? @acelyc111

totalo: I think it's because #1096 was removed, but I didn't merge master, it used to run d

I think it's because #1096 was removed, but I didn't merge the master, I ran the previous complete code so that my pr could pass, and then the master didn't run ci again after the pr was merged, so I didn't find this problem.

Maybe we need to run ci after merging to master.

foreverneverer: I remember github will show message and suggest merge master latest code if pr behind master, but recent, this feature seem to be closed? @acelyc111 

Keep latest code base master before merging is necessary, I think we should revert the feature as above. @acelyc111

acelyc111: @foreverneverer Rebase the master branch for every PR before merging into the master branch is too costly, and some unit tests are flacky, it will last a very long time.

I agree to @totalo , we can add some regular CI on master branch for Java, Go, and etc, code, as how cpp code is doing now.

foreverneverer: Ok, it's also a resolution

acelyc111 (approved):
Good work! Sorry to late reply.

empiredan (approved):

Issue #1131, opened 2022-08-22
Limit udp service by config


GehaFearless:
## Feature Request

**Describe the feature you'd like:**
Config `rpc_call_channel` select what channel this RPC item used. But the UDP service always start on server whether RPC use UDP or not. 

**Describe alternatives you've considered:**
Add a configuration 'enable_udp' with a default value of 'false' in the 'network' section. It gives the server control over whether to start the UDP service.

If one of RPC config `rpc_call_channel` is `RPC_CHANNEL_UDP`, should make 'enable_udp' is `true` first.

PR #1132, opened 2022-08-22
feat(server): disable UDP service by config

GehaFearless:
issue: https://github.com/apache/incubator-pegasus/issues/1131

Disable UDP service when not use it.

Add a configuration 'enable_udp' with a default value of 'false' in the 'network' section. It gives the server control over whether to start the UDP service.

### DIFF
--- a/src/runtime/nativerun.cpp
+++ b/src/runtime/nativerun.cpp
@@ -34,9 +34,12 @@
  */
 
 #include "runtime/nativerun.h"
+#include "utils/flags.h"
 
 namespace dsn {
 namespace tools {
+
+DSN_DECLARE_bool(enable_udp);
 
 void nativerun::install(service_spec &spec)
 {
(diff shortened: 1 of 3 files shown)


src/server/config.ini
```diff
@@ -94,6 +94,7 @@
   io_service_worker_count = 4
   ; how many connections can be established from one ip address to a server(both replica and meta), 0 means no threshold
   conn_threshold_per_ip = 0
+  enable_udp = false
```
acelyc111, line 97: UDP is enabled before, the default value would better to be `true` to keep compatiblity, you can define it what ever you like in your own envs.

src/server/config.ini
```diff
@@ -708,12 +709,10 @@
 
 [task.RPC_FD_FAILURE_DETECTOR_PING]
   rpc_call_header_format = NET_HDR_DSN
-  rpc_call_channel = RPC_CHANNEL_UDP
```
acelyc111, line 711: same, not remove it.

src/rdsn/src/runtime/providers.common.cpp
```diff
@@ -64,8 +64,11 @@ void register_common_providers()
 
     register_std_lock_providers();
 
+    if (dsn_config_get_value_bool(
```
acelyc111, line 67: use the new config framework `DSN_DEFINE_bool(...)`

src/rdsn/src/runtime/nativerun.cpp
```diff
@@ -59,19 +60,22 @@ void nativerun::install(service_spec &spec)
...
-        spec.network_default_client_cfs[RPC_CHANNEL_UDP] = cs;
-    }
-    {
-        network_server_config cs2;
-        cs2.port = 0;
-        cs2.channel = RPC_CHANNEL_UDP;
-        cs2.factory_name = "dsn::tools::asio_udp_provider";
-        cs2.message_buffer_block_size = 1024 * 64;
-        spec.network_default_server_cfs[cs2] = cs2;
+    if (dsn_config_get_value_bool(
```
acelyc111, line 63: use the recommend new style DSN_DECLARE_bool, FLAGS_enable_udp

src/rdsn/src/runtime/nativerun.cpp
```diff
@@ -59,19 +62,21 @@ void nativerun::install(service_spec &spec)
...
-        spec.network_default_server_cfs[cs2] = cs2;
+    if (FLAGS_enable_udp) {
+        {
+            network_client_config cs;
+            cs.factory_name = "dsn::tools::asio_udp_provider";
+            cs.message_buffer_block_size = 1024 * 64;
+            spec.network_default_client_cfs[RPC_CHANNEL_UDP] = cs;
+        }
+        {
+            network_server_config cs2;
```
acelyc111, line 73: Now that they are in different scope, it's not needed to use different name, `cs2` is not a very good name.

src/rdsn/src/runtime/task/task_spec.cpp
```diff
@@ -232,6 +236,12 @@ bool task_spec::init()
                 return false;
             }
         }
+
+        if (spec->rpc_call_channel == RPC_CHANNEL_UDP && !dsn::tools::FLAGS_enable_udp) {
+            derror("task rpc_call_channel RPC_CHANNEL_UCP need udp service, set network "
+                   "configuration enable_udp true");
```
acelyc111, line 242: Better to clarify what config it is, that is `[network].enable_udp`

acelyc111 (approved):

empiredan (approved):

PR #1137, opened 2022-08-23
[client-java] Use properties for depend

liangyuanpeng:
Use properties for depend version. Have not code change.

##### Tests 

- No code


acelyc111 (approved):
Thanks for your contribution!

Apache9 (approved):

Issue #1157, opened 2022-09-09
Wrong logging format was found in meta server_state


empiredan:
Some logs are found for meta server as follows:
```
F2022-09-09 17:02:17.184 (1662714137184230445 11231)   meta.meta_state0.0202000700000001: server_state.cpp:895:on_config_sync(): assertion expression: false
F2022-09-09 17:02:17.184 (1662714137184247617 11231)   meta.meta_state0.0202000700000001: server_state.cpp:895:on_config_sync(): gpid({}) on node({}) is not exist on meta server, administrator should check consistency of meta data
```

The key point is `gpid({}) on node({})` in logs of meta server rather than the fatal level, which is the result of connecting to a wrong replica server. `gpid({}) on node({})` must be caused by invalid logging format.

empiredan:
This issue has been fixed by https://github.com/apache/incubator-pegasus/pull/1217.

PR #1158, opened 2022-09-09
fix: fix invalid logging format for meta server_state

empiredan:
This PR is to fix https://github.com/apache/incubator-pegasus/issues/1157.

This issue is caused by invalid logging format as follows:
```C++
                        dassert(false,
                                "gpid({}) on node({}) is not exist on meta server, administrator "
                                "should check consistency of meta data",
                                rep.pid,
                                request.node.to_string());
```

For `{}` in logging, `dassert_f` should be used instead of `dassert`. This PR has also replaced all logging format with `*_f` for meta server_state, including:

- dassert -> dassert_f
- derror -> derror_f
- dwarn -> dwarn_f
- ddebug -> ddebug_f
- dinfo -> dinfo_f


empiredan:
This issue has been fixed by https://github.com/apache/incubator-pegasus/pull/1217.

Issue #1166, opened 2022-09-19
Pegasus code style guide


acelyc111:
As the Pegasus project developing, more and more developer joined into the project, and there are also many code legacy from years ago, I propose to design a code style for Pegasus project.

- [ ] enable unused-result warning

PR #1169, opened 2022-09-22
refactor(function_test): use test_util to reduce duplicate code

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1166

This is a flow up to https://github.com/apache/incubator-pegasus/pull/1149, now refactoring these function tests to inherit from test_util to reduce duplicate code.
And also include some other non-functional refactors.

### DIFF
--- a/src/test/function_test/utils/utils.h
+++ b/src/test/function_test/utils/utils.h
@@ -19,15 +19,16 @@
 
 #pragma once
 
-#include <dsn/utility/rand.h>
 #include <dsn/c/api_utilities.h>
 #include <dsn/dist/fmt_logging.h>
+#include <dsn/utility/rand.h>
+#include <dsn/utility/strings.h>
 
 #define RETRY_OPERATION(CLIENT_FUNCTION, RESULT)                                                   \
     do {                                                                                           \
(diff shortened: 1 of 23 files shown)


src/test/function_test/utils/utils.h
```diff
@@ -201,3 +202,15 @@ inline void compare(const T &expect, const U &actual)
         ASSERT_NO_FATAL_FAILURE(compare(it1->second, it2->second, it1->first));
     }
 }
+
+inline void run_cmd(const std::string &cmd, std::string *output = nullptr)
+{
+    std::stringstream ss;
+    int ret = dsn::utils::pipe_execute(cmd.c_str(), ss);
```
levy5307, line 209: In my opinion, it's better to reduce the usage of raw pointer. Using std::string instead is a better choice.

acelyc111: Agree, we can refactor functions like `pipe_execute ` in another patch, this patch only refactor function tests.

levy5307 (approved):

foreverneverer (approved):

Issue #1183, opened 2022-10-12
update app_name on 'on_config_sync '


GehaFearless:
## Bug Report

When I execute `recall app_id new_app_name`, the app_name on zookeeper changed but file '.app_info' for replica not. 
And If I use this 'app_info' rebuild metadata, I will get one outdated app_name.

So I want to add `update_app_name` on `on_config_sync`.

GehaFearless:
I will solve by this issue: https://github.com/apache/incubator-pegasus/issues/1185

PR #1184, opened 2022-10-12
fix: update app_name on 'on_config_sync'

GehaFearless:

https://github.com/apache/incubator-pegasus/issues/1183

 I want to add `update_app_name` on `on_config_sync`. 

`app_name` on data_dirs file `.app_info` will be changed when execute `recall`.

### DIFF
--- a/src/rdsn/src/replica/replica.h
+++ b/src/rdsn/src/replica/replica.h
@@ -496,6 +496,7 @@ private:
     clear_on_failure(replica_stub *stub, replica *rep, const std::string &path, const gpid &pid);
 
     void update_app_max_replica_count(int32_t max_replica_count);
+    void update_app_name(std::string app_name);
 
 private:
     friend class ::dsn::replication::test::test_checker;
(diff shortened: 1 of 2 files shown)


src/rdsn/src/replica/replica.h
```diff
@@ -496,6 +496,7 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
     clear_on_failure(replica_stub *stub, replica *rep, const std::string &path, const gpid &pid);
 
     void update_app_max_replica_count(int32_t max_replica_count);
+    void update_app_name(std::string app_name);
```
acelyc111, line 499: better to use const reference

src/rdsn/src/replica/replica_config.cpp
```diff
@@ -1143,6 +1144,23 @@ void replica::on_config_sync(const app_info &info,
...
+{
+    if (app_name == _app_info.app_name) {
+        return;
+    }
+
+    auto old_app_name = _app_info.app_name;
+    _app_info.app_name = app_name;
+
+    auto ec = store_app_info(_app_info);
+    dassert_replica(ec == ERR_OK,
```
acelyc111, line 1157: use `dcheck_eq_replica` to simplify the code

Issue #1177, opened 2022-09-28
Download apache-zookeeper-3.7.0-bin.tar.gz during release building or third-parties image building for unit tests


empiredan:
Since https://github.com/apache/incubator-pegasus/issues/1164 had been resolved, apache-zookeeper-3.7.0-bin.tar.gz has failed to be downloaded very frequently. For example, for only one run [3140473266](https://github.com/apache/incubator-pegasus/actions/runs/3140473266/) of CI there are 3 jobs that are failed due to apache-zookeeper-3.7.0-bin.tar.gz is failed to be downloaded:

![image](https://user-images.githubusercontent.com/743379/192726370-e344a5c3-063c-47a6-b59a-c71aadbd8561.png)

![image](https://user-images.githubusercontent.com/743379/192725053-3faf99fb-8fd4-493a-9e58-87a47c1766cb.png)

![image](https://user-images.githubusercontent.com/743379/192726161-b6136a6f-57ec-4d24-9d69-ec38506b1ee0.png)

The solution to this issue is also same with https://github.com/apache/incubator-pegasus/issues/1164: 2 pull requests will be created for this issue:

PR #1186, opened 2022-10-13
ci: download apache-zookeeper-3.7.0-bin.tar.gz during release building for unit tests once third-parties have change

empiredan:
This PR is to fix https://github.com/apache/incubator-pegasus/issues/1177. Once third-parties have changed, image will not be used thus apache-zookeeper-3.7.0-bin.tar.gz should be downloaded.


acelyc111 (approved):

levy5307 (approved):
LGTM

Issue #1162, opened 2022-09-14
detect_hotspot_test.write_hotspot_data failed frequently for UBSAN due to member access within misaligned address


empiredan:
```
2022-09-13 05:54:54 (3.42 MB/s) - 'apache-zookeeper-3.7.0-bin.tar.gz' saved [12387614/12387614]

Decompressing zookeeper...
ZooKeeper JMX enabled by default
Using config: /__w/incubator-pegasus/incubator-pegasus/.zk_install/apache-zookeeper-3.7.0-bin/bin/../conf/zoo.cfg
Starting zookeeper ... STARTED
Zookeeper started at port 22181
starting server
cd /__w/incubator-pegasus/incubator-pegasus/onebox/meta1 && /__w/incubator-pegasus/incubator-pegasus/onebox/meta1/pegasus_server config.ini -app_list meta &>result &
cd /__w/incubator-pegasus/incubator-pegasus/onebox/meta2 && /__w/incubator-pegasus/incubator-pegasus/onebox/meta2/pegasus_server config.ini -app_list meta &>result &
root         282     178  0 05:54 ?        00:00:00 /__w/incubator-pegasus/incubator-pegasus/onebox/meta2/pegasus_server config.ini -app_list meta
cd /__w/incubator-pegasus/incubator-pegasus/onebox/meta3 && /__w/incubator-pegasus/incubator-pegasus/onebox/meta3/pegasus_server config.ini -app_list meta &>result &
root         290     178  0 05:54 ?        00:00:00 /__w/incubator-pegasus/incubator-pegasus/onebox/meta3/pegasus_server config.ini -app_list meta
cd /__w/incubator-pegasus/incubator-pegasus/onebox/replica1 && /__w/incubator-pegasus/incubator-pegasus/onebox/replica1/pegasus_server config.ini -app_list replica &>result &
cd /__w/incubator-pegasus/incubator-pegasus/onebox/replica2 && /__w/incubator-pegasus/incubator-pegasus/onebox/replica2/pegasus_server config.ini -app_list replica &>result &
root         307     178  0 05:54 ?        00:00:00 /__w/incubator-pegasus/incubator-pegasus/onebox/replica2/pegasus_server config.ini -app_list replica
cd /__w/incubator-pegasus/incubator-pegasus/onebox/replica3 && /__w/incubator-pegasus/incubator-pegasus/onebox/replica3/pegasus_server config.ini -app_list replica &>result &
[...]

PR #1187, opened 2022-10-15
chore(CI): disable some flaky tests

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1162

### DIFF
--- a/.github/workflows/lint_and_test_cpp.yaml
+++ b/.github/workflows/lint_and_test_cpp.yaml
@@ -257,13 +257,17 @@ jobs:
           - dsn_runtime_tests
           - dsn_utils_tests
           - dsn.zookeeper.tests
-          - partition_split_test
+          # TODO(yingchun): Disable it because we find it's too flacky, we will re-enable it after
+          # it has been optimized.
+#          - partition_split_test
           - pegasus_geo_test
           - pegasus_rproxy_test
           - pegasus_unit_test
           - recovery_test
           - restore_test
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_and_test_cpp.yaml
```diff
@@ -257,13 +257,17 @@ jobs:
           - dsn_runtime_tests
           - dsn_utils_tests
           - dsn.zookeeper.tests
-          - partition_split_test
+          # TODO(yingchun): Disable it because we find it's too flacky, we will re-enable it after
```
empiredan, line 260:
```suggestion
          # TODO(yingchun): Disable it because we find it's too flaky, we will re-enable it after
```

.github/workflows/lint_and_test_cpp.yaml
```diff
@@ -257,13 +257,17 @@ jobs:
...
+          # it has been optimized.
+#          - partition_split_test
```
empiredan, line 262: Actually `partition_split_test` and `dsn.meta.test` often failed for `test_Release` too, should both of them be disabled temporarily ?

acelyc111: I've updated to disable `partition_split_test` in release type. But I think `dsn.meta.test` can be kept since it will not cost too much time if retring it.

empiredan (approved):

levy5307 (approved):

Issue #1191, opened 2022-10-19
Refactor(metrics): drop template specialization for gauge 


empiredan:
Template specialization is used at the beginning for value initialization of each data type of gauge. Since each data type of gauge must be arithmetical, the value of each type can be zero-initialized by `T()`. Thus template specialization can be dropped.

PR #1192, opened 2022-10-19
refactor(metrics): drop template specialization for gauge

empiredan:
This PR is to solve https://github.com/apache/incubator-pegasus/issues/1191.

### DIFF
--- a/src/rdsn/src/utils/metrics.h
+++ b/src/rdsn/src/utils/metrics.h
@@ -389,32 +389,34 @@ template <typename T, typename = typename std::enable_if<std::is_arithmetic<T>::
 class gauge : public metric
 {
 public:
-    T value() const { return _value.load(std::memory_order_relaxed); }
+    using value_type = T;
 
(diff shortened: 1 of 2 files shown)


acelyc111 (approved):

src/rdsn/src/utils/metrics.h
```diff
@@ -389,53 +389,55 @@ template <typename T, typename = typename std::enable_if<std::is_arithmetic<T>::
 class gauge : public metric
```
levy5307, line 389: Why didn't you change `typename T` to `typename value_type` in the above line?

empiredan: > Why didn't you change `typename T` to `typename value_type` in the above line?

`value_type` may be a combination of multiple template parameters. For example, in std::map, `value_type` is defined as `std::pair<const _Key, _Tp>`. Thus customarily `value_type` is used to decouple from template parameters though for `gauge` it is just the same.

levy5307 (approved):

Issue #1197, opened 2022-10-20
Support to update the configs which are used once when bootstrap


acelyc111:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Some configs are used only once in the process bootstrap stage, and will never use it then. Although Pegasus supports update FLAGS_* at runtime, but they will not be used after bootstrap finished.
For example, `config_sync_interval_ms` is used to init a timer task when replica server bootstrap, but even you update the config (by http) at runtime, it will not take effect. We need an method to update the timer task's schedule interval.

**Describe the feature you'd like:**

We need an feature to update some configs even FLAGS_* only used once.

**Describe alternatives you've considered:**

none.

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #1198, opened 2022-10-20
feat(config): Make the configs take effect for the FLAGS_* only used once

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1197

This patch implement a mechanism to make the configs take effect for the FLAGS_* only used once. They are used once to create some runtime objects, and then will never use the FLAGS_* again, which means update the config by HTTP is meanless.

This patch takes `config_sync_interval_ms` for example, showing how to update a config.

### DIFF
--- a/src/http/http_server.cpp
+++ b/src/http/http_server.cpp
@@ -88,11 +88,11 @@ void http_service::register_handler(std::string sub_path, http_callback cb, std:
         return;
     }
     auto call = make_unique<http_call>();
-    if (this->path().empty()) {
-        call->path = std::move(sub_path);
-    } else {
-        call->path = this->path() + "/" + std::move(sub_path);
+    call->path = this->path();
+    if (!call->path.empty()) {
+        call->path += '/';
     }
+    call->path += sub_path;
     call->callback = std::move(cb);
     call->help = std::move(help);
     http_call_registry::instance().add(std::move(call));
--- a/src/http/http_server.cpp
+++ b/src/http/http_server.cpp
@@ -16,21 +16,37 @@
 // under the License.
 
 #include "http_server.h"
-#include "runtime/tool_api.h"
-#include "utils/time_utils.h"
+
 #include <boost/algorithm/string.hpp>
 #include <fmt/ostream.h>
 
-#include "http_message_parser.h"
-#include "pprof_http_service.h"
 #include "builtin_http_calls.h"
-#include "uri_decoder.h"
 #include "http_call_registry.h"
+#include "http_message_parser.h"
 #include "http_server_impl.h"
+#include "pprof_http_service.h"
+#include "runtime/tool_api.h"
+#include "uri_decoder.h"
+#include "utils/output_utils.h"
+#include "utils/time_utils.h"
 
 namespace dsn {
 
 DSN_DEFINE_bool("http", enable_http_server, true, "whether to enable the embedded HTTP server");
+
+namespace {
+error_s update_config(const http_request &req)
+{
+    if (req.query_args.size() != 1) {
+        return error_s::make(ERR_INVALID_PARAMETERS,
+                             "there should be exactly one config to be updated once");
+    }
+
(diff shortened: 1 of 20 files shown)


neverchanje:
Why didn't you place this config under builtin HTTP service?

acelyc111:
> Why didn't you place this HTTP interface under builtin HTTP service?

Because the builtin HTTP service is stateless, it can't access any specify member of a server(except these singletons).
I've updated the code to define a new base http service, which is able to access members of a specify sub-class, for example, access the `replica_stub` for Replica server.

foreverneverer (dismissed):

src/replica/replica_stub.h
```diff
@@ -230,6 +230,8 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
     // query last checkpoint info for follower in duplication process
     void on_query_last_checkpoint(query_last_checkpoint_info_rpc rpc);
 
+    virtual void update_config(const std::string &name);
```
empiredan, line 233: Is this necessary to be declared as `virtual` ?

src/http/http_server.cpp
```diff
@@ -65,21 +81,38 @@ DSN_DEFINE_bool("http", enable_http_server, true, "whether to enable the embedde
...
     http_call_registry::instance().add(std::move(call));
 }
 
+void http_server_base::update_config_handler(const http_request &req, http_response &resp)
+{
+    auto res = dsn::update_config(req);
+    if (res.is_ok()) {
+        CHECK_EQ(1, req.query_args.size());
+        update_config(req.query_args.begin()->first);
+    }
```
empiredan, line 107: Even the returned `res` is not ok, the status code of http response is still http_status_code::ok, rather than `http_status_code::bad_request`, right ? The error message will be put in response body, as `"update_status"` given by `res.description()`  ?

src/http/http_server.h
```diff
@@ -100,7 +100,31 @@ class http_service
 
     virtual std::string path() const = 0;
 
-    void register_handler(std::string path, http_callback cb, std::string help);
+    void register_handler(std::string sub_path, http_callback cb, std::string help);
```
empiredan, line 103:
```suggestion
    void register_handler(const char *sub_path, http_callback cb, const char *help);
```

src/runtime/task/task.h
```diff
@@ -359,15 +359,18 @@ class timer_task : public task
...
 protected:
     void clear_non_trivial_on_task_end() override { _cb = nullptr; }
 
 private:
-    // ATTENTION: if _interval_milliseconds <= 0, then timer task will just be executed once;
-    // otherwise, timer task will be executed periodically(period = _interval_milliseconds)
-    int _interval_milliseconds;
+    // ATTENTION: if _interval_ms == 0, then timer task will just be executed once;
+    // otherwise, timer task will be executed periodically(period = _interval_ms)
+    int _interval_ms;
```
empiredan, line 370: Declared as `uint32_t` ?

src/http/http_server.cpp
acelyc111: Yes, the http request is ok itself.
And it's the behavior as before https://github.com/apache/incubator-pegasus/pull/1198/files#diff-6d8d586cfd392c8aa927b6d2de0af7e3664180413e518f3a244839b0900c67a9L31

src/runtime/task/task.h
acelyc111: Good suggestion, but I don't want to change it in this patch. There are many callers have to update if changing it, these changes are not closer to this patch topic.

src/http/http_server.h
acelyc111: There may be very little performance improvement, and we have some methods to avoid memory copy. C++ style would be more readable IMO.

empiredan (dismissed):

empiredan (approved):

levy5307 (approved):

Issue #1204, opened 2022-10-26
refactor: use CHECK* macros to replace assert macros


acelyc111:
This refactor work contains the following motivations:
1. macros are offen recommend to use all upper case.
2. the beginning char 'd' is easy to misunderstand as 'only enable in **D**ebug version'
3. there are 2 styles, `dassert_f` and `dassert`, we can unify them

- [x] replace dassert which check a single condition
- [x] replace dcheck_eq*
- [x] replace dcheck_ge*
- [x] replace dcheck_le*
- [x] replace dcheck_gt*
- [x] replace dcheck_lt*

PR #1211, opened 2022-10-27
refactor(macro): use CHECK to replace dassert_f/dassert (part2)

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1204

- use `CHECK` to replace more part of `dassert_f`, which are used to check a single condition
- use `CHECK` to replace more part of `dassert`, which are used to check a single condition. And update to use `fmt` style

### DIFF
--- a/src/nfs/nfs_client_impl.cpp
+++ b/src/nfs/nfs_client_impl.cpp
@@ -450,7 +450,7 @@ void nfs_client_impl::continue_write()
     std::string file_path =
         dsn::utils::filesystem::path_combine(fc->user_req->file_size_req.dst_dir, fc->file_name);
     std::string path = dsn::utils::filesystem::remove_file_name(file_path.c_str());
-    CHECK(dsn::utils::filesystem::create_directory(path), "create directory {} failed");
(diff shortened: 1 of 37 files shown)


levy5307 (dismissed):

src/nfs/nfs_client_impl.cpp
```diff
@@ -449,9 +450,7 @@ void nfs_client_impl::continue_write()
     std::string file_path =
         dsn::utils::filesystem::path_combine(fc->user_req->file_size_req.dst_dir, fc->file_name);
     std::string path = dsn::utils::filesystem::remove_file_name(file_path.c_str());
-    if (!dsn::utils::filesystem::create_directory(path)) {
-        dassert(false, "create directory %s failed", path.c_str());
-    }
+    CHECK(dsn::utils::filesystem::create_directory(path), "create directory {} failed");
```
empiredan, line 453:
```suggestion
    CHECK(dsn::utils::filesystem::create_directory(path), "create directory {} failed", path);
```

levy5307 (approved):

empiredan (approved):

Issue #1218, opened 2022-10-31
Feature(new_metrics): take snapshot of each metric as json format


empiredan:
It has been designed in https://github.com/apache/incubator-pegasus/issues/1206 that the new framework will provide a RESTful service which return current value of each metric as json. Therefore, each metric should firstly support taking snapshot as json format.

An abstract method can be added for base class of metric. And each derived class (i.e., 3 types of metrics: gauge, counter and percentile) will implement this abstract method to take snapshot.

PR #1219, opened 2022-10-31
feat(new_metrics): take snapshot of each metric as json format

empiredan:
This PR is solve https://github.com/apache/incubator-pegasus/issues/1218.

### DIFF
--- a/src/utils/metrics.h
+++ b/src/utils/metrics.h
@@ -345,7 +345,7 @@ class metric : public ref_counter
 public:
     const metric_prototype *prototype() const { return _prototype; }
 
-    // Take snapshot from the metric to json format.
+    // Take snapshot of each metric to collect current values as json format.
     virtual void take_snapshot(dsn::json::JsonWriter &writer) = 0;
 
 protected:
@@ -397,6 +397,14 @@ public:
 
     value_type value() const { return _value.load(std::memory_order_relaxed); }
 
+    // The snapshot collected has following json format:
+    // {
+    //     "name": "<metric_name>",
+    //     "value": ...
+    // }
+    // where "name" is the name of the gauge in string type, and "value" is just current value
+    // of the gauge fetched by `value()`, in numeric types (i.e. integral or floating-point type,
+    // determined by `value_type`).
     virtual void take_snapshot(json::JsonWriter &writer) override
     {
         writer.StartObject();
@@ -510,6 +518,13 @@ public:
         return _adder.fetch_and_reset();
     }
 
+    // The snapshot collected has following json format:
+    // {
+    //     "name": "<metric_name>",
+    //     "value": ...
+    // }
+    // where "name" is the name of the counter in string type, and "value" is just current value
+    // of the counter fetched by `value()`, in integral type (namely int64_t).
     virtual void take_snapshot(json::JsonWriter &writer) override
     {
         writer.StartObject();
@@ -704,6 +719,17 @@ public:
         return _kth_percentile_bitset.test(index);
     }
 
+    // The snapshot collected has following json format:
+    // {
+    //     "name": "<metric_name>",
+    //     "p50": ...,
+    //     "p90": ...,
+    //     "p95": ...,
+    //     ...
+    // }
+    // where "name" is the name of the percentile in string type, with each configured kth
+    // percentile followed, such as "p50", "p90", "p95", etc. All of them are in numeric types
+    // (i.e. integral or floating-point type, determined by `value_type`).
     virtual void take_snapshot(json::JsonWriter &writer) override
     {
         writer.StartObject();
--- a/src/utils/metrics.h
+++ b/src/utils/metrics.h
@@ -33,11 +33,12 @@
 #include <boost/asio/deadline_timer.hpp>
 
 #include "api_utilities.h"
-#include "fmt_logging.h"
 #include "alloc.h"
 #include "autoref_ptr.h"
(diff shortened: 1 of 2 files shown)


src/utils/metrics.h
```diff
@@ -393,6 +397,27 @@ class gauge : public metric
...
 
+    // The snapshot collected has following json format:
+    // {
+    //     "name": "<metric_name>",
+    //     "value": ...
+    // }
+    // where "name" is the name of the gauge in string type, and "value" is just current value
+    // of the gauge fetched by `value()`, in numeric types (i.e. integral or floating-point type,
+    // determined by `value_type`).
+    virtual void take_snapshot(json::JsonWriter &writer) override
```
acelyc111, line 408: better to omit `virtual` coz `override` occured.

src/utils/metrics.h
```diff
@@ -493,6 +518,26 @@ class counter : public metric
...
     }
 
+    // The snapshot collected has following json format:
+    // {
+    //     "name": "<metric_name>",
+    //     "value": ...
+    // }
+    // where "name" is the name of the counter in string type, and "value" is just current value
+    // of the counter fetched by `value()`, in integral type (namely int64_t).
+    virtual void take_snapshot(json::JsonWriter &writer) override
```
acelyc111, line 528: better to omit `virtual` coz `override` occured.

src/utils/metrics.h
```diff
@@ -559,6 +604,7 @@ ENUM_REG(kth_percentile_type::P999)
 ENUM_END(kth_percentile_type)
 
 const std::vector<double> kKthDecimals = {0.5, 0.9, 0.95, 0.99, 0.999};
+const std::vector<std::string> kKthNames = {"p50", "p90", "p95", "p99", "p999"};
```
acelyc111, line 607: how about make them as a map or something, then we are sure that they are in pairs.

src/utils/metrics.h
```diff
@@ -662,10 +715,40 @@ class percentile : public closeable_metric
...
+    //     "name": "<metric_name>",
+    //     "p50": ...,
+    //     "p90": ...,
+    //     "p95": ...,
+    //     ...
+    // }
+    // where "name" is the name of the percentile in string type, with each configured kth
+    // percentile followed, such as "p50", "p90", "p95", etc. All of them are in numeric types
+    // (i.e. integral or floating-point type, determined by `value_type`).
+    virtual void take_snapshot(json::JsonWriter &writer) override
```
acelyc111, line 733: better to omit `virtual` coz `override` occured.

src/utils/test/metrics_test.cpp
```diff
@@ -903,4 +906,254 @@ TEST(metrics_test, percentile_double)
...
+    rapidjson::Document doc;
+    rapidjson::ParseResult result = doc.Parse(json_string.c_str());
+    ASSERT_FALSE(result.IsError());
+
+    ASSERT_TRUE(doc.IsObject());
+    for (const auto &elem : doc.GetObject()) {
+        ASSERT_TRUE(elem.name.IsString());
+
+        if (elem.value.IsString()) {
+            ASSERT_STREQ(elem.name.GetString(), "name");
```
acelyc111, line 938: isn't it always string?

src/utils/test/metrics_test.cpp
```diff
@@ -903,4 +906,254 @@ TEST(metrics_test, percentile_double)
...
+}
+
+template <typename T, typename = typename std::enable_if<std::is_floating_point<T>::value>::type>
+void compare_floating_metric_value_map(const metric_value_map<T> &actual_value_map,
+                                       const metric_value_map<T> &expected_value_map)
+{
+    ASSERT_EQ(actual_value_map.size(), expected_value_map.size());
+
+    typename metric_value_map<T>::const_iterator actual_iter = actual_value_map.begin();
+    typename metric_value_map<T>::const_iterator expected_iter = expected_value_map.begin();
```
acelyc111, line 979: does use `auto` work?

src/utils/test/metrics_test.cpp
```diff
@@ -903,4 +906,254 @@ TEST(metrics_test, percentile_double)
...
+    for (; actual_iter != actual_value_map.end() && expected_iter != expected_value_map.end();
+         ++actual_iter, ++expected_iter) {
+        ASSERT_EQ(actual_iter->first, expected_iter->first);
+        ASSERT_DOUBLE_EQ(actual_iter->second, expected_iter->second);
+    }
+}
+
+#define test_metric_snapshot_with_single_value(                                                    \
```
acelyc111, line 987: How about use all uppper case for macros?

acelyc111 (approved):

src/utils/metrics.h
```diff
@@ -493,6 +518,26 @@ class counter : public metric
...
+    {
+        writer.StartObject();
+
+        writer.Key("name");
+        json::json_encode(writer, prototype()->name().data());
+
+        writer.Key("value");
+        json::json_encode(writer, value());
+
+        writer.EndObject();
```
foreverneverer, line 538: it's same with `guage`,  why not put the `base class`?

empiredan: > it's same with `guage`, why not put the `base class`?

The reason is that `value()` is not an abstract method of the base class. For example, the percentile type does not need a `value()` method, since each has multiple values.

foreverneverer (approved):

Issue #1227, opened 2022-11-07
Improve checking if C string is empty


empiredan:
Sometimes we want to check if a C-style string is empty. For example, in `src/common/common.cpp` there is following code:

```c++
namespace dsn {
DSN_DEFINE_string("replication", cluster_name, "", "name of this cluster");

/*extern*/ const char *get_current_cluster_name()
{
    CHECK_GT_MSG(strlen(FLAGS_cluster_name), 0, "cluster_name is not set");
    return FLAGS_cluster_name;
}
} // namespace dsn
```

Checking by `strlen()` will be slow since strlen will traverse the whole string especially the string is very long.

According to [Apache Commons Lang](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty-java.lang.CharSequence-), firstly we can define that a C-style string is empty once the string:

- is nullptr, or
- does not include any character.

Therefore, checking if a C-style string is empty can just be implemented as:
```c++
inline bool is_empty(const char *str)
{
    return str == nullptr || *str == '\0';
}
```

PR #1229, opened 2022-11-07
feat: improve checking if C string is empty

empiredan:
This PR is to implement https://github.com/apache/incubator-pegasus/issues/1227.

### DIFF
--- a/src/utils/strings.h
+++ b/src/utils/strings.h
@@ -36,6 +36,11 @@
 namespace dsn {
 namespace utils {
 
+inline bool is_empty(const char *str)
+{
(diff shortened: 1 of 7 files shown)


src/utils/strings.h
```diff
@@ -36,6 +36,8 @@
 namespace dsn {
 namespace utils {
 
+inline bool is_empty(const char *str) { return str == nullptr || *str == '\0'; }
```
acelyc111, line 39: How about adding some test?

acelyc111 (dismissed):

acelyc111 (approved):

Issue #1236, opened 2022-11-10
Feature(new_metrics): support filters for selected metric fields in response


empiredan:
## 1. Motivation
According to https://github.com/apache/incubator-pegasus/issues/1206, `with_metric_fields` and `without_metric_fields` will be provided as the parameters in the RESTful API to filter the fields of metric in the response to client.

## 2. Usage for `with_metric_fields`
`with_metric_fields` will ask for the metric fields that are **not** included in the request. For example, if the request is given as below:

```url
/metrics?with_metric_fields=value,p99
```

The response will only contain the fields of "**value**" and "**p99**":

```json
[
    { // entity 1
        "metrics": [
            // gauge
            {
                "value": 100
            },
            // counter
            {
                "value": 10000
            },
            // percentile
            {
                "p99": 10
            }
        ]
    },
    ...
]
```

## 3. Usage for `without_metric_fields`

Similarly, `without_metric_fields` will ask for the metric fields that are **not** included in the request. For example, if the request is given as below:

```url
/metrics?without_metric_fields=value,p99
```

The response will contain all the fields except "**value**" and "**p99**":

```json
[
    { // entity 1
        "metrics": [
            // gauge
            {
                "name": "<gauge_name>"
            },
            // counter
            {
                "name": "<counter_name>"
            },
            // percentile
            {
                "name": "<percentile_name>",
                "p50": 1,
                "p90": 5,
                "p95": 8,
                "p999": 15
            }
        ]
    },
    ...
]
```

## 4. Wrong usage

Once both `with_metric_fields` and `without_metric_fields` are provided in the request, it will considered as invalid. We can discuss the reason in 2 conditions.

Firstly, suppose both `with_metric_fields` and `without_metric_fields` are provided, and they include the same fields. In this condition, each conflicts with another. For example, `with_metric_fields=a,b` and `without_metric_fields=b,c` will lead to contradiction: we cannot decide if `b` should be put in the response to client.

On the contrary, suppose `with_metric_fields` and `without_metric_fields` do not include any same field. In this condition, they can be simplified as a unique `with_metric_fields`. For example, `with_metric_fields=a,b` and `without_metric_fields=c,d` can be simplified as `with_metric_fields=a,b`: `without_metric_fields=c,d` is useless here since we have declared that only `a` and `b` fields are needed according to `with_metric_fields=a,b`.

Therefore, we can draw a conclusion that `with_metric_fields` and `without_metric_fields` should not be provided at the same time.

PR #1237, opened 2022-11-11
feat(new_metrics): support filters for selected metric fields in response

empiredan:
This PR is to solve https://github.com/apache/incubator-pegasus/issues/1236.

### DIFF
--- a/src/utils/metrics.cpp
+++ b/src/utils/metrics.cpp
@@ -164,13 +164,6 @@ metric_prototype::metric_prototype(const ctor_args &args) : _args(args) {}
 
 metric_prototype::~metric_prototype() {}
 
-metric_fields_type get_all_single_value_metric_fields()
-{
-    metric_fields_type fields = kMetricPrototypeFields;
-    fields.insert(kMetricSingleValueField);
-    return fields;
-}
-
 metric::metric(const metric_prototype *prototype) : _prototype(prototype) {}
 
 closeable_metric::closeable_metric(const metric_prototype *prototype) : metric(prototype) {}
@@ -272,15 +265,6 @@ std::set<kth_percentile_type> get_all_kth_percentile_types()
         all_types.insert(static_cast<kth_percentile_type>(i));
     }
     return all_types;
-}
-
-metric_fields_type get_all_kth_percentile_fields()
-{
-    metric_fields_type fields = {kMetricNameField};
-    for (const auto &kth : kAllKthPercentiles) {
-        fields.insert(kth.name);
-    }
-    return fields;
 }
 
 } // namespace dsn
--- a/src/utils/metrics.cpp
+++ b/src/utils/metrics.cpp
@@ -175,6 +175,9 @@ metric_prototype::~metric_prototype() {}
 
 metric::metric(const metric_prototype *prototype) : _prototype(prototype) {}
 
+const std::string metric::kMetricNameField = "name";
+const std::string metric::kMetricSingleValueField = "value";
+
 closeable_metric::closeable_metric(const metric_prototype *prototype) : metric(prototype) {}
 
 uint64_t percentile_timer::generate_initial_delay_ms(uint64_t interval_ms)
--- a/src/utils/metrics.cpp
+++ b/src/utils/metrics.cpp
@@ -258,13 +258,4 @@ void percentile_timer::on_timer(const boost::system::error_code &ec)
 #undef TRY_PROCESS_TIMER_CLOSING
 }
 
-std::set<kth_percentile_type> get_all_kth_percentile_types()
-{
-    std::set<kth_percentile_type> all_types;
-    for (size_t i = 0; i < static_cast<size_t>(kth_percentile_type::COUNT); ++i) {
-        all_types.insert(static_cast<kth_percentile_type>(i));
-    }
-    return all_types;
-}
-
 } // namespace dsn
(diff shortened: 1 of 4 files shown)


src/utils/metrics.h
```diff
@@ -131,6 +132,61 @@
...
+
+// This struct includes a set of filters for both entities and metrics requested by client.
+struct metric_filters
+{
+    using fields_filter_fn = std::function<bool(const std::string &)>;
+
+    // According to the parameters requested by client, this function will filter metric
+    // fields that will be put in the response. `with_metric_fields` includes all the metric
+    // fields that are wanted by client; on the contrary, once client wants all except some
+    // metric fields, it could specify them in `without_metric_fields`.
```
acelyc111, line 145: Since there are dozens of metrics, what's the use case of specify 'without_metric_fields'?

src/utils/metrics.h
```diff
@@ -405,15 +498,12 @@ class gauge : public metric
...
+    void take_snapshot(json::JsonWriter &writer, const metric_filters &filters) override
     {
         writer.StartObject();
 
-        writer.Key("name");
-        json::json_encode(writer, prototype()->name().data());
-
-        writer.Key("value");
-        json::json_encode(writer, value());
+        encode_name(writer, filters);
```
acelyc111, line 505: if the result has been filtered, is the output like:
```
{
}
```
rather than:
```
```

src/utils/metrics.h
```diff
@@ -736,20 +826,18 @@ class percentile : public closeable_metric
...
-    void take_snapshot(json::JsonWriter &writer) override
+    void take_snapshot(json::JsonWriter &writer, const metric_filters &filters) override
     {
         writer.StartObject();
 
-        writer.Key("name");
-        json::json_encode(writer, prototype()->name().data());
+        encode_name(writer, filters);
 
         for (size_t i = 0; i < static_cast<size_t>(kth_percentile_type::COUNT); ++i) {
```
acelyc111, line 835: can it skip the loop?

src/utils/metrics.h
empiredan: > Since there are dozens of metrics, what's the use case of specify 'without_metric_fields'?

For example, `without_metric_fields` will be useful while we request for all fields except `desc`.

acelyc111: It's a bit of complex, `without_metric_fields` use cases can be replaced by just using `with_metric_fields`.

If **metric fields** have both `with` and `without` filters, do the **metric names**, **entity** have the 2 filters? I think it's rarely because there maybe dozens or hundreds of metrics/entities, how do a user just want metrics excludes a few ones?

empiredan: > It's a bit of complex, `without_metric_fields` use cases can be replaced by just using `with_metric_fields`.
> 
> If **metric fields** have both `with` and `without` filters, do the **metric names**, **entity** have the 2 filters? I think it's rarely because there maybe dozens or hundreds of metrics/entities, how do a user just want metrics excludes a few ones?

OK, I'll drop `without_metric_fields`.

src/utils/metrics.h
empiredan: Yes.

src/utils/metrics.h
```diff
@@ -331,6 +353,17 @@ class metric_prototype_with : public metric_prototype
     DISALLOW_COPY_AND_ASSIGN(metric_prototype_with);
 };
 
+const std::string kMetricTypeField = "type";
+const std::string kMetricNameField = "name";
+const std::string kMetricUnitField = "unit";
+const std::string kMetricDescField = "desc";
+const metric_fields_type kMetricPrototypeFields = {
```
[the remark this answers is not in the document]
empiredan: > Will kMetricPrototypeFields be used anywhere else except get_all_single_value_metric_fields(), if not, make it there as a local variable and do not expose it in header.

OK, I'll change these variables as local.

src/utils/metrics.h
```diff
@@ -131,6 +132,26 @@
...
+
+// This struct includes a set of filters for both entities and metrics requested by client.
+struct metric_filters
+{
+    // According to the parameters requested by client, this function will filter metric
+    // fields that will be put in the response. `with_metric_fields` includes all the metric
+    // fields that are wanted by client.
+    bool has_metric_field(const std::string &field_name) const
+    {
+        if (with_metric_fields.empty()) {
```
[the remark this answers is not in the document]
empiredan: Actually when `with_metric_fields` is empty, this means every metric field is needed. I'll add more comments for this.

acelyc111 (approved):
LGTM

src/utils/metrics.cpp
```diff
@@ -267,4 +258,13 @@ void percentile_timer::on_timer(const boost::system::error_code &ec)
 #undef TRY_PROCESS_TIMER_CLOSING
 }
 
+std::set<kth_percentile_type> get_all_kth_percentile_types()
+{
+    std::set<kth_percentile_type> all_types;
+    for (size_t i = 0; i < static_cast<size_t>(kth_percentile_type::COUNT); ++i) {
```
levy5307, line 264: ```   for (size_t i = static_cast<size_t>(kth_percentile_type::COUNT - 1); i >= 0; --i) {```
reduce some cast operation

empiredan: > ` for (size_t i = static_cast<size_t>(kth_percentile_type::COUNT - 1); i >= 0; --i) {` reduce some cast operation

OK, I'll use macro to simplify this.

acelyc111 (approved):

levy5307 (approved):

Issue #1234, opened 2022-11-10
Code refactor on rdsn module


acelyc111:
Inspired by and thanks to [rdsn](https://github.com/microsoft/rDSN), the Pegasus project could be developed nowadays.

The xiaomi/rdsn repo has been merged into apache/pegasus for months, now I consider the Pegasus branch rdsn is deeply modified and only used for Pegasus, and it's very hard and maybe meaningless to contribute these changes back upstream, so I'm planning to continue the work of refactoring rdsn module, including:

- [ ] remove the C language sopporting
- [ ] make it as a internel module but not a public library
- [ ] ...

PR #1238, opened 2022-11-11
refactor: use linux_fd_t instead of dsn_handle_t for file descriptor

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1234

including changes:
1. change `dsn_handle_t` which is used for files descriptor to `linux_fd_t` which wrap an `int`
2. remove useless `dsn_handle_t` type member from `class aio_context`, `struct callback_para`
3. change `file_object` from `void *` to `disk_file *`

### DIFF
--- a/src/aio/disk_engine.cpp
+++ b/src/aio/disk_engine.cpp
@@ -253,7 +253,7 @@ void disk_engine::complete_io(aio_task *aio, error_code err, uint64_t bytes)
         auto dfile = aio->get_aio_context()->dfile;
         if (aio->get_aio_context()->type == AIO_Read) {
             auto wk = dfile->on_read_completed(aio, err, (size_t)bytes);
-            if (wk != nullptr) {
+            if (wk) {
                 _provider->submit_aio_task(wk);
             }
         }
@@ -262,7 +262,7 @@ void disk_engine::complete_io(aio_task *aio, error_code err, uint64_t bytes)
         else {
             uint64_t sz;
             auto wk = dfile->on_write_completed(aio, (void *)&sz, err, (size_t)bytes);
-            if (wk != nullptr) {
+            if (wk) {
                 process_write(wk, sz);
             }
         }
--- a/src/aio/disk_engine.cpp
+++ b/src/aio/disk_engine.cpp
@@ -182,9 +182,10 @@ void disk_engine::write(aio_task *aio)
     }
 
     auto dio = aio->get_aio_context();
+    CHECK_NE(dio->fd, DSN_INVALID_FILE_HANDLE);
     auto dfile = (disk_file *)(void *)(uintptr_t)(dio->fd);
-    dio->fd = dfile->native_handle();
-    dio->dfile = dfile;
+    CHECK_EQ(dio->fd, dfile->native_handle());
+    dio->dfile = dfile; // dio->dfile is initialized here
     dio->engine = this;
     dio->type = AIO_Write;
 
--- a/src/aio/disk_engine.cpp
+++ b/src/aio/disk_engine.cpp
@@ -91,7 +91,7 @@ aio_task *disk_write_queue::unlink_next_workload(void *plength)
     return first;
 }
 
-disk_file::disk_file(dsn_handle_t handle) : _handle(handle) {}
+disk_file::disk_file(int fd) : _fd(fd) {}
 
 aio_task *disk_file::read(aio_task *tsk)
 {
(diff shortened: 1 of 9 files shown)


neverchanje:
I personally suggest that we should mix the uses of int and the file handle type. It will easily lead to some misuse. Maybe you can wrap int under a `struct linux_fd_t { int v; }` in order to prevent any int to be converted to `linux_fd_t` without explicit declaration.

acelyc111:
> I personally suggest that we shouldn't mix the uses of int and the file handle type. It will easily lead to some misuse. Maybe you can wrap int under a `struct linux_fd_t { int v; }` in order to prevent any int to be converted to `linux_fd_t` without explicit declaration.

Good suggestion, done.

neverchanje (dismissed):
LGTM

src/aio/disk_engine.cpp
```diff
@@ -91,7 +91,7 @@ aio_task *disk_write_queue::unlink_next_workload(void *plength)
     return first;
 }
 
-disk_file::disk_file(dsn_handle_t handle) : _handle(handle) {}
+disk_file::disk_file(linux_fd_t fd) : _fd(fd.fd) {}
```
empiredan, line 94: Use copy constructor `_fd(fd)` instead ?

src/aio/file_io.cpp
```diff
@@ -32,12 +32,12 @@ namespace file {
 
 /*extern*/ disk_file *open(const char *file_name, int flag, int pmode)
 {
-    dsn_handle_t nh = disk_engine::provider().open(file_name, flag, pmode);
-    if (nh != DSN_INVALID_FILE_HANDLE) {
-        return new disk_file(nh);
-    } else {
+    linux_fd_t fd = disk_engine::provider().open(file_name, flag, pmode);
```
empiredan, line 35:
```suggestion
    auto fd = disk_engine::provider().open(file_name, flag, pmode);
```

src/aio/disk_engine.cpp
```diff
@@ -256,9 +250,9 @@ void disk_engine::complete_io(aio_task *aio, error_code err, uint64_t bytes)
 
     // no batching
     else {
-        auto df = (disk_file *)(aio->get_aio_context()->file_object);
+        auto dfile = aio->get_aio_context()->dfile;
         if (aio->get_aio_context()->type == AIO_Read) {
-            auto wk = df->on_read_completed(aio, err, (size_t)bytes);
+            auto wk = dfile->on_read_completed(aio, err, (size_t)bytes);
             if (wk) {
```
empiredan, line 256:
```suggestion
            if (wk != nullptr) {
```

src/aio/disk_engine.cpp
```diff
@@ -267,7 +261,7 @@ void disk_engine::complete_io(aio_task *aio, error_code err, uint64_t bytes)
         // write
         else {
             uint64_t sz;
-            auto wk = df->on_write_completed(aio, (void *)&sz, err, (size_t)bytes);
+            auto wk = dfile->on_write_completed(aio, (void *)&sz, err, (size_t)bytes);
             if (wk) {
```
empiredan, line 265:
```suggestion
            if (wk != nullptr) {
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -42,33 +42,33 @@ native_linux_aio_provider::native_linux_aio_provider(disk_engine *disk) : aio_pr
 
 native_linux_aio_provider::~native_linux_aio_provider() {}
 
-dsn_handle_t native_linux_aio_provider::open(const char *file_name, int flag, int pmode)
+linux_fd_t native_linux_aio_provider::open(const char *file_name, int flag, int pmode)
 {
-    dsn_handle_t fh = (dsn_handle_t)(uintptr_t)::open(file_name, flag, pmode);
-    if (fh == DSN_INVALID_FILE_HANDLE) {
+    int fd = ::open(file_name, flag, pmode);
```
empiredan, line 47:
```suggestion
    auto fd = ::open(file_name, flag, pmode);
```

src/aio/native_linux_aio_provider.cpp
```diff
@@ -42,33 +42,33 @@ native_linux_aio_provider::native_linux_aio_provider(disk_engine *disk) : aio_pr
...
 {
-    if (fh == DSN_INVALID_FILE_HANDLE || ::close((int)(uintptr_t)(fh)) == 0) {
+    if (fd.is_invalid() || ::close(fd.fd) == 0) {
         return ERR_OK;
-    } else {
-        LOG_ERROR("close file failed, err = %s", strerror(errno));
-        return ERR_FILE_OPERATION_FAILED;
     }
+
+    LOG_ERROR_F("close file failed, err = {}", strerror(errno));
```
empiredan, line 60:
```suggestion
    LOG_ERROR_F("close file failed, err = {}", utils::safe_strerror(errno));
```

empiredan (approved):

foreverneverer (approved):

Issue #1239, opened 2022-11-14
Refactor: improve split_args for strings


empiredan:
While analyzing the code of `split_args` and `trim_string`, some problems could be found.

- Both `std::string v(args)` and `v.substr(...)` will copy string, which will lead to overhead for performance.
- In `trim_string`, all trailing spaces will be updated with `\0`, which is unnecessary.

Based on the above problems, I think we can improve `split_args`.

The code of current `split_args` and `trim_string` are listed as below:

```c++
void split_args(const char *args,
                /*out*/ std::vector<std::string> &sargs,
                char splitter,
                bool keep_place_holder)
{
    sargs.clear();
    std::string v(args);
    uint64_t last_pos = 0;
    while (true) {
        auto pos = v.find(splitter, last_pos);
        if (pos != std::string::npos) {
            std::string s = trim_string((char *)v.substr(last_pos, pos - last_pos).c_str());
            if (!s.empty()) {
                sargs.push_back(s);
            } else if (keep_place_holder) {
                sargs.emplace_back("");
            }
            last_pos = pos + 1;
        } else {
            std::string s = trim_string((char *)v.substr(last_pos).c_str());
            if (!s.empty()) {
                sargs.push_back(s);
            } else if (keep_place_holder) {
                sargs.emplace_back("");
            }
            break;
        }
    }
}
```

```c++
char *trim_string(char *s)
{
    while (*s != '\0' && (*s == ' ' || *s == '\t')) {
        s++;
    }
    char *r = s;
    s += strlen(s);
    while (s >= r && (*s == '\0' || *s == ' ' || *s == '\t' || *s == '\r' || *s == '\n')) {
        *s = '\0';
        s--;
    }
    return r;
}
```

PR #1240, opened 2022-11-14
refactor: use state machine to improve split_args for strings

empiredan:
This PR is to solve https://github.com/apache/incubator-pegasus/issues/1239.

We can scan the whole string and use state machine to extract the tokens without any unnecessary copy and update.


acelyc111 (approved):

empiredan:
> I believe that we shouldn't invest too much effort in writing commonly-used utilities on our own. Instead, we should find if there's a trustworthy library providing this functionality, such as https://abseil.io/docs/cpp/guides/strings#abslstrsplit-for-splitting-strings or https://www.boost.org/doc/libs/1_80_0/doc/html/string_algo/usage.html.
> 
> I feel that C++ coders rarely consider relying on the community. Programmers using other languages would never think of writing their string split unless there are some special requirements. Therefore, if possible, please try out the above two libraries and see if they can address your problem.

Good advice ! I strongly agree that we should use mature libraries for basic tools. And initially I was going to use [boost split](https://www.boost.org/doc/libs/1_69_0/doc/html/boost/algorithm/split.html) since we'd introduced boost 1.69.0 as our third-party package. However, according to the doc boost only supports outputting tokens to sequence containers such as `std::vector` or `std::list` even in newest version 1.80.0. Actually tokens are also extracted to associative containers such as `unordered_set` which has already been supported by old `split_args`.

According to [str_split_test.cc](https://github.com/abseil/abseil-cpp/blob/master/absl/strings/str_split_test.cc), it seems that abseil has supported all types of containers. However, there is some special logic for `split_args`: for example, after split for each time, old `split_args` will strip the leading and trailing spaces; moreover, the characters belonging to leading spaces are not the same with those belonging to trailing spaces though I wonder why this happened and which code has depended on this.

I think by [StripAsciiWhitespace](https://github.com/abseil/abseil-cpp/blob/master/absl/strings/ascii.h) each split can be stripped. However, this will erase the string while new `split_args` will just skip leading and trailing spaces of tokens after split without extra stripping operations. Also, the reason why inequality of characters of leading and trailing spaces should be checked. Abseil uses `absl::ascii_isspace` to check both leading and trailing spaces. Even if striping can be implemented by ourselves, it will have to do extra operations and another copy.

To sum up, I think we can just `split_args` firstly. I believe the common library will function nevertheless. However, it will cost some time to verify. I'll create other new issues to track. This is an important job that will be supported for long term.

Issue #1199, opened 2022-10-23
Adjust the strange LOG level


acelyc111:
For most logging systems, levels are ordered in:
DEBUG,INFO,WARNING,ERROR,FATAL
for example: [spdlog](https://github.com/gabime/spdlog/blob/d546201f127c306ec8a0082d57562a05a049af77/include/spdlog/common.h#L214), [glog](https://github.com/google/glog/blob/master/src/glog/log_severity.h#L53), [Log4J2](https://logging.apache.org/log4j/2.x/manual/customloglevels.html)
But in Pegasus, they are ordered in:
INFO,DEBUG,WARNING,ERROR,FATAL

It's strange, we should adjust the levels to keep consistent with well known and wildly used logging systems.

- [x] https://github.com/apache/incubator-pegasus/pull/1200
- [x] https://github.com/apache/incubator-pegasus/pull/1201
- [x] https://github.com/apache/incubator-pegasus/pull/1202

PR #1246, opened 2022-11-17
fix: fix log header not corresponding to level

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1199

In commit https://github.com/apache/incubator-pegasus/pull/1200, we exchange DEBUG level and INFO level, but the log header prefixes have been forgot to update. This patch fix it.

### DIFF
--- a/src/utils/simple_logger.cpp
+++ b/src/utils/simple_logger.cpp
@@ -56,6 +56,12 @@ DSN_DEFINE_validator(stderr_start_level, [](const char *level) -> bool {
 
 static void print_header(FILE *fp, dsn_log_level_t log_level)
 {
+    // The leading character of each log lines, corresponding to the log level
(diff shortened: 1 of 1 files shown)


totalo (dismissed):

src/utils/simple_logger.cpp
```diff
@@ -56,7 +56,7 @@ DSN_DEFINE_validator(stderr_start_level, [](const char *level) -> bool {
 
 static void print_header(FILE *fp, dsn_log_level_t log_level)
 {
-    static char s_level_char[] = "IDWEF";
```
levy5307, line 59: It's better to add an explain to show what is 'I'/'D', and so on.

foreverneverer (approved):

empiredan (approved):

PR #1248, opened 2022-11-19
refactor: enable warning on unused result

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1166

Deal with function return value to ensure seriours error could be found immediatelly.
You can add macro `WARN_UNUSED_RESULT` after the function declaration to mark the function's return value must be dealt.
This patch add the macro and make `command_manager::register_command` as `WARN_UNUSED_RESULT`, all callers have been refactored to adapt this semantics. It's a guarantee that related objects are destroyed in a meaningful order.


src/utils/logging.cpp
```diff
@@ -158,13 +128,11 @@ void dsn_log(const char *file,
 
 namespace dsn {
 
-std::unique_ptr<logging_provider> logging_provider::_logger =
-    std::unique_ptr<logging_provider>(nullptr);
+std::unique_ptr<logging_provider> logging_provider::_logger = nullptr;
```
empiredan, line 131:
```suggestion
std::unique_ptr<logging_provider> logging_provider::_logger;
```

src/utils/simple_logger.h
```diff
@@ -71,22 +72,22 @@ class simple_logger : public logging_provider
 {
 public:
     simple_logger(const char *log_dir);
-    virtual ~simple_logger(void);
+    ~simple_logger(void);
```
empiredan, line 75: Should we mark `override` for those destructors whose ancestors are declared as `virtual` ?

src/utils/logging_provider.h
```diff
@@ -80,10 +74,16 @@ class logging_provider
 
     virtual void flush() = 0;
 
-private:
+    void deregister_commands() { _cmds.clear(); }
+
+protected:
+    logging_provider(const char *) {}
```
empiredan, line 80:
```suggestion
    logging_provider(const char *) = default;
```

src/utils/logging_provider.h
```diff
@@ -55,9 +51,7 @@ class logging_provider
     typedef logging_provider *(*factory)(const char *);
 
 public:
-    logging_provider(const char *) {}
-
-    virtual ~logging_provider(void) {}
+    virtual ~logging_provider() {}
```
empiredan, line 54:
```suggestion
    virtual ~logging_provider() = default;
```

src/block_service/directio_writable_file.cpp
```diff
@@ -96,12 +103,16 @@ bool direct_io_writable_file::finalize()
 
     if (_offset > 0) {
         if (::write(_fd, _buffer, _buffer_size) != _buffer_size) {
```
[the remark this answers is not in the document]
acelyc111: We can do that, but not in this patch.
TODO added.

src/utils/simple_logger.h
acelyc111: I think it would be better to mark `override`.
https://github.com/isocpp/CppCoreGuidelines/issues/721

src/utils/logging_provider.h
acelyc111: Removed it, we don't need it.

empiredan (approved):

foreverneverer (approved):

Issue #1245, opened 2022-11-16
Feature(new_metrics): take snapshot of each entity as json format


empiredan:
As each metric has supported snapshot as json format in https://github.com/apache/incubator-pegasus/issues/1218, entity should also support this feature just like what https://github.com/apache/incubator-pegasus/issues/1206 has described.

Firstly, REST service will construct the filters in `metric_filters` according to the request by client. Then, the filters are used to select the entities and metrics conforming to the following rules:

- Empty filter means there is no need to check for this filter, just go ahead with the next filter;
- If the filter for `types` is not empty, the type of each selected entity should be included in the filter;
- If the filter for `ids` is not empty, the ID of each selected entity should be included in the filter;
- If the filter for `attributes` is not empty, at least one attribute of each selected entity should be included in the filter (both name and value of this attribute should be matched with the filter);
- If the filter for `metrics` is not empty, each selected metric should be included in the filter.

If the entity is not chosen by the filters, or none of its metrics is selected, nothing for this entity will be put in the response returned back to the client; otherwise, if at least one metric of the entity is selected, following fields will be put in the response as json format: `type`,  `id`, `attributes` and `metrics`. Even the value of any field is empty, for example, there is not any attribute for an entity, attributes will still be put in response in json format `"attributes": {}`.

PR #1258, opened 2022-11-28
feat(new_metrics): take snapshot of each entity as json format

empiredan:
This PR is to solve https://github.com/apache/incubator-pegasus/issues/1245.


src/utils/metrics.h
```diff
@@ -222,6 +225,95 @@ class metric_entity : public ref_counter
...
+    bool match_entity_attrs(const metric_entity::attr_map &candidates) const
+    {
+        RETURN_MATCHED_WITH_EMPTY_WHITE_LIST(entity_attrs);
+
+        // The size of container must be divisible by 2, since attribute name always pairs
+        // with value in it.
+        CHECK_EQ(entity_attrs.size() & 1, 0);
+
+        for (entity_attrs_type::size_type i = 0; i < entity_attrs.size(); i += 2) {
+            metric_entity::attr_map::const_iterator iter = candidates.find(entity_attrs[i]);
```
acelyc111, line 284:
```suggestion
            const auto &iter = candidates.find(entity_attrs[i]);
```

src/utils/metrics.cpp
```diff
@@ -24,8 +24,10 @@
 
 namespace dsn {
 
-metric_entity::metric_entity(const std::string &id, attr_map &&attrs)
-    : _id(id), _lock(), _attrs(std::move(attrs)), _metrics()
+metric_entity::metric_entity(const metric_entity_prototype *prototype,
+                             const std::string &id,
+                             const attr_map &attrs)
+    : _prototype(prototype), _id(id), _lock(), _attrs(attrs), _metrics()
```
acelyc111, line 30: How about let _lock and _metrics construct implicitly and automatically?

src/utils/metrics.cpp
```diff
@@ -81,19 +83,115 @@ metric_entity::metric_map metric_entity::metrics() const
...
+
+void metric_entity::encode_id(metric_json_writer &writer) const
+{
+    writer.Key(kMetricEntityIdField.c_str());
+    json::json_encode(writer, _id);
+}
+
+/*static*/ void metric_entity::encode_attrs(metric_json_writer &writer, const attr_map &attrs)
+{
+    // Empty attributes are allowed and will just be encoded as {}.
```
acelyc111, line 106: It's a bit of confused, why design it like this?

src/utils/metrics.h
```diff
@@ -209,8 +201,19 @@ class metric_entity : public ref_counter
...
     void close(close_option option);
 
-    void set_attributes(attr_map &&attrs);
+    void set_attributes(const attr_map &attrs);
+
+    void encode_type(metric_json_writer &writer) const;
+
+    void encode_id(metric_json_writer &writer) const;
+
+    static void encode_attrs(metric_json_writer &writer, const attr_map &attrs);
```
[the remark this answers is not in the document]
empiredan: Good advice, I'll move this function to `.cpp`.

src/utils/metrics.h
```diff
@@ -209,8 +201,19 @@ class metric_entity : public ref_counter
...
 
+    static void encode_metrics(metric_json_writer &writer,
```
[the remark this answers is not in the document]
empiredan: OK.

src/utils/metrics.cpp
empiredan: OK.

src/utils/metrics.cpp
empiredan: This is learned from Kudu metrics. For entity type of `server`, its attributes are always empty; it will be returned in json format as `{}`:
image

acelyc111 (approved):

foreverneverer (approved):

src/utils/metrics.h
kirbyzhou: "const auto& iter = ..." will get a const iterator, not a const_iterator, which is different.

Issue #1261, opened 2022-11-29
Refactor: remove zkclient dependency


Apache9:
In the current java client implementation, we only need to read zk once when loading configuration. We do not need to dependency on any wrapper implementation, just use the official zookeeper client is enough.

Apache9:
Fixed by #1264

PR #1264, opened 2022-11-29
refactor: remove zkclient dependency

Apache9:
https://github.com/apache/incubator-pegasus/issues/1261

Use official zookeeper client directly as we only use a very small set of feature in java client implementation. Exclude unnecessary transitive dependencies such as jline and log4j. Remove the shading and relocating for zk client as we do not depend on it any more.

I choose zookeeper version 3.4.14 because it is the oldest version which does not have a CVE.


java-client/src/test/resources/simplelogger.properties
```diff
@@ -17,9 +17,6 @@
...
 
-log4j.rootLogger=DEBUG, stdout
-
-log4j.appender.stdout=org.apache.log4j.ConsoleAppender
-log4j.appender.stdout.Target=System.out
-log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
-log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
+org.slf4j.simpleLogger.defaultLogLevel=debug
+org.slf4j.simpleLogger.showDateTime=true
+org.slf4j.simpleLogger.dateTimeFormat=yyyy/MM/dd HH:mm:ss.SSS
```
empiredan, line 22: Seems that newline is missing at the end of file ?

Apache9: Done.

empiredan (approved):

acelyc111 (approved):

Issue #1283, opened 2022-12-04
Unify the *.thrift files use by cpp server and all clients


acelyc111:
Current now, cpp server and all clients were using their own idl thrift files to generate code, it's hard to keep consistency and may introduce bugs by mistaken.
We are planning to unify the idl thrift files.

PR #1284, opened 2022-12-04
refactor(idl): unify the *.thrift files used by cpp and go-client

acelyc111:

This patch including changes:
- move common *.thrift files from `src/**/*` to `idl/*`
- make an independed file `src/common/duplication_internal.thrift`, it is used by server internally
- remove `go-client/idl/*.thrift` files, and remove the go-lang files generated by the former
- update `.licenserc.yaml` because some files are moved or removed
- run `thrift --gen` at first in `go-client/Makefile` to generate files automatically
- update some structures' and fields' name, there are differents of *.thrift files between `src/**/*` and `go-client/idl/*` before
- move lint step after build step in `.github/workflows/lint_and_test_go-client.yml` because some files are lacked before building, they will be generated in build stage

## Compatiblity changes for go client

| old name | new name |
| ---- | ---- |
| DropAppRequest | ConfigurationDropAppRequest |
| DropAppResponse | ConfigurationDropAppResponse |
| CreateAppRequest | ConfigurationCreateAppRequest |
| CreateAppResponse | ConfigurationCreateAppResponse |
| RecallAppRequest | ConfigurationRecallAppRequest |
| RecallAppResponse | ConfigurationRecallAppResponse |
| ListAppsRequest | ConfigurationListAppsRequest |
| ListAppsResponse | ConfigurationListAppsResponse |
| UpdateAppEnvRequest | ConfigurationUpdateAppEnvRequest |
| UpdateAppEnvResponse | ConfigurationUpdateAppEnvResponse |
| ListNodesRequest | ConfigurationListNodesRequest |
| ListNodesResponse | ConfigurationListNodesResponse |
| ClusterInfoRequest | ConfigurationClusterInfoRequest |
| ClusterInfoResponse | ConfigurationClusterInfoResponse |
| MetaControlRequest | ConfigurationMetaControlRequest |
| MetaControlResponse | ConfigurationMetaControlResponse |
| QueryBackupPolicyRequest | ConfigurationQueryBackupPolicyRequest |
| QueryBackupPolicyResponse | ConfigurationQueryBackupPolicyResponse |
| BalanceRequest | ConfigurationBalancerRequest |
| BalanceResponse | ConfigurationBalancerResponse |
| RestoreAppRequest | ConfigurationRestoreRequest |
| CreateAppResponse | ConfigurationCreateAppResponse |


.github/workflows/lint_and_test_go-client.yml
```diff
@@ -36,27 +36,12 @@ on:
...
-          working-directory: ./go-client
-
   build:
-    name: Test
-    runs-on: ubuntu-latest
+    name: Test and Lint
+    runs-on: ubuntu:20.04
     steps:
+    - name: Install thrift
+      run: sudo apt-get install -y thrift-compiler
```
Apache9, line 44: We do not need to specify the version of the thrift compiler?

go-client/generator/admin.csv
```diff
@@ -1,20 +1,20 @@
-RPC_CM_DROP_APP,DropApp,DropAppRequest,DropAppResponse
-RPC_CM_CREATE_APP,CreateApp,CreateAppRequest,CreateAppResponse
-RPC_CM_RECALL_APP,RecallApp,RecallAppRequest,RecallAppResponse
-RPC_CM_LIST_APPS,ListApps,ListAppsRequest,ListAppsResponse
+RPC_CM_DROP_APP,DropApp,ConfigurationDropAppRequest,ConfigurationDropAppResponse
```
Apache9, line 1: No compatiblity issue?

.github/workflows/lint_and_test_go-client.yml
acelyc111: Yes, so I specify the runner as ubuntu-20.04, on which the thrift-compiler is 0.13.0

go-client/generator/admin.csv
acelyc111: There are some compatiblity issues, because the structure names changed.
go-client defined it's own idl before, they have the same structure content (the fields type and sequence number) but different names, the go-client side names are more simple and clean, but consider the consistency and maintainability, I think use the unify idl and break some compatiblity would be better, the go-client is not wildly used currently.

.github/workflows/lint_and_test_go-client.yml
Apache9: `sudo apt-get install -y thrift-compiler='0.13.0-*'`

And we use thrift 0.11.0 in java client?

go-client/generator/admin.csv
Apache9: Let's at least send a NOTICE email to the mailing list first...

.github/workflows/lint_and_test_go-client.yml
acelyc111: tracked by https://github.com/apache/incubator-pegasus/issues/1283

acelyc111:
> I did intentionally name the thrift structs to be different with rDSN. Because the names in go-client are much shorter, clearer than the rDSN ones, aren't they? 🤣 But certainly, you can rollback to the naming style that is aligned with the core project. I have no opinion with that. Or if possible, I suggest that you can rename all ConfigurationXXXResponse to XXXResponse and I think it would make more senses.

I agree, really the go-client side are more tidy. I'm planning to unify all server side and all client sides in following patches at first, and then refactor to a better naming, of course, make sure there is no compatiblity issue.

go-client/generator/admin.csv
acelyc111: Will, after this patch merged.

foreverneverer (approved):

PR #1285, opened 2022-12-04
feat(new_metrics): traverse the whole registry to choose entities and metrics according to the filters

### LINKED ISSUE #1279: Feature(new_metrics): traverse the whole registry to choose entities and metrics according to the filters
In https://github.com/apache/incubator-pegasus/issues/1245, we've supported taking snapshot as json format for an entity according to the filters constructed by the client request. The next job is to traverse the whole registry, taking snapshot for each entity.

empiredan:
This PR is to resolve #1279.

### DIFF
--- a/src/utils/test/metrics_test.cpp
+++ b/src/utils/test/metrics_test.cpp
@@ -2202,18 +2202,11 @@ void check_entities_from_json_string(const std::string &json_string,
             ASSERT_TRUE(elem.name.IsString());
 
             if (kMetricEntityTypeField == elem.name.GetString()) {
-                ASSERT_STREQ(elem.value.GetString(), "my_server");
+                ASSERT_STREQ("my_server", elem.value.GetString());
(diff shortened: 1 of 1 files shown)


src/utils/test/metrics_test.cpp
```diff
@@ -2178,4 +2178,121 @@ TEST(metrics_test, take_snapshot_entity)
...
+                        if (kMetricNameField == field.name.GetString()) {
+                            ASSERT_TRUE(field.value.IsString());
+                            actual_entity_metrics.emplace(field.value.GetString());
+                        }
+                    }
+                }
+
+                static const std::unordered_set<std::string> kExpectedEntityMetrics = {
+                    "test_gauge_int64", "test_counter"};
+                ASSERT_EQ(actual_entity_metrics, kExpectedEntityMetrics);
```
acelyc111, line 2236: It's better to use it like `ASSERT_EQ(expect_value, actual_value);`

acelyc111 (approved):

PR #1290, opened 2022-12-12
refactor(idl): unify the *.thrift files used by cpp and java-client

acelyc111:

This patch including changes:
- move thrift file from `src/runtime/security/security.thrift` to `idl/security.thrift`
- remove `java-client/idl/*.thrift`
- add `struct request_meta` to *.thrift which is used by java-client only
- add java related namespace to *.thrift
- remove `java-client/idl/apache-licence-template` while it is not used now
- move `recompile_thrift.sh` from `java-client/idl` to `java-client/scripts`
- update some structures' and fields' name, there are differents of *.thrift files between `idl/*` and `java-client/idl/*` before

## Compatiblity changes

Because Java client is more wildly used than cpp-client and go-client, so I do the following refactor to make sure Java-client has less compatiblity issue:

### cpp-client

| old name | new name |
| ---- | ---- |
| configuration_query_by_index_request | query_cfg_request |
| configuration_query_by_index_response | query_cfg_response |

### go-client

| old name | new name |
| ---- | ---- |
| ConfigurationQueryByIndexRequest | QueryCfgRequest |
| ConfigurationQueryByIndexResponse | QueryCfgResponse |


foreverneverer (approved):

acelyc111:
> I think we'd beter bump the major version to include this break changes?
> 
> And do we have a compatibilty guide about what things can be broken while upgrading?
> 
> And it is a bit strange that, why java client and cpp client can use different thrift defination while they connect to the same server?

1. That would be better, but I'm afraid there may be some more compatiblity changes in the future, the major version will inscrease too frequently then.
2. We will point that out obviously in the release note, as how we do before.
3. No matter what the name of the struct or field is, they are compatiable. When decode a request message from client on server, it depends on what the rpc code is, then use the related struct to decode it. For the fields, only the numberic tags are used, no matter what the names are.

Apache9:
OK, found this on stackoverflow

https://stackoverflow.com/questions/52882370/is-safe-to-rename-a-field-in-thrift-idl

SO thrift only relies on the field id, not the name.

Then I think it is less hurt for our users to upgrade.

Apache9 (approved):

Issue #1299, opened 2022-12-19
Feature: decide whether two C strings are equal and support CHECK_STREQ* and CHECK_STRNE*


empiredan:
While asserting equality between two C strings now we have to check if `strcmp` returns 0, for example:

```c++
CHECK_EQ_MSG(std::strcmp(prototype->name(), iter->second->prototype()->name()),
             0,
             "new prototype '{}' is inconsistent with old prototype '{}' for entity '{}'",
             prototype->name(),
             iter->second->prototype()->name(),
             id);
```

We can support CHECK_STREQ* and CHECK_STRNE* in common macros for C strings.

PR #1300, opened 2022-12-19
feat: decide whether two C strings are equal and support CHECK_STREQ* and CHECK_STRNE*

empiredan:
This PR is to resolve https://github.com/apache/incubator-pegasus/issues/1299.

There are several types of comparisons for equality between two strings:

- if both strings are of `std::string`, to compare case-sensitive equality, we can use `operator==`
- if both strings are of `std::string`, to compare case-insensitive equality, we can use `boost::iequals`
- if one string is of `std::string` and another is of C string, to compare case-sensitive equality we can also use `operator==`
- if one string is of `std::string` and another is of C string, to compare case-insensitive equality we can use `strcasecmp`
- if both strings are of C string, to compare case-sensitive equality we can use `strcmp`
- if both strings are of C string, to compare case-insensitive equality we can also use `strcasecmp`
- to compare the equality of first n bytes, we can use `strncmp`, `strncasecmp`, `memcmp`

However, the problem of `strcmp`, `strcmp`, `strcasecmp` `strncmp`, `strncasecmp` and `memcmp` is that once the compared strings is NULL pointer, these functions may lead to undefined behavior. Thus in this PR we will encapsulate the following functions to solve this problem:
```c++
// Decide whether two C strings are equal, even if one of them is NULL.
// The second function is similar except it compares the only first (at most) n bytes
// of both strings.
bool equals(const char *lhs, const char *rhs);
bool equals(const char *lhs, const char *rhs, size_t n);

// Decide whether two C strings are equal, ignoring the case of the characters,
// even if one of them is NULL.
// The second function is similar except it compares the only first (at most) n bytes
// of both strings.
bool iequals(const char *lhs, const char *rhs);
bool iequals(const char *lhs, const char *rhs, size_t n);

// Decide whether two strings one of which is C string are equal, ignoring the
// case of the characters, even if the C string is NULL.
bool iequals(const std::string &lhs, const char *rhs);
bool iequals(const std::string &lhs, const char *rhs, size_t n);
bool iequals(const char *lhs, const std::string &rhs);
bool iequals(const char *lhs, const std::string &rhs, size_t n);

// Decide whether the first n bytes of two memory areas are equal, even if one of them is NULL.
bool mequals(const void *lhs, const void *rhs, size_t n);
```

This PR will also implement the following macros to help check the equality between C strings and true or false: 
- CHECK_STREQ*
- CHECK_STRNE*
- CHECK_TRUE
- CHECK_FALSE

### DIFF
--- a/src/shell/commands/detect_hotkey.cpp
+++ b/src/shell/commands/detect_hotkey.cpp
@@ -15,15 +15,16 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include "shell/commands.h"
-#include "shell/argh.h"
+#include "backup_types.h"
+#include "bulk_load_types.h"
+#include "consensus_types.h"
+#include "duplication_types.h"
 #include "meta_admin_types.h"
 #include "partition_split_types.h"
-#include "duplication_types.h"
-#include "bulk_load_types.h"
-#include "backup_types.h"
-#include "consensus_types.h"
 #include "replica_admin_types.h"
+#include "shell/argh.h"
+#include "shell/commands.h"
+#include "utils/strings.h"
 
 bool generate_hotkey_request(dsn::replication::detect_hotkey_request &req,
                              const std::string &hotkey_action,
@@ -32,9 +33,9 @@ bool generate_hotkey_request(dsn::replication::detect_hotkey_request &req,
                              int partition_index,
                              std::string &err_info)
 {
-    if (!strcasecmp(hotkey_type.c_str(), "read")) {
+    if (dsn::utils::iequals(hotkey_type.c_str(), "read")) {
         req.type = dsn::replication::hotkey_type::type::READ;
-    } else if (!strcasecmp(hotkey_type.c_str(), "write")) {
+    } else if (dsn::utils::iequals(hotkey_type.c_str(), "write")) {
         req.type = dsn::replication::hotkey_type::type::WRITE;
     } else {
         err_info = fmt::format("\"{}\" is an invalid hotkey type (should be 'read' or 'write')\n",
@@ -42,11 +43,11 @@ bool generate_hotkey_request(dsn::replication::detect_hotkey_request &req,
         return false;
     }
 
-    if (!strcasecmp(hotkey_action.c_str(), "start")) {
+    if (dsn::utils::iequals(hotkey_action.c_str(), "start")) {
         req.action = dsn::replication::detect_action::START;
-    } else if (!strcasecmp(hotkey_action.c_str(), "stop")) {
+    } else if (dsn::utils::iequals(hotkey_action.c_str(), "stop")) {
         req.action = dsn::replication::detect_action::STOP;
-    } else if (!strcasecmp(hotkey_action.c_str(), "query")) {
+    } else if (dsn::utils::iequals(hotkey_action.c_str(), "query")) {
         req.action = dsn::replication::detect_action::QUERY;
     } else {
         err_info =
--- a/src/shell/commands/detect_hotkey.cpp
+++ b/src/shell/commands/detect_hotkey.cpp
@@ -33,9 +33,9 @@ bool generate_hotkey_request(dsn::replication::detect_hotkey_request &req,
                              int partition_index,
                              std::string &err_info)
 {
-    if (dsn::utils::iequals(hotkey_type.c_str(), "read")) {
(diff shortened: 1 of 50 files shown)


src/shell/commands/global_properties.cpp
```diff
@@ -44,11 +45,11 @@ bool process_escape_all(command_executor *e, shell_context *sc, arguments args)
         fprintf(stderr, "Current escape_all: %s.\n", sc->escape_all ? "true" : "false");
         return true;
     } else if (args.argc == 2) {
-        if (strcmp(args.argv[1], "true") == 0) {
+        if (dsn::utils::equals(args.argv[1], "true")) {
```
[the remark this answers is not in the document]
empiredan: > Does is make sense if introducing `iequals` to compare strings but ignore cases?

Good advice ! I think this is necessary, I'll add it.

empiredan:
> How about to replace `memcmp` too?

OK, I think I can replace with a new function `mequals` that also process NULL pointers.

src/shell/commands/detect_hotkey.cpp
```diff
@@ -32,21 +33,21 @@ bool generate_hotkey_request(dsn::replication::detect_hotkey_request &req,
                              int partition_index,
                              std::string &err_info)
 {
-    if (!strcasecmp(hotkey_type.c_str(), "read")) {
+    if (dsn::utils::iequals(hotkey_type.c_str(), "read")) {
```
acelyc111, line 36: how about define the parameter as `string_view`? so you it's not needed to convet to C string.

src/utils/test/utils.cpp
```diff
@@ -115,6 +115,52 @@ TEST(core, check_c_string_empty)
...
+        {"abc", nullptr, false},  {"", "", true},        {"", "a", false},
+        {"", "abc", false},       {"a", "", false},      {"abc", "", false},
+        {"a", "a", true},         {"a", "A", false},     {"A", "A", true},
+        {"abc", "abc", true},     {"aBc", "abc", false}, {"abc", "ABC", false},
+        {"a", "abc", false},      {"A", "abc", false},   {"abc", "a", false},
+        {"Abc", "a", false},
+    };
+
+    for (const auto &test : tests) {
+        EXPECT_EQ(test.is_equal, dsn::utils::equals(test.lhs, test.rhs));
```
acelyc111, line 137: This style of unit test is wildly used, but it has a shortcoming, when some case failed, it's hard to find out which one failed, right?

src/utils/fmt_logging.h
```diff
@@ -107,6 +162,11 @@
             var1 < var2, _v1 < _v2, "{} vs {} {}", _v1, _v2, fmt::format(__VA_ARGS__));            \
     } while (false)
 
+#define CHECK_STREQ(var1, var2) CHECK_STREQ_MSG(var1, var2, "")
+#define CHECK_STRNE(var1, var2) CHECK_STRNE_MSG(var1, var2, "")
+#define CHECK_STRCASEEQ(var1, var2) CHECK_STRCASEEQ_MSG(var1, var2, "")
```
acelyc111, line 167: According to the macro names, it's confused to judge which one is case sensitive.
I found there is no caller currently, can you remove `CHECK_STRCASEEQ/NE`?

empiredan: Actually the naming convention has followed `ASSERT_STREQ/NE` and `ASSERT_STRCASEEQ/NE` of google test. And now `CHECK_STRCASEEQ/NE` is not used in fact, I can remove them.

src/utils/test/utils.cpp
empiredan: Yes, once a case failed it is really hard to look out it. I'll use `TEST_P` to refactor this.

src/shell/commands/detect_hotkey.cpp
empiredan: Actually `string_view` is used not only for the string of the printable characters, but also binary bytes. Once `string_view` is used as the type of the parameters, we have to reimplement comparison ignoring cases based on `string_view` where binary bytes such as '\0' should be skipped. Therefore here we just declare other functions one parameter of which is declared as `std::string`.  Comparing two `std::string`-typed strings we can use `boost::iequals`.

acelyc111 (approved):

foreverneverer (approved):

Issue #1303, opened 2023-01-05
Feature(new_metrics): retire stale metric entities that are not used by any other object


empiredan:
As has been described in https://github.com/apache/incubator-pegasus/issues/922, the framework of new metrics should check and retire the old metrics.

PR #1304, opened 2023-01-05
feat(new_metrics): retire stale metric entities that are not used by any other object

empiredan:
This PR is to resolve https://github.com/apache/incubator-pegasus/issues/1303.


src/utils/metrics.h
```diff
@@ -354,10 +392,96 @@ class metrics_http_service : public http_server_base
...
+
 class metric_registry : public utils::singleton<metric_registry>
 {
 public:
     using entity_map = std::unordered_map<std::string, metric_entity_ptr>;
+    using old_entity_map = std::unordered_map<std::string, metric_entity::old_metric_list>;
+
+    struct collected_old_entities_info
+    {
+        old_entity_map old_entities;
```
acelyc111, line 453: Could you add some comments about it?

src/utils/metrics.cpp
```diff
@@ -383,15 +547,134 @@ metric_entity_ptr metric_registry::find_or_create_entity(const metric_entity_pro
...
+    }
+
+    entities_info.num_all_entities = _entities.size();
+    return entities_info;
+}
+
+metric_registry::retired_entities_stat
+metric_registry::retire_old_metrics(const old_entity_map &old_entities)
+{
+    if (old_entities.empty()) {
```
acelyc111, line 583: If some other thread is writing `old_entities`, will it cause data race?

src/utils/metrics.cpp
```diff
@@ -177,6 +203,109 @@ void metric_entity::take_snapshot(metric_json_writer &writer, const metric_filte
...
+    // Since this entity itself is still being accessed, its reference count should be 1
+    // at least.
+    CHECK_GE(get_count(), 1);
+
+    // Once this entity did not have any metric, and had only one reference kept in the
+    // registry, this entity would be considered useless.
+    return _metrics.empty() && get_count() == 1;
+}
+
+metric_entity::collected_old_metrics_info metric_entity::collect_old_metrics() const
```
acelyc111, line 217: what dose 'old' mean?

src/utils/metrics.h
```diff
@@ -153,6 +153,36 @@ class metric_entity : public ref_counter
 public:
     using attr_map = std::unordered_map<std::string, std::string>;
     using metric_map = std::unordered_map<const metric_prototype *, metric_ptr>;
+    using old_metric_list = std::unordered_set<const metric_prototype *>;
+
+    struct collected_old_metrics_info
+    {
+        old_metric_list old_metrics;
```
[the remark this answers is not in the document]
empiredan: OK.

src/utils/metrics.h
empiredan: OK.

src/utils/metrics.cpp
empiredan: Actually the variable that `old_entities` references is a local constant in `process_old_metrics()`, which cannot be mutated and is also not shared by other threads.

src/utils/metrics.cpp
empiredan: `old` means a metric or entity is no longer used. Typically a metric or entity is referenced by a class. For example, the class `replica` may use a metric as the measurement of QPS for a replica.

However, once the replica is removed, this metric will not be used. The framework of metrics then will find this metric and mark it `old`. After a configurable retention interval, this metric will finally be removed from the repository of metrics.

src/utils/metrics.cpp
```diff
@@ -177,6 +203,109 @@ void metric_entity::take_snapshot(metric_json_writer &writer, const metric_filte
...
+metric_entity::collected_stale_metrics_info metric_entity::collect_stale_metrics() const
+{
+    collected_stale_metrics_info collected_info;
+
+    auto now = dsn_now_ms();
+
+    utils::auto_read_lock l(_lock);
+
+    for (const auto &m : _metrics) {
+        if (!m.second->is_stale()) {
```
acelyc111, line 226: I'm a bit of confused why only partial metrics of the entity are stale, aren't all of  them become stale at the same time when the entity they belong to is destroyed?

src/utils/metrics.cpp
```diff
@@ -383,6 +434,111 @@ metric_entity_ptr metric_registry::find_or_create_entity(const metric_entity_pro
...
+{
+    LOG_INFO("begin to process stale metric entities");
+
+    const auto &collected_info = collect_stale_entities();
+    const auto &retired_stat = retire_stale_entities(collected_info.stale_entities);
+
+    LOG_INFO("stat for metric entities: total={}, collected={}, retired={}, scheduled={}, "
+             "recently_scheduled={}, reemployed={}",
+             collected_info.num_all_entities,
+             collected_info.stale_entities.size(),
```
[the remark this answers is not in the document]
empiredan: OK, in comparison to `stale_entities`, `collected_entities` is more accurate.

src/utils/metrics.cpp
```diff
@@ -41,17 +48,15 @@ metric_entity::~metric_entity()
...
-    //
-    // Thus, to close all metrics owned by an entity, it's more efficient to firstly issue a close
-    // request for all metrics; then, just wait for all of the close operations to be finished.
-    // It's inefficient to wait for each metric to be closed one by one.
+    // To close all metrics owned by an entity, it's more efficient to firstly issue an asynchronous
+    // close request to each metric; then, just wait for all of the close operations to be finished.
+    // It's inefficient to wait for each metric to be closed one by one. Therefore, the metric is
+    // not
+    // closed in its destructor.
     for (auto &m : _metrics) {
```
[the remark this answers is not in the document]
empiredan: Compilation can be passed while this function is qualified with `const` since actually it does not modify the direct member (add, remove or update), such as `_metrics`; however, it will change the state of the member of `_metrics`. Therefore I think it's better to declare it without `const`.

acelyc111 (approved):

foreverneverer (approved):

Issue #1305, opened 2023-01-10
Refactor log macros


acelyc111:
There are two type of log macros in the code, one is using C string format specifiers, like `LOG_INFO`, the other is using `libfmt`, like `LOG_INFO_F`.

It's strange to keep the two macros in long term, I'm planning to refactor them and only leave `libfmt` style one.

acelyc111:
Because there are so many code to be modified, I'll separet the whole work into smaller patches to make reviewers happy 😆 
- LOG_DEBUG
  - all
- LOG_INFO
  - replica module
  - meta module
  - others
- LOG_WARN
  - replica module
  - meta module
  - others
- LOG_WARN
  - replica module
  - others
- LOG_FATAL
  - all

empiredan:
After all `LOG_*` macros have been replaced with `LOG_*_F`, would `LOG_*_F` be renamed as `LOG_*`? For example, `LOG_INFO_F` is renamed as `LOG_INFO` ?

acelyc111:
Yes, I'm planning to still use `LOG_*_F` before all refactor work been finished, then remove `_F` postfix at last in one patch.

acelyc111:
> After all `LOG_*` macros have been replaced with `LOG_*_F`, would `LOG_*_F` be renamed as `LOG_*`? For example, `LOG_INFO_F` is renamed as `LOG_INFO` ?

Yes, I'm planning to still use `LOG_*_F` before all refactor work been finished, then remove `_F` postfix at last in one patch.

acelyc111:
> > After all `LOG_*` macros have been replaced with `LOG_*_F`, would `LOG_*_F` be renamed as `LOG_*`? For example, `LOG_INFO_F` is renamed as `LOG_INFO` ?
> 
> Yes, I'm planning to still use `LOG_*_F` before all refactor work been finished, then remove `_F` postfix at last in one patch.

In this way, all macros with or without `_F` postfix are both in the same semantics in the progress.

empiredan:
> > > After all `LOG_*` macros have been replaced with `LOG_*_F`, would `LOG_*_F` be renamed as `LOG_*`? For example, `LOG_INFO_F` is renamed as `LOG_INFO` ?
> > 
> > 
> > Yes, I'm planning to still use `LOG_*_F` before all refactor work been finished, then remove `_F` postfix at last in one patch.
> 
> In this way, all macros with or without `_F` postfix are both in the same semantics in the progress.

[...]

PR #1310, opened 2023-01-13
refactor(log): use LOG_INFO_F instead of LOG_INFO (2/3)

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1305

This patch aim to use LOG_INFO_F instead of LOG_INFO Includes:
- Use `LOG_INFO_F` instead of `LOG_INFO` in meta module.


empiredan (approved):

levy5307 (approved):
LGTM

PR #1311, opened 2023-01-13
refactor(log): use LOG_INFO_F instead of LOG_INFO (3/3)

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1305

This patch aim to use LOG_INFO_F instead of LOG_INFO for all the remain parts:
- Remove the defination of `LOG_INFO`
- Use `LOG_INFO_F` instead of `LOG_INFO`
- Use `LOG_INFO_PREFIX` instead of `LOG_INFO` for class `proxy_session`, `pegasus_server_impl `
- Use upper case for commands in class `redis_parser`
- Remove some commentted `LOG_INFO` code
- Some small fixes

TODO:
1. print dsn_primary_address in log is useless, since it must be itself in whose log, e.g. in `src/common/fs_manager.cpp`, `src/replica/replica_stub.cpp`
2. try to use `LOG_*_PREFIX` for class `policy_context`, `cold_backup_context`

### DIFF
--- a/src/replica/backup/cold_backup_context.cpp
+++ b/src/replica/backup/cold_backup_context.cpp
@@ -189,22 +189,23 @@ void cold_backup_context::check_backup_on_remote()
         LPC_BACKGROUND_COLD_BACKUP,
         [this, current_chkpt_file](const dist::block_service::create_file_response &resp) {
             if (!is_ready_for_check()) {
-                LOG_INFO("%s: backup status has changed to %s, ignore checking backup on remote",
-                         name,
-                         cold_backup_status_to_string(status()));
+                LOG_INFO_F("{}: backup status has changed to {}, ignore checking backup on remote",
+                           name,
+                           cold_backup_status_to_string(status()));
                 ignore_check();
             } else if (resp.err == ERR_OK) {
                 const dist::block_service::block_file_ptr &file_handle = resp.file_handle;
                 CHECK_NOTNULL(file_handle, "");
                 if (file_handle->get_md5sum().empty() && file_handle->get_size() <= 0) {
-                    LOG_INFO("%s: check backup on remote, current_checkpoint file %s is not exist",
-                             name,
-                             current_chkpt_file.c_str());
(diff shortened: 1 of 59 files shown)


src/replica/backup/cold_backup_context.cpp
```diff
@@ -567,31 +573,31 @@ void cold_backup_context::on_upload_chkpt_dir()
     }
 
     if (checkpoint_files.size() <= 0) {
-        LOG_INFO("%s: checkpoint dir is empty, so upload is complete and just start write "
-                 "backup_metadata",
-                 name);
+        LOG_INFO_F("{}: checkpoint dir is empty, so upload is complete and just start write "
+                   "backup_metadata",
+                   name);
```
Smityz, line 578:
```suggestion
        LOG_INFO_F("{}: checkpoint dir is empty, so upload is complete and just start write", name);
```

Smityz (dismissed):

src/replica/backup/cold_backup_context.cpp
```diff
@@ -567,31 +573,29 @@ void cold_backup_context::on_upload_chkpt_dir()
     }
 
     if (checkpoint_files.size() <= 0) {
-        LOG_INFO("%s: checkpoint dir is empty, so upload is complete and just start write "
-                 "backup_metadata",
-                 name);
+        LOG_INFO_F("{}: checkpoint dir is empty, so upload is complete and just start write", name);
```
empiredan, line 576:
```suggestion
        LOG_INFO_F("{}: checkpoint dir is empty, so upload is complete and just start to write backup_metadata", name);
```

src/replica/backup/cold_backup_context.cpp
acelyc111: "backup_metadata" is not an independent parameter, but a part of the first string parameter, so it should not be removed.

empiredan (approved):

Smityz (approved):

PR #1317, opened 2023-01-17
refactor(idl): unify the *.thrift files used by cpp and python-client

acelyc111:

This patch including changes:
- add python related namespace to *.thrift
- remove `python-client/idl/*.thrift`
- remove generated python files from `python-client/pypegasus/**.py`
- add `python-client/requirement.txt`
- update some fields of `python-client/setup.py`
- generate python files before run `python-client/test.sh`

NOTE:
- this patch also uses thrift 0.13 to generate python files, which is the imported thrift package version in pypegasus
- make sure use thrfit >= 0.10 (0.13 is recommend) to generate python files, which support Python3


Smityz:
I have a question, if the idle is changed but the client is not upgraded in time, will it cause an exception？

acelyc111:
> I have a question, if the idle is changed but the client is not upgraded in time, will it cause an exception？

Usally,

> I have a question, if the idle is changed but the client is not upgraded in time, will it cause an exception？

As long as the protocol is compatiable, they will work well. AFAIK, change structure or field name without changing sequence number, add optional fields are compatiable. You can read the generated read/write functions for more information.

empiredan (approved):

Smityz (approved):

Issue #1337, opened 2023-02-02
Feature(utils): Add a method to obtain the string prefix by intercepting the string with the specified character


WHBANG:
Add a find_string_prefix method in string. h. You can specify a character to intercept the string and get the first part of the intercept.

PR #1338, opened 2023-02-02
feat(utils): add an utility function to find string prefix by separator

WHBANG:
https://github.com/apache/incubator-pegasus/issues/1337

- Unit test

### DIFF
--- a/src/utils/strings.cpp
+++ b/src/utils/strings.cpp
@@ -420,9 +420,9 @@ std::string string_md5(const char *buffer, unsigned length)
 
 std::string find_string_prefix(const std::string &input, char separator)
 {
-    std::size_t current = input.find(separator);
+    auto current = input.find(separator);
     if (current == 0 || current == std::string::npos) {
-        return "";
+        return std::string();
     }
     return input.substr(0, current);
 }
--- a/src/utils/strings.cpp
+++ b/src/utils/strings.cpp
@@ -417,5 +417,14 @@ std::string string_md5(const char *buffer, unsigned length)
 
     return result;
 }
+
+std::string find_string_prefix(const std::string &input, char separator)
+{
+    std::size_t current = input.find(separator);
+    if (current == 0 || current == std::string::npos) {
+        return "";
+    }
+    return input.substr(0, current);
+}
 } // namespace utils
 } // namespace dsn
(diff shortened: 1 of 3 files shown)


acelyc111 (dismissed):

src/utils/strings.cpp
```diff
@@ -417,5 +417,14 @@ std::string string_md5(const char *buffer, unsigned length)
 
     return result;
 }
+
+std::string find_string_prefix(const std::string &input, char separator)
+{
+    std::size_t current = input.find(separator);
```
empiredan, line 423:
```suggestion
    auto current = input.find(separator);
```

src/utils/strings.cpp
```diff
@@ -417,5 +417,14 @@ std::string string_md5(const char *buffer, unsigned length)
...
+    if (current == 0 || current == std::string::npos) {
+        return "";
```
empiredan, line 425: Initializing with c-style empty string literal may be less efficient for compilers with older version even if optimization is enabled. Use default constructor instead:
```suggestion
        return std::string();
```

src/utils/test/utils.cpp
```diff
@@ -443,6 +443,26 @@ TEST(core, dlink)
...
 }
 
+TEST(core, find_string_prefix)
+{
+    struct test_case
+    {
+        std::string input;
+        char separator;
+        std::string expected_prefix;
+    } tests[] = {{"", ' ', ""},
```
empiredan, line 453: Could add cases that the separator is just the whole source string, and the source string only include separators, such as:
```c++
{{" ", ' ', ""}},
{{"..", '.', ""}},
```

src/utils/test/utils.cpp
```diff
@@ -443,6 +443,26 @@ TEST(core, dlink)
...
+                 {"abc.def", ' ', ""},
+                 {"abc.def", '.', "abc"},
+                 {"ab.cd.ef", '.', "ab"},
+                 {"abc...def", '.', "abc"},
+                 {".abc.def", '.', ""}};
+    std::string actual_output;
+    for (const auto &test : tests) {
+        actual_output = find_string_prefix(test.input, test.separator);
```
empiredan, line 461:
```suggestion
        auto actual_output = find_string_prefix(test.input, test.separator);
```

empiredan (approved):

acelyc111 (approved):

Issue #1336, opened 2023-01-30
Feature(new_metrics): migrate replica-level metrics for write service


empiredan:
As is described in https://github.com/apache/incubator-pegasus/issues/1326,  since both `pegasus_server_write` and `pegasus_write_service` are derived from `replica_base`, their replica-level metrics can be attached directly to *replica* entity. `pegasus_write_service` is one of the members of `pegasus_server_write`.

Following metrics are the members of `pegasus_server_write` ([pegasus_server_write.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/server/pegasus_server_write.cpp)). Their variables in current source files and new types/computations in new framework are listed as below:

| Variables | Types/Computations |
| :-------: | :------------------: |
| _pfc_recent_corrupt_write_count | increase(Counter) |

<br>

Following metrics are the members of `pegasus_write_service` ([pegasus_write_service.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/server/pegasus_write_service.cpp)). Their variables in current source files and new types/computations in new framework are listed as below:

| Variables | Types/Computations |
| :-------: | :------------------: |
| _pfc_put_qps | rate(Counter) |
| _pfc_multi_put_qps | rate(Counter) |
| _pfc_remove_qps | rate(Counter) |
| _pfc_multi_remove_qps | rate(Counter) |
| _pfc_check_and_set_qps | rate(Counter) |
| _pfc_check_and_mutate_qps | rate(Counter) |
| _pfc_put_latency | Percentile |
| _pfc_multi_put_latency | Percentile |
| _pfc_remove_latency | Percentile |
| _pfc_multi_remove_latency | Percentile |
| _pfc_incr_latency | Percentile |
| _pfc_check_and_set_latency | Percentile |
| _pfc_check_and_mutate_latency | Percentile |
| _pfc_duplicate_qps | rate(Counter) |
| _pfc_dup_time_lag | Percentile |
| _pfc_dup_lagging_writes | increase(Counter) |

acelyc111:
@empiredan  It would be better to describe the relationship between old metric names and the new ones, then users know how to change their monitor tools.

empiredan:
> @empiredan It would be better to describe the relationship between old metric names and the new ones, then users know how to change their monitor tools.

OK, I'll add both old and new names into the table. Also, I'll summarize all the changes into a big table.

PR #1351, opened 2023-02-14
feat(new_metrics): migrate replica-level metrics for write service

empiredan:
This PR is to resolve https://github.com/apache/incubator-pegasus/issues/1336. Temporarily disable `detect_hotspot_test`, for details please see in https://github.com/apache/incubator-pegasus/issues/1353.

### DIFF
--- a/src/utils/metrics.h
+++ b/src/utils/metrics.h
@@ -144,7 +144,7 @@
 // Declare a metric variable in user class.
 //
 // Since a type tends to be a class template where there might be commas, use variadic arguments
-// instead of a single fixed argumen to represent a type.
+// instead of a single fixed argument to represent a type.
 #define METRIC_VAR_DECLARE(name, ...) __VA_ARGS__ _##name
 #define METRIC_VAR_DECLARE_gauge_int64(name) METRIC_VAR_DECLARE(name, dsn::gauge_ptr<int64_t>)
 #define METRIC_VAR_DECLARE_counter(name)                                                           \
@@ -606,7 +606,7 @@ const std::vector<uint64_t> kMetricLatencyConverterFromNS = {
 
 inline uint64_t convert_metric_latency_from_ns(uint64_t latency_ns, metric_unit target_unit)
 {
-    if (target_unit == metric_unit::kNanoSeconds) {
+    if (dsn_likely(target_unit == metric_unit::kNanoSeconds)) {
         // Since nanoseconds are used as the latency unit more frequently, eliminate unnecessary
         // conversion by branch prediction.
         return latency_ns;


src/server/pegasus_write_service.h
```diff
@@ -199,28 +197,29 @@ class pegasus_write_service : dsn::replication::replica_base
...
+    dsn::percentile_ptr<int64_t> _incr_latency_ns;
+    dsn::percentile_ptr<int64_t> _check_and_set_latency_ns;
+    dsn::percentile_ptr<int64_t> _check_and_mutate_latency_ns;
+
+    dsn::counter_ptr<> _dup_counter;
+    dsn::percentile_ptr<int64_t> _dup_time_lag_ms;
+    dsn::counter_ptr<> _dup_lagging_write_counter;
+
+    // Record batch size for put and remove requests.
+    uint32_t _put_batch_size;
```
acelyc111, line 221: nit: Use atomic?

src/server/pegasus_write_service.cpp
```diff
@@ -36,111 +124,36 @@ pegasus_write_service::pegasus_write_service(pegasus_server_impl *server)
       _server(server),
       _impl(new impl(server)),
       _batch_start_time(0),
-      _cu_calculator(server->_cu_calculator.get())
+      _cu_calculator(server->_cu_calculator.get()),
```
acelyc111, line 127: How about use macros to format/simplify these codes? (Maybe the member variables and metrics should have some similar naming?)

src/server/pegasus_write_service.cpp
```diff
@@ -36,111 +124,36 @@ pegasus_write_service::pegasus_write_service(pegasus_server_impl *server)
...
+      _put_counter(METRIC_put_requests.instantiate(replica_metric_entity())),
```
acelyc111, line 128: Will the counter_ptr/percentile_ptr get any error if not initialized in c'tor? It will be helpful to prevent missing initialized.

src/utils/time_utils.h
```diff
@@ -126,5 +130,29 @@ inline int64_t hh_mm_today_to_unix_sec(string_view hhmm_of_day)
...
+class chronograph
+{
+public:
+    chronograph() : chronograph(dsn_now_ns()) {}
+    chronograph(uint64_t start_time_ns) : _start_time_ns(start_time_ns) {}
+    ~chronograph() = default;
+
+    inline void reset_start_time() { _start_time_ns = dsn_now_ns(); }
+
+    template <typename T, typename = typename std::enable_if<std::is_integral<T>::value>::type>
```
acelyc111, line 142: How about set a default type `uint64_t` for `T`?

src/server/pegasus_write_service.h
```diff
@@ -199,28 +197,29 @@ class pegasus_write_service : dsn::replication::replica_base
...
-    ::dsn::perf_counter_wrapper _pfc_multi_put_latency;
-    ::dsn::perf_counter_wrapper _pfc_remove_latency;
-    ::dsn::perf_counter_wrapper _pfc_multi_remove_latency;
-    ::dsn::perf_counter_wrapper _pfc_incr_latency;
-    ::dsn::perf_counter_wrapper _pfc_check_and_set_latency;
-    ::dsn::perf_counter_wrapper _pfc_check_and_mutate_latency;
-
-    // Records all requests.
-    std::vector<::dsn::perf_counter *> _batch_qps_perfcounters;
-    std::vector<::dsn::perf_counter *> _batch_latency_perfcounters;
```
acelyc111, line 223: Have we lost some metrics?

src/server/pegasus_write_service.cpp
```diff
@@ -296,15 +307,23 @@ void pegasus_write_service::set_default_ttl(uint32_t ttl) { _impl->set_default_t
...
-    for (dsn::perf_counter *pfc : _batch_latency_perfcounters)
-        pfc->set(latency);
-
-    _batch_qps_perfcounters.clear();
-    _batch_latency_perfcounters.clear();
+#define PROCESS_WRITE_BATCH(op)                                                                    \
+    do {                                                                                           \
+        _##op##_counter->increment_by(static_cast<int64_t>(_##op##_batch_size));                   \
+        for (uint32_t i = 0; i < _##op##_batch_size; ++i) {                                        \
+            _##op##_latency_ns->set(latency_ns);                                                   \
```
acelyc111, line 314: How about encapsulate another function like `set(count, value)` for percentile metrics?

src/server/pegasus_write_service.cpp
empiredan: OK, I'll add a new macro system to simplify/standardize operations for metric variables.

src/server/pegasus_write_service.h
empiredan: There's only one thread serving all write operations on one replica; besides, previously `std::vector` is used to store pointers of perf-counters which is not thread-safe. Therefore I think `uint32_t` is enough.

src/server/pegasus_write_service.h
empiredan: Actually there are only pointers of `put` and `remove` perf-counters in the `std::vector`, which are heavy for the problem. We can just use 2 counters instead, and there will be no metric lost.

src/utils/time_utils.h
empiredan: OK, since internally we use `dsn_now_ns()` to save timestamp, I'll change the type to `uint64_t`.

src/server/pegasus_write_service.cpp
empiredan: Good idea, I'll encapsulate this interface which is actually needed by unit-tests.

src/server/pegasus_write_service.cpp
empiredan: Normally instantiations for all metric types will not lead to error. Once instantiation fails, the process will crash which means there must have been some bug. For example, the metric entity is uninitialized.

src/utils/metrics.h
```diff
@@ -553,6 +592,31 @@ enum class metric_unit
...
+METRIC_ASSERT_UNIT_LATENCY(kMilliSeconds, 2);
+METRIC_ASSERT_UNIT_LATENCY(kSeconds, 3);
+
+const std::vector<uint64_t> kMetricLatencyConverterFromNS = {
+    1, 1000, 1000 * 1000, 1000 * 1000 * 1000};
+
+inline uint64_t convert_metric_latency_from_ns(uint64_t latency_ns, metric_unit target_unit)
+{
+    if (target_unit == metric_unit::kNanoSeconds) {
+        // Since nanoseconds are used as the latency unit more frequently, eliminate unnecessary
```
acelyc111, line 610: How about add `dsn_likely`?

acelyc111 (approved):

src/utils/metrics.h
empiredan: OK.

Issue #1347, opened 2023-02-11
Docker image build failed


acelyc111:
The docker image build failed when tis patch [1] merged, see [2].
@GehaFearless Could you please help to check and fix it?

1. https://github.com/apache/incubator-pegasus/commit/4476219048eea20155c54f58f9d7512f9f1af4e7
2. https://github.com/apache/incubator-pegasus/actions/runs/4115308630/jobs/7103848560

GehaFearless:
<img width="1045" alt="image" src="https://user-images.githubusercontent.com/48315319/218980206-1615fc53-47d5-4e23-9731-267db615d852.png">



This job has been a failure for a long time. Let me try it on local.

PR #1356, opened 2023-02-16
fix: fix centos image build

GehaFearless:
issue: https://github.com/apache/incubator-pegasus/issues/1347

This pr fix  centos7 firstly. Install thrift has some trouble on centos6.


docker/pegasus-build-env/centos6/Dockerfile
```diff
@@ -76,12 +76,6 @@ ENV MAVEN_HOME=/opt/rh/rh-maven33/root/usr/
 ENV PATH=$MAVEN_HOME/bin:$JAVA_HOME/bin:$GCC_HOME/bin:$PYTHON2_HOME/bin/:$PATH
 ENV LD_LIBRARY_PATH=$JAVA_HOME/jre/lib/amd64/server:$PYTHON2_HOME/lib64/:$LD_LIBRARY_PATH
 
-RUN pip install --no-cache-dir cmake
-
-RUN wget --progress=dot:giga https://github.com/apache/thrift/archive/refs/tags/0.11.0.tar.gz -P /opt/thrift && \
```
acelyc111, line 81: thrift is used to generate cpp/java code from IDL, why removed it?

docker/pegasus-build-env/centos7/Dockerfile
```diff
@@ -54,21 +54,21 @@ RUN yum -y install centos-release-scl \
...
 
-RUN pip3 install --no-cache-dir cmake
+RUN pip3 install --upgrade pip --no-cache-dir && pip3 install --no-cache-dir cmake
 
 RUN wget --progress=dot:giga https://archive.apache.org/dist/maven/maven-3/3.8.3/binaries/apache-maven-3.8.3-bin.tar.gz -P /opt/maven \
     && cd /opt/maven \
     && tar -zxf apache-maven-3.8.3-bin.tar.gz \
     && rm apache-maven-3.8.3-bin.tar.gz
 
+ENV JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk
```
acelyc111, line 64: What issue to solve by reordering?

docker/pegasus-build-env/centos6/Dockerfile
acelyc111: This patch added these code: https://github.com/apache/incubator-pegasus/pull/1045

That's because cpp code link thrift in version 0.9.3 (build by thirdparty), java code use thrift 0.11.0 binary (download here) to generate code from *.thrift files. Although it's confused, we can improve the different version lib/bin in another patch, but it would better to keep the logic in this patch.

acelyc111 (approved):

Issue #1054, opened 2022-07-18
Feature: Integrate with Apache Ranger


acelyc111:
Apache Ranger™ [1] is a framework to enable, monitor and manage comprehensive data security across the Hadoop platform. There are many big data components support to integate with Ranger, like HDFS，HBase，Hive，Yarn，Kafka，Kudu.

Now Pegasus supports Kerberos and built-in ACL, but it's a bit of difficult to manage it, we can make Pegasus interact with Ranger to make it easier for management.

1. https://ranger.apache.org/

acelyc111:
It is also tracked in Ranger community https://issues.apache.org/jira/browse/RANGER-3831

kirbyzhou:
If I understand it correctly, the ACL model is so simple now. There are two access_controller class.  

- meta_access_controller:

Super User is allowed to do anything. 
All other users are allowed to do the things listed in FLAGS_meta_acl_rpc_allow_list. There is no per-user settings.
Default meta_acl_rpc_allow_list are 
```
RPC_CM_LIST_APPS 
RPC_CM_LIST_NODES 
RPC_CM_CLUSTER_INFO 
RPC_CM_QUERY_PARTITION_CONFIG_BY_INDEX
```

- replica_access_controller： 

Super User is allowed to do anything. 
Users in the users list are allowed to do anything. Users are set by 'replica_access_controller.allowed_users' vars in replica::update_ac_allowed_users
There seems no per-user settings too.


It seems we first have to enhance the access controller mech, add per-user / per-table support.
Such as :
{ Table1: { user1: read, user2: read+write }, Table2: { user3: read, user4: read+write } }

kirbyzhou:
Add a draft of Ranger Service definition
[here](https://issues.apache.org/jira/secure/attachment/13047331/ranger-servicedef-pegasus.json)

acelyc111:
The current ACL is described in https://github.com/apache/incubator-pegasus/issues/170 and [Pegasus 安全认证](https://mp.weixin.qq.com/s?__biz=MzkzNzAzNDgyNg==&mid=2247483801&idx=1&sn=0516fcdfb3667b85d03eb5a40293e6de&chksm=c294d4bcf5e35daa67a4eb3317749c7a643e09a54c5c886be407a1105d96c6607c416cb4310c&token=1577740624&lang=en_US#rd)

It can be summarized as following:


operation \ user | super user | table owner | other users
-- | -- | -- | --
[...]

PR #1360, opened 2023-02-22
feat(security): Use Apache Ranger for access control(1/n)

WHBANG:

This patch aim to define ranger policy:
- Define the ranger policy data structure.
- Realized the judgment logic function when performing ACL with different priorities.
- Add some unit test for test

### DIFF
--- a/src/runtime/ranger/ranger_resource_policy.cpp
+++ b/src/runtime/ranger/ranger_resource_policy.cpp
@@ -1,0 +1,79 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include "ranger_resource_policy.h"
+
+namespace dsn {
+namespace ranger {
+
+bool policy_item::match(const access_type &ac_type, const std::string &user_name) const
+{
+    return (access_types & ac_type) != 0 && users.count(user_name) != 0;
+}
+
+bool acl_policies::allowed(const access_type &ac_type, const std::string &user_name) const
+{
+    // 1. Check if it is not allowed.
+    for (const auto &deny_policy : deny_policies) {
+        // 1.1. In 'deny_policies'.
+        if (!deny_policy.match(ac_type, user_name)) {
+            continue;
+        }
+        bool in_deny_policies_exclude = false;
+        for (const auto &deny_policy_exclude : deny_policies_exclude) {
+            if (deny_policy_exclude.match(ac_type, user_name)) {
+                in_deny_policies_exclude = true;
+                break;
+            }
+        }
+        // 1.2. Not in any 'deny_policies_exclude', it's not allowed.
+        if (!in_deny_policies_exclude) {
+            return false;
+        }
+    }
+
+    // 2. Check if it is allowed.
+    for (const auto &allow_policy : allow_policies) {
+        // 2.1. In 'allow_policies'.
+        if (!allow_policy.match(ac_type, user_name)) {
+            continue;
+        }
+        for (const auto &allow_policy_exclude : allow_policies_exclude) {
+            // 2.2. In some 'allow_policies_exclude', it's not allowed.
+            if (allow_policy_exclude.match(ac_type, user_name)) {
+                return false;
+            }
+        }
+        // 2.3. Not in any 'allow_policies_exclude', it's allowed.
+        return true;
+    }
+
+    // 3. Otherwise, it's not allowed.
+    return false;
+}
+
+void ranger_resource_policy::create_default_database_policy(ranger_resource_policy &acl)
+{
+    acl.name = "default database policy";
+    acl.database_names = {"*"};
+    policy_item item;
+    item.access_types = CREATE | DROP | LIST | METADATA | CONTROL;
+    acl.policies.allow_policies.emplace_back(item);
+}
+
+} // namespace ranger
+} // namespace dsn
--- a/src/runtime/ranger/ranger_resource_policy.cpp
+++ b/src/runtime/ranger/ranger_resource_policy.cpp
@@ -66,14 +66,5 @@ bool acl_policies::allowed(const access_type &ac_type, const std::string &user_n
     return false;
 }
 
-void ranger_resource_policy::create_default_database_policy(ranger_resource_policy &acl)
-{
-    acl.name = "default database policy";
-    acl.database_names = {"*"};
-    policy_item item;
-    item.access_types = CREATE | DROP | LIST | METADATA | CONTROL;
-    acl.policies.allow_policies.emplace_back(item);
-}
-
 } // namespace ranger
 } // namespace dsn
--- a/src/runtime/ranger/ranger_resource_policy.cpp
+++ b/src/runtime/ranger/ranger_resource_policy.cpp
@@ -20,9 +20,21 @@
 namespace dsn {
 namespace ranger {
 
+/*extern*/ access_type operator|(access_type lhs, access_type rhs)
+{
+    using act = std::underlying_type<access_type>::type;
+    return access_type(static_cast<act>(lhs) | static_cast<act>(rhs));
+}
+
+/*extern*/ access_type operator&(access_type lhs, access_type rhs)
+{
+    using act = std::underlying_type<access_type>::type;
+    return access_type(static_cast<act>(lhs) & static_cast<act>(rhs));
+}
+
 bool policy_item::match(const access_type &ac_type, const std::string &user_name) const
 {
-    return (access_types & ac_type) != 0 && users.count(user_name) != 0;
+    return static_cast<bool>(access_types & ac_type) && users.count(user_name) != 0;
 }
 
 bool acl_policies::allowed(const access_type &ac_type, const std::string &user_name) const
--- a/src/runtime/ranger/ranger_resource_policy.h
+++ b/src/runtime/ranger/ranger_resource_policy.h
@@ -31,7 +31,7 @@ namespace dsn {
 namespace ranger {
 
 // ACL type defined in Range service for RPC matching policy
-enum class access_type : int8_t
+enum class access_type : uint8_t
 {
     KRead = 1,
     KWrite = 1 << 1,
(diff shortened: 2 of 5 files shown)


src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,109 @@
...
+
+#include <rapidjson/document.h>
+
+#include "common/json_helper.h"
+#include "utils/fmt_logging.h"
+
+namespace dsn {
+namespace ranger {
+
+enum access_type
```
acelyc111, line 33: In header files, it's necessary to add required comments to describe enums, structs/classes, member functions and variables of structs/classes.

src/runtime/test/ranger_resource_policy_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+        bool expected_result;
+    } tests[] = {{access_type::INVALID, "", false},
+                 {access_type::INVALID, "user", false},
+                 {access_type::READ, "user", false},
+                 {access_type::LIST, "user1", false},
+                 {access_type::READ, "user1", true},
+                 {access_type::WRITE, "user2", true}};
+    for (const auto &test : tests) {
+        auto actual_result = item.match(test.ac_type, test.user_name);
+        EXPECT_EQ(actual_result, test.expected_result);
```
acelyc111, line 42: it would be better to swap the order of "expect" value and "actual" value, i.e. `EXPECT_EQ(test.expected_result, actual_result);`

src/runtime/test/ranger_resource_policy_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+                 {access_type::READ, "user3", false},
+                 {access_type::CREATE, "user3", true},
+                 {access_type::LIST, "user3", false},
+                 {access_type::READ, "user4", true},
+                 {access_type::WRITE, "user4", false},
+                 {access_type::CREATE, "user4", true},
+                 {access_type::LIST, "user4", false}};
+    for (const auto &test : tests) {
+        auto actual_result = policy.allowed(test.ac_type, test.user_name);
+        EXPECT_EQ(actual_result, test.expected_result);
```
acelyc111, line 77: same

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,109 @@
...
+ENUM_END(access_type)
+
+ENUM_TYPE_SERIALIZATION(access_type, INVALID)
+
+struct policy_item
+{
+    std::set<access_type> access_types;
+    std::set<std::string> users;
+    std::set<std::string> groups; // not use
+    std::set<std::string> roles;  // not use
```
acelyc111, line 64: If they are no used now, can we remove them?

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,109 @@
...
+public:
+    ranger_resource_policy() = default;
+    ~ranger_resource_policy() = default;
+
+    // Create the default policy to adapt legacy tables.
+    static void create_default_database_policy(ranger_resource_policy &acl);
+
+public:
+    std::string name;
+    std::set<std::string> global_names; // TODO(yingchun): not in use
```
acelyc111, line 100: can we remove it?

src/runtime/test/ranger_resource_policy_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        access_type ac_type;
+        std::string user_name;
+        bool expected_result;
+    } tests[] = {{access_type::INVALID, "", false},
+                 {access_type::INVALID, "user", false},
+                 {access_type::READ, "user", false},
+                 {access_type::LIST, "user1", false},
+                 {access_type::READ, "user1", true},
+                 {access_type::WRITE, "user2", true}};
```
acelyc111, line 39: How about check all access types for user1 and user2?

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,112 @@
...
+};
+
+// A policy data structure definition of ranger resources
+struct ranger_resource_policy
+{
+public:
+    ranger_resource_policy() = default;
+    ~ranger_resource_policy() = default;
+
+    // Create the default policy to adapt legacy tables.
```
acelyc111, line 98:
```suggestion
    // Create the default policy to adapt legacy tables.
    // In Ranger policy, database prefix is required to identify the database it belongs to, but for the tables which are created before Ranger policy is introduced to Pegasus, they don't have such a prefix. So we create a default database policy for them.
```

src/runtime/ranger/ranger_resource_policy.cpp
```diff
@@ -0,0 +1,79 @@
...
+    // 3. Otherwise, it's not allowed.
+    return false;
+}
+
+void ranger_resource_policy::create_default_database_policy(ranger_resource_policy &acl)
+{
+    acl.name = "default database policy";
+    acl.database_names = {"*"};
+    policy_item item;
+    item.access_types.insert(access_type::ALL);
```
acelyc111, line 74: `users` field of policy_item is left as empty, is it what we expect?

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,112 @@
...
+ENUM_REG(CONTROL)
+ENUM_REG(ALL)
+ENUM_END(access_type)
+
+ENUM_TYPE_SERIALIZATION(access_type, INVALID)
+
+// Ranger policy data structure
+struct policy_item
+{
+    std::set<access_type> access_types;
```
acelyc111, line 63: Use bitmap instead of std::set, `match` is a very frequently called function in hot path, we must ensure it's performance.
It's recommend to add a simple benchmark test for `match`.

src/runtime/ranger/ranger_resource_policy.h
acelyc111: I meant remove the whole line.

src/runtime/ranger/ranger_resource_policy.cpp
WHBANG: Yes, `users` of database resource policy is empty

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,102 @@
...
+{
+    INVALID = 0,
+    READ = 1,
+    WRITE = 1 << 1,
+    CREATE = 1 << 2,
+    DROP = 1 << 3,
+    LIST = 1 << 4,
+    METADATA = 1 << 5,
+    CONTROL = 1 << 6,
+    ALL = 1 << 7,
```
acelyc111, line 44: When will `INVALID` and `ALL` be used? Are they necessary? Is it correct to set all bits for `ALL` but not only 1 bit?

src/runtime/ranger/ranger_resource_policy.cpp
```diff
@@ -0,0 +1,79 @@
...
+    // 3. Otherwise, it's not allowed.
+    return false;
+}
+
+void ranger_resource_policy::create_default_database_policy(ranger_resource_policy &acl)
+{
+    acl.name = "default database policy";
+    acl.database_names = {"*"};
+    policy_item item;
+    item.access_types = 127;
```
acelyc111, line 74: How about use `READ | WRITE | CREATE | ...` instead of `127`, it could improve the readability.

src/runtime/test/ranger_resource_policy_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <gtest/gtest.h>
+
+#include "runtime/ranger/ranger_resource_policy.h"
+
+namespace dsn {
+namespace ranger {
+
+TEST(ranger_resource_policy_test, policy_item_match)
+{
+    policy_item item = {7, {"user1", "user2"}};
```
acelyc111, line 27: Avoid to use magic number.

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,102 @@
...
+};
+
+// Ranger policy data structure
+struct policy_item
+{
+    int8_t access_types;
+    std::set<std::string> users;
```
acelyc111, line 51: How many user typically in a policy? Would it better to use unordered_set?

src/runtime/ranger/ranger_resource_policy.cpp
```diff
@@ -0,0 +1,79 @@
...
+    // 3. Otherwise, it's not allowed.
+    return false;
+}
+
+void ranger_resource_policy::create_default_database_policy(ranger_resource_policy &acl)
+{
+    acl.name = "default database policy";
+    acl.database_names = {"*"};
+    policy_item item;
+    item.access_types = CREATE | DROP | LIST | METADATA | CONTROL;
```
acelyc111, line 74: miss READ and WRITE?

acelyc111 (approved):
LGTM, thanks to your contribution!

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+// A policy data structure definition of ranger resources
+struct ranger_resource_policy
+{
+public:
+    ranger_resource_policy() = default;
+    ~ranger_resource_policy() = default;
+
+public:
```
empiredan, line 78: `public` could be omitted.

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,86 @@
...
+
+    // Check whether the 'user_name' is allowed to access the resource by type of 'ac_type'.
+    bool allowed(const access_type &ac_type, const std::string &user_name) const;
+};
+
+// A policy data structure definition of ranger resources
+struct ranger_resource_policy
+{
+public:
+    ranger_resource_policy() = default;
```
empiredan, line 75: Is it necessary to add both default constructor and destructor here ?

src/runtime/ranger/ranger_resource_policy.cpp
```diff
@@ -0,0 +1,70 @@
...
+// under the License.
+
+#include "ranger_resource_policy.h"
+
+namespace dsn {
+namespace ranger {
+
+bool policy_item::match(const access_type &ac_type, const std::string &user_name) const
+{
+    return (access_types & ac_type) != 0 && users.count(user_name) != 0;
```
empiredan, line 25: Is `ac_type` allowed to just include single type or multiple types ? Are following examples matched ?
```
access_types = READ | DROP, ac_type = READ | CREATE
access_types = WRITE, ac_type = READ | WRITE | CREATE
```

WHBANG: One rpc request can only match one `ac_type`, so `access_types = READ | DROP` will not occur.

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -0,0 +1,85 @@
...
+#include <rapidjson/document.h>
+
+#include "common/json_helper.h"
+#include "utils/fmt_logging.h"
+
+namespace dsn {
+namespace ranger {
+
+// ACL type defined in Range service for RPC matching policy
+enum class access_type : int8_t
```
empiredan, line 34:
```suggestion
enum class access_type : uint8_t
```

empiredan (approved):

acelyc111 (approved):

Issue #1323, opened 2023-01-28
Unify configuration launch method


acelyc111:
There are two types of configuration launch methods in the code.
One is using gflag like macros, e.g.
```
#define DSN_DEFINE_uint64(section, name, val, desc)
```
The other is using functions, e.g.
```
uint64_t dsn_config_get_value_uint64(const char *section,
                                     const char *key,
                                     uint64_t default_value,
                                     const char *dsptr);
```

It's strange to keep the two methods in long term, We're planning to refactor them and only leave `DSN_DEFINE_*` ones.

To make the patches easy to review, I want to split the whole work into small patches, each patch only refactor one type of configuration.
- DSN_DEFINE_int32:
  - https://github.com/apache/incubator-pegasus/pull/1324
  - https://github.com/apache/incubator-pegasus/pull/1346
- DSN_DEFINE_uint32
  - https://github.com/apache/incubator-pegasus/pull/1352
- DSN_DEFINE_int64
  - https://github.com/apache/incubator-pegasus/pull/1357
- DSN_DEFINE_uint64
  - https://github.com/apache/incubator-pegasus/pull/1359
- DSN_DEFINE_double
  - https://github.com/apache/incubator-pegasus/pull/1362
- DSN_DEFINE_bool
  - https://github.com/apache/incubator-pegasus/pull/1363
- DSN_DEFINE_string
  - https://github.com/apache/incubator-pegasus/pull/1371

acelyc111:
There are some configs are launched by macros `CONFIG_BEGIN`, `CONFIG_FLD*` and `CONFIG_END`, we can refactor them in another issuse, this one closed.

PR #1362, opened 2023-02-24
refactor(conf): use DSN_DEFINE_double to load double type of configs

acelyc111:

This patch refactors the code to use `DSN_DEFINE_double` instead of `dsn_config_get_value_double` to load double type of configurations, and doesn't introduce any functional changes.

### DIFF
--- a/src/runtime/service_api_c.cpp
+++ b/src/runtime/service_api_c.cpp
@@ -50,6 +50,15 @@
 #include "utils/time_utils.h"
 #include "utils/process_utils.h"
 
+#ifdef DSN_ENABLE_GPERF
+DSN_DEFINE_double(core,
+                  tcmalloc_release_rate,
+                  1.,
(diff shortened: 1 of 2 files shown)


src/runtime/service_api_c.cpp
```diff
@@ -50,6 +50,15 @@
 #include "utils/time_utils.h"
 #include "utils/process_utils.h"
 
+#ifdef DSN_ENABLE_GPERF
+DSN_DEFINE_double(core,
+                  tcmalloc_release_rate,
+                  1.,
+                  "the memory releasing rate of tcmalloc, default "
+                  "is 1.0 in gperftools, value range is "
+                  "[0.0~10.0]");
```
empiredan, line 59:
```suggestion
                  "[0.0, 10.0]");
```

empiredan (approved):

Smityz (approved):

PR #1363, opened 2023-02-24
refactor(conf): use DSN_DEFINE_bool to load bool type of configs

acelyc111:

This patch refactors the code to use `DSN_DEFINE_bool` instead of `dsn_config_get_value_bool` to load bool type of configurations, and doesn't introduce any functional changes.
- move the defination of flags closer to the places where they're used

### DIFF
--- a/src/replica/replica_stub.cpp
+++ b/src/replica/replica_stub.cpp
@@ -70,7 +70,41 @@ DSN_DEFINE_bool(replication,
                 ignore_broken_disk,
                 true,
                 "true means ignore broken data disk when initialize");
-
+DSN_DEFINE_bool(replication,
+                deny_client_on_start,
+                false,
+                "whether to deny client read and write requests when starting the server, default "
+                "is false");
+DSN_DEFINE_bool(replication,
+                verbose_client_log_on_start,
+                false,
+                "whether to print verbose error log when reply to client read and write requests "
+                "when starting the server, default is false");
+DSN_DEFINE_bool(replication,
+                mem_release_enabled,
+                true,
+                "whether to enable periodic memory release");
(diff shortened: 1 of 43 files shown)


src/utils/simple_logger.cpp
```diff
@@ -40,7 +40,6 @@ namespace dsn {
 namespace tools {
 
 DSN_DEFINE_bool(tools.simple_logger, fast_flush, false, "whether to flush immediately");
-
 DSN_DEFINE_bool(tools.simple_logger,
                 short_header,
```
empiredan, line 44: Add another `short_header` for `tools.screen_logger` ?

src/replica/replica_stub.cpp
```diff
@@ -719,7 +752,7 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
     }
 
     // gc
-    if (false == _options.gc_disabled) {
+    if (false == FLAGS_gc_disabled) {
```
empiredan, line 755:
```suggestion
    if (!FLAGS_gc_disabled) {
```

src/utils/simple_logger.cpp
acelyc111: It seems `screen_logger(const char *log_dir)` is only used in one test, and the parameter `log_dir` is useless, so I removed the c'stor and updated the only caller.

ps: Two sections with one same key is not implemented yet.

empiredan: OK, that means `[tools.screen_logger].short_header` is not used anywhere.

empiredan (approved):

levy5307 (approved):

Issue #1368, opened 2023-02-27
The script tool supports specifying configuration file when using shell client


WHBANG:
Many script tools use the shell client through "--cluster meta_server_list".  we hope to add "--config config_path" parameter to specify the configuration file. as:
`./run.sh shell --config shell-config-path`

PR #1370, opened 2023-02-27
chore(script): Script migrate_node.sh supports using specified config file when using shell config

WHBANG:
https://github.com/apache/incubator-pegasus/issues/1368

- Manual test
image

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1729,7 +1729,7 @@ function run_migrate_node()
     echo
     cd ${ROOT}
     echo "------------------------------"
-    if [ ${CONFIG_SPECIFIED} -eq 0 ]; then
+    if [ "$CLUSTER" != "" ]; then
         ./scripts/migrate_node.sh $CLUSTER $NODE "$APP" $TYPE
     else
         ./scripts/migrate_node.sh $CONFIG $NODE "$APP" $TYPE -f
--- a/run.sh
+++ b/run.sh
@@ -1647,6 +1647,7 @@ function usage_migrate_node()
     echo "Options for subcommand 'migrate_node':"
     echo "   -h|--help            print the help info"
     echo "   -c|--cluster <str>   cluster meta lists"
+    echo "   -f|--config <str>    shell config path"
     echo "   -n|--node <str>      the node to migrate primary replicas out, should be ip:port"
     echo "   -a|--app <str>       the app to migrate primary replicas out, if not set, means migrate all apps"
     echo "   -t|--type <str>      type: test or run, default is test"
@@ -1655,6 +1656,7 @@ function usage_migrate_node()
 function run_migrate_node()
 {
     CLUSTER=""
+    CONFIG=""
     NODE=""
     APP="*"
     TYPE="test"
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -1660,6 +1663,11 @@ function run_migrate_node()
                 CLUSTER="$2"
                 shift
                 ;;
+            -p|--config)
+                CONFIG="$2"
+                CONFIG_SPECIFIED=1
```
acelyc111, line 1668: `CONFIG_SPECIFIED` can be removed, just judge whether `CONFIG` is empty or not?

run.sh
```diff
@@ -1703,14 +1711,22 @@ function run_migrate_node()
...
     echo "APP=$APP"
     echo "TYPE=$TYPE"
     echo
     cd ${ROOT}
     echo "------------------------------"
-    ./scripts/migrate_node.sh $CLUSTER $NODE "$APP" $TYPE
+    if [ ${CONFIG_SPECIFIED} -eq 0 ]; then
+        ./scripts/migrate_node.sh $CLUSTER $NODE "$APP" $TYPE
+    else
+       ./scripts/migrate_node.sh $CLUSTER $NODE "$APP" $TYPE $CONFIG_SPECIFIED
```
acelyc111, line 1728:
```suggestion
        ./scripts/migrate_node.sh $CLUSTER $NODE "$APP" $TYPE $CONFIG_SPECIFIED
```

acelyc111, line 1728: Should use `$CONFIG` instead of `$CLUSTER`?

scripts/migrate_node.sh
```diff
@@ -20,37 +20,52 @@ set -e
...
+  node=$2
+  app_name=$3
+  type=$4
+else
+  CONFIG_SPECIFIED=1
+  config=$1
+  node=$2
+  app_name=$3
+  type=$4
+fi
```
acelyc111, line 48:
```suggestion
if [ $# -eq 4 ]; then
  CONFIG_SPECIFIED=0
  cluster=$1
else
  CONFIG_SPECIFIED=1
  config=$1
fi
node=$2
app_name=$3
type=$4
```

empiredan:
If it's necessary to be compatible with the previous format of arguments for `migrate_node.sh`, we could add an option `-t` for `migrate_node.sh`:

- `migrate_node.sh -t config <arguments ...>` means the first argument is config, or
- `migrate_node.sh -t cluster <arguments ...>` or just `migrate_node.sh <arguments ...>` (`-t` is omitted) means the first argument is cluster.

empiredan (approved):

scripts/migrate_node.sh
```diff
@@ -20,37 +20,58 @@ set -e
...
 fi
 
 echo "UID=$UID"
 echo "PID=$PID"
 echo
 
-echo "set_meta_level steady" | ./run.sh shell --cluster $cluster &>/tmp/$UID.$PID.pegasus.set_meta_level
-
-echo ls | ./run.sh shell --cluster $cluster &>/tmp/$UID.$PID.pegasus.ls
+if [ ${CONFIG_SPECIFIED} -eq 0 ]; then
```
acelyc111, line 68: `CONFIG_SPECIFIED` can be replaced by `cluster`?

acelyc111 (approved):

empiredan (approved):

PR #1371, opened 2023-02-27
refactor(conf): use DSN_DEFINE_string to load string type of configs

acelyc111:

This patch refactors the code to use `DSN_DEFINE_string` instead of `dsn_config_get_value_string` to load string type of configurations, and doesn't introduce any functional changes.
- move the defination of flags closer to the places where they're used

### DIFF
--- a/src/zookeeper/zookeeper_session_mgr.cpp
+++ b/src/zookeeper/zookeeper_session_mgr.cpp
@@ -38,16 +38,16 @@
 #include <stdio.h>
 #include <zookeeper/zookeeper.h>
 #include <stdexcept>
+#include "utils/flags.h"
 
 namespace dsn {
 namespace dist {
 
+DSN_DEFINE_string(zookeeper, logfile, "", "The Zookeeper logfile");
+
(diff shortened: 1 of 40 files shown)


src/zookeeper/zookeeper_session_mgr.cpp
```diff
@@ -38,16 +38,16 @@
...
 
+DSN_DEFINE_string(zookeeper, logfile, "", "The Zookeeper logfile");
+
 zookeeper_session_mgr::zookeeper_session_mgr()
 {
-    _zoo_hosts = dsn_config_get_value_string("zookeeper", "hosts_list", "", "zookeeper_hosts");
-    _zoo_logfile = dsn_config_get_value_string("zookeeper", "logfile", "", "zookeeper logfile");
-
-    FILE *fp = fopen(_zoo_logfile.c_str(), "a");
+    FILE *fp = fopen(FLAGS_logfile, "a");
```
Smityz, line 50: what's the meaning of "a"?

empiredan (approved):

levy5307 (approved):

PR #1378, opened 2023-03-01
feat(Ranger): Support encode and decode Ranger policies

WHBANG:

This patch is to prepare for parse policies and dump policies:
- 'DEFINE_JSON_SERIALIZATION' for data structure.
- Preparations for json parsing
- add unit test for 'parse_policies_from_json'

### DIFF
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -347,8 +347,8 @@ inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
 {
     dverify(in.IsUint64());
     auto ans = in.GetUint64();
-    auto act_min = static_cast<uint64_t>(dsn::ranger::kAccessTypeMin);
-    auto act_max = static_cast<uint64_t>(dsn::ranger::kAccessTypeMax);
+    auto act_min = static_cast<uint64_t>(dsn::ranger::kAccessTypeNone);
+    auto act_max = static_cast<uint64_t>(dsn::ranger::kAccessTypeAll);
     dverify(ans >= act_min && ans <= act_max);
     t = static_cast<dsn::ranger::access_type>(ans);
     return true;
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -25,14 +25,15 @@
  */
 #pragma once
 
-#include <vector>
+#include <cctype>
 #include <map>
-#include <unordered_map>
 #include <set>
 #include <sstream>
 #include <string>
 #include <type_traits>
-#include <cctype>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
 
 #include <rapidjson/ostreamwrapper.h>
 #include <rapidjson/prettywriter.h>
@@ -57,6 +58,7 @@
 #include "consensus_types.h"
 #include "replica_admin_types.h"
 #include "common/replication_enums.h"
+#include "runtime/ranger/access_type.h"
 
 #define JSON_ENCODE_ENTRY(out, prefix, T)                                                          \
     out.Key(#T);                                                                                   \
@@ -335,6 +337,28 @@ INT_TYPE_SERIALIZATION(int16_t)
 INT_TYPE_SERIALIZATION(int32_t)
 INT_TYPE_SERIALIZATION(int64_t)
 
+template <typename Writer>
+inline void json_encode(Writer &out, dsn::ranger::access_type t)
+{
+    out.Uint64(static_cast<uint64_t>(t));
+}
+
+inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
+{
+    dverify(in.IsUint64());
+    int64_t ans = in.GetUint64();
+    int64_t act_min = static_cast<uint64_t>(dsn::ranger::access_type::KInvalid);
+    int64_t act_max = static_cast<uint64_t>(
+        dsn::ranger::access_type::KRead | dsn::ranger::access_type::KWrite |
+        dsn::ranger::access_type::KCreate | dsn::ranger::access_type::KDrop |
+        dsn::ranger::access_type::KList | dsn::ranger::access_type::KMetadata |
+        dsn::ranger::access_type::KControl);
+    dverify(ans >= act_min && ans <= act_max);
+    using act = std::underlying_type<dsn::ranger::access_type>::type;
+    t = (dsn::ranger::access_type) static_cast<act>(ans);
+    return true;
+}
+
 // json serialization for uint types
 #define UINT_TYPE_SERIALIZATION(TName)                                                             \
     template <typename Writer>                                                                     \
@@ -452,6 +476,19 @@ inline bool json_decode_map(const JsonObject &in, TMap &t)
     return true;
 }
 
+template <typename TSet>
+inline bool json_decode_set(const JsonObject &in, TSet &t)
+{
+    dverify(in.IsArray());
+    t.clear();
+    for (rapidjson::Value::ConstValueIterator it = in.Begin(); it != in.End(); ++it) {
+        typename TSet::value_type value;
+        dverify(json_forwarder<decltype(value)>::decode(*it, value));
+        dverify(t.emplace(std::move(value)).second);
+    }
+    return true;
+}
+
 template <typename T>
 inline void json_encode(JsonWriter &out, const std::vector<T> &t)
 {
@@ -482,15 +519,19 @@ inline void json_encode(JsonWriter &out, const std::set<T> &t)
 template <typename T>
 inline bool json_decode(const JsonObject &in, std::set<T> &t)
 {
-    dverify(in.IsArray());
-    t.clear();
+    return json_decode_set(in, t);
+}
 
-    for (rapidjson::Value::ConstValueIterator it = in.Begin(); it != in.End(); ++it) {
-        T value;
-        dverify(json_forwarder<T>::decode(*it, value));
-        dverify(t.emplace(std::move(value)).second);
-    }
-    return true;
+template <typename T>
+inline void json_encode(JsonWriter &out, const std::unordered_set<T> &t)
+{
+    json_encode_iterable(out, t);
+}
+
+template <typename T>
+inline bool json_decode(const JsonObject &in, std::unordered_set<T> &t)
+{
+    return json_decode_set(in, t);
 }
 
 template <typename T1, typename T2>
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -347,15 +347,10 @@ inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
 {
     dverify(in.IsUint64());
     int64_t ans = in.GetUint64();
-    int64_t act_min = static_cast<uint64_t>(dsn::ranger::access_type::KInvalid);
-    int64_t act_max = static_cast<uint64_t>(
-        dsn::ranger::access_type::KRead | dsn::ranger::access_type::KWrite |
-        dsn::ranger::access_type::KCreate | dsn::ranger::access_type::KDrop |
-        dsn::ranger::access_type::KList | dsn::ranger::access_type::KMetadata |
-        dsn::ranger::access_type::KControl);
+    int64_t act_min = static_cast<uint64_t>(dsn::ranger::access_type_min);
+    int64_t act_max = static_cast<uint64_t>(dsn::ranger::access_type_max);
     dverify(ans >= act_min && ans <= act_max);
-    using act = std::underlying_type<dsn::ranger::access_type>::type;
-    t = (dsn::ranger::access_type) static_cast<act>(ans);
+    t = static_cast<dsn::ranger::access_type>(ans);
     return true;
 }
 
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -347,9 +347,9 @@ inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
 {
     dverify(in.IsUint64());
     auto ans = in.GetUint64();
-    auto act_min = static_cast<uint64_t>(dsn::ranger::kAccessTypeNone);
-    auto act_max = static_cast<uint64_t>(dsn::ranger::kAccessTypeAll);
-    dverify(ans >= act_min && ans <= act_max);
+    auto act_none = static_cast<uint64_t>(dsn::ranger::kAccessTypeNone);
+    auto act_all = static_cast<uint64_t>(dsn::ranger::kAccessTypeAll);
+    dverify(ans >= act_none && ans <= act_all);
     t = static_cast<dsn::ranger::access_type>(ans);
     return true;
 }
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -346,9 +346,9 @@ inline void json_encode(Writer &out, dsn::ranger::access_type t)
 inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
 {
     dverify(in.IsUint64());
-    int64_t ans = in.GetUint64();
-    int64_t act_min = static_cast<uint64_t>(dsn::ranger::access_type_min);
-    int64_t act_max = static_cast<uint64_t>(dsn::ranger::access_type_max);
+    auto ans = in.GetUint64();
+    auto act_min = static_cast<uint64_t>(dsn::ranger::access_type_min);
+    auto act_max = static_cast<uint64_t>(dsn::ranger::access_type_max);
     dverify(ans >= act_min && ans <= act_max);
     t = static_cast<dsn::ranger::access_type>(ans);
     return true;
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -347,8 +347,8 @@ inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
 {
     dverify(in.IsUint64());
     auto ans = in.GetUint64();
-    auto act_min = static_cast<uint64_t>(dsn::ranger::access_type_min);
-    auto act_max = static_cast<uint64_t>(dsn::ranger::access_type_max);
+    auto act_min = static_cast<uint64_t>(dsn::ranger::kAccessTypeMin);
+    auto act_max = static_cast<uint64_t>(dsn::ranger::kAccessTypeMax);
     dverify(ans >= act_min && ans <= act_max);
     t = static_cast<dsn::ranger::access_type>(ans);
     return true;
--- a/src/runtime/ranger/access_type.h
+++ b/src/runtime/ranger/access_type.h
@@ -25,21 +25,28 @@ namespace ranger {
 // ACL type defined in Range service for RPC matching policy
 enum class access_type : uint8_t
 {
-    KInvalid = 0,
-    KRead = 1,
-    KWrite = 1 << 1,
-    KCreate = 1 << 2,
-    KDrop = 1 << 3,
-    KList = 1 << 4,
-    KMetadata = 1 << 5,
-    KControl = 1 << 6
+    kInvalid = 0,
+    kRead = 1,
+    kWrite = 1 << 1,
+    kCreate = 1 << 2,
+    kDrop = 1 << 3,
+    kList = 1 << 4,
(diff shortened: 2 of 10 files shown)


foreverneverer (dismissed):

src/common/json_helper.h
```diff
@@ -493,6 +494,26 @@ inline bool json_decode(const JsonObject &in, std::set<T> &t)
     return true;
 }
 
+template <typename T1, typename T2>
+inline void json_encode(JsonWriter &out, const std::unordered_set<T1, T2> &t)
```
acelyc111, line 498: Is it possiable to make std::unordered_set as a template type too? then it can replace the template function above.

src/common/json_helper.h
```diff
@@ -493,6 +494,26 @@ inline bool json_decode(const JsonObject &in, std::set<T> &t)
     return true;
 }
 
+template <typename T1, typename T2>
```
acelyc111, line 497: `typename T2` can be removed?

src/runtime/test/ranger_resource_policy_manager_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+TEST(ranger_resource_policy_manager_test, parse_policies_from_json_for_test)
+{
+    std::string data =
+        "[{\"accesses\":[{\"type\":\"create\",\"isAllowed\":true},{\"type\":\"drop\",\"isAllowed\":"
+        "true},{\"type\":\"control\",\"isAllowed\":true},{\"type\":\"metadata\",\"isAllowed\":true}"
+        ",{\"type\":\"list\",\"isAllowed\":true}],\"users\":[\"user1\",\"user2\"],\"groups\":[],"
+        "\"roles\":[],\"conditions\":[],\"delegateAdmin\":true},{\"accesses\":[{\"type\":\"read\","
+        "\"isAllowed\":true},{\"type\":\"write\",\"isAllowed\":true}],\"users\":[\"user2\"],"
+        "\"groups\":[],\"roles\":[],\"conditions\":[],\"delegateAdmin\":true}]";
```
acelyc111, line 34: Use string literal, ref: https://en.cppreference.com/w/c/language/string_literal, src/meta/test/meta_http_service_test.cpp:L55

src/runtime/ranger/ranger_resource_policy_manager.h
```diff
@@ -32,8 +32,28 @@ namespace replication {
 class meta_service;
 }
 
+enum class resource_type
+{
+    KGlobal = 0,
+    Kdatabase,
+    KDatabase_table,
```
acelyc111, line 39:
```suggestion
    KDatabaseTable,
```

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -36,6 +64,39 @@ void register_rpc_access_type(access_type ac_type,
...
+                                                       {"LIST", access_type::KList},
+                                                       {"METADATA", access_type::KMetadata},
+                                                       {"CONTROL", access_type::KControl}});
+
+// Parse Ranger ACL policies in JSON format 'data' into 'policies'.
+void parse_policies_from_json(const rapidjson::Value &data, std::vector<policy_item> &policies)
+{
+    CHECK(policies.empty(), "Ranger policy list should not be empty.");
+    RETURN_VOID_IF_NOT_ARRAY(data);
+    for (auto &item : data.GetArray()) {
```
acelyc111, line 82:
```suggestion
    for (const auto &item : data.GetArray()) {
```

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -36,6 +64,39 @@ void register_rpc_access_type(access_type ac_type,
...
+        policy_item pi;
```
acelyc111, line 83: In `struct policy_item`, add the default value `access_type::KInvalid` to access_types like:
```
struct policy_item
{
    access_type access_types = access_type::KInvalid;
    ...
```
Then the assignment here can be omitted.

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -36,6 +64,39 @@ void register_rpc_access_type(access_type ac_type,
...
+        pi.access_types = access_type::KInvalid;
+        CONTINUE_IF_MISSING_MEMBER(item, "accesses");
+        for (const auto &access : item["accesses"].GetArray()) {
+            if (access["isAllowed"].GetBool()) {
```
acelyc111, line 87: Need check `access` has "isAllowed" at first?

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -36,6 +64,39 @@ void register_rpc_access_type(access_type ac_type,
...
+                std::string type = access["type"].GetString();
```
acelyc111, line 88: Check has "type" at first?

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -36,6 +64,39 @@ void register_rpc_access_type(access_type ac_type,
...
+                std::transform(type.begin(), type.end(), type.begin(), toupper);
+                pi.access_types = pi.access_types | access_type_maping[type];
```
acelyc111, line 90:
```suggestion
                pi.access_types |= access_type_maping[type];
```

src/common/json_helper.h
WHBANG: No need, if i want to use custom type data, just add a new hash class<T2>.

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -97,5 +158,11 @@ ranger_resource_policy_manager::ranger_resource_policy_manager(
                               "RPC_CM_RENAME_APP"},
                              _ac_type_of_database_rpcs);
 }
+
+void parse_policies_from_json_for_test(const rapidjson::Value &data,
+                                       std::vector<policy_item> &policies)
+{
+    parse_policies_from_json(data, policies);
```
acelyc111, line 165: It's a bit of strenge, how about define `parse_policies_from_json` as a static private function instead?

src/runtime/test/ranger_resource_policy_manager_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include <gtest/gtest.h>
+
+#include "runtime/ranger/ranger_resource_policy.h"
+#include "runtime/ranger/ranger_resource_policy_manager.h"
+
+namespace dsn {
+namespace ranger {
+
+TEST(ranger_resource_policy_manager_test, parse_policies_from_json_for_test)
```
acelyc111, line 26: It's needed to add tests to check that the objects can be serialized correctly too.

src/common/json_helper.h
WHBANG: I don't agree, the template type is too deep, it will only increase the complexity of code reading, not friendly to developers and viewers.

acelyc111: Template is often used to reduce duplicate code, now the new added function is duplicate with `inline bool json_decode(const JsonObject &in, std::set<T> &t)`.
Reference to `json_encode_map` how to deal with `map` and `unordered_map`

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -46,12 +47,18 @@ extern access_type operator|(access_type lhs, access_type rhs);
 
 extern access_type operator&(access_type lhs, access_type rhs);
 
+extern access_type &operator|=(access_type &lhs, access_type rhs);
```
acelyc111, line 50: Just a bit of curious that can we remove the `extern`?

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -46,12 +47,18 @@ extern access_type operator|(access_type lhs, access_type rhs);
...
+
+extern uint8_t access_type_to_int8_t(const access_type &ac_type);
+
 // Ranger policy data structure
 struct policy_item
 {
-    access_type access_types;
+    uint8_t access_types;
```
acelyc111, line 57: I guess you changed it from `access_type` to `uint8_t` is because that enum class is not convenient in same cases, right?

src/runtime/test/ranger_resource_policy_manager_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+{
+    std::string data =
+        R"([{"accesses":)"
```
acelyc111, line 29: Futher more, string literal can simplify multiple lines of string.

src/runtime/ranger/ranger_resource_policy_manager.cpp
WHBANG: Yes, when serializing like `access_type::KRead | access_type::KWrite`, will fail

src/runtime/ranger/ranger_resource_policy.h
WHBANG: Yes, when serializing like access_type::KRead | access_type::KWrite, will fail

src/runtime/test/ranger_resource_policy_manager_test.cpp
WHBANG: done

foreverneverer (approved):

src/runtime/ranger/ranger_resource_policy.h
acelyc111: I found you have overrided some operators, e.g. `|`, `&`, and etc, so where else needed to use `uint8_t`?

@empiredan do you think it's worth to use enum class? It seems there are too many static_casts.

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -22,6 +22,34 @@
...
+#define CONTINUE_IF_MISSING_MEMBER(obj, member)                                                    \
+    do {                                                                                           \
+        if (!obj.IsObject() || !obj.HasMember(member)) {                                           \
+            continue;                                                                              \
+        }                                                                                          \
+    } while (0)
+
+#define RETURN_ERR_IF_NOT_ARRAY(obj)                                                               \
+    do {                                                                                           \
+        if (!obj.IsArray() || obj.Size() == 0) {                                                   \
```
acelyc111, line 41:
```suggestion
        if (!obj.IsArray() || obj.Empty()) {                                                   \
```

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -22,6 +22,34 @@
...
+            return dsn::ERR_RANGER_PARSE_ACL;                                                      \
+        }                                                                                          \
+    } while (0)
+
+#define RETURN_VOID_IF_NOT_ARRAY(obj)                                                              \
+    do {                                                                                           \
+        if (!obj.IsArray() || obj.Size() == 0) {                                                   \
```
acelyc111, line 48:
```suggestion
        if (!obj.IsArray() || obj.Empty()) {                                                   \
```

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -97,5 +134,32 @@ ranger_resource_policy_manager::ranger_resource_policy_manager(
...
                              _ac_type_of_database_rpcs);
 }
+
+void ranger_resource_policy_manager::parse_policies_from_json(const rapidjson::Value &data,
+                                                              std::vector<policy_item> &policies)
+{
+    CHECK(policies.empty(), "Ranger policy list should not be empty.");
+    RETURN_VOID_IF_NOT_ARRAY(data);
+    for (const auto &item : data.GetArray()) {
+        policy_item pi;
```
acelyc111, line 144: nit: move to L155 where is closer to the place use it?

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -42,16 +43,22 @@ enum class access_type : uint8_t
...
+
+access_type &operator|=(access_type &lhs, access_type rhs);
+
+uint8_t access_type_to_int8_t(const access_type &ac_type);
 
 // Ranger policy data structure
 struct policy_item
 {
-    access_type access_types;
+    uint8_t access_types;
```
acelyc111, line 57:
```suggestion
    uint8_t access_types = 0;
```

src/runtime/ranger/ranger_resource_policy.h
WHBANG: For example: I need to store `access_type::KRead | access_type::KWrite` in zk, but such a combination cannot be serialized correctly, it will be serialized as `unknown`, if I deserialize it, it will lose `access_type::KRead | access_type::KWrite` meaning, so I need to convert it to `uint8_t`.
mainly after c++11, it is recommended to use enum class. If it can handle the logical relationship well, I think it is no problem. What do you think, hope your reply

src/common/json_helper.h
```diff
@@ -335,6 +337,28 @@ INT_TYPE_SERIALIZATION(int16_t)
...
+inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
+{
+    dverify(in.IsUint64());
+    int64_t ans = in.GetUint64();
+    int64_t act_min = static_cast<uint64_t>(dsn::ranger::access_type::KInvalid);
+    int64_t act_max = static_cast<uint64_t>(
+        dsn::ranger::access_type::KRead | dsn::ranger::access_type::KWrite |
+        dsn::ranger::access_type::KCreate | dsn::ranger::access_type::KDrop |
+        dsn::ranger::access_type::KList | dsn::ranger::access_type::KMetadata |
+        dsn::ranger::access_type::KControl);
```
acelyc111, line 355: It would great if define `act_min` and `act_max` as static variables in runtime/ranger/access_type.h

src/common/json_helper.h
```diff
@@ -335,6 +337,28 @@ INT_TYPE_SERIALIZATION(int16_t)
...
+    dverify(ans >= act_min && ans <= act_max);
+    using act = std::underlying_type<dsn::ranger::access_type>::type;
```
acelyc111, line 357: How about define move `using act = std::underlying_type<dsn::ranger::access_type>::type;` to access_type.h too? Then remove duplicate codes.

src/common/json_helper.h
```diff
@@ -335,6 +337,28 @@ INT_TYPE_SERIALIZATION(int16_t)
...
+    t = (dsn::ranger::access_type) static_cast<act>(ans);
```
acelyc111, line 358: Why nor `t = static_cast<dsn::ranger::access_type>(ans);` directly?

src/common/json_helper.h
```diff
@@ -335,6 +337,23 @@ INT_TYPE_SERIALIZATION(int16_t)
...
+{
+    out.Uint64(static_cast<uint64_t>(t));
+}
+
+inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
+{
+    dverify(in.IsUint64());
+    int64_t ans = in.GetUint64();
+    int64_t act_min = static_cast<uint64_t>(dsn::ranger::access_type_min);
+    int64_t act_max = static_cast<uint64_t>(dsn::ranger::access_type_max);
```
acelyc111, line 351: better to use `uint64_t` or `auto`, since you have used `GetUint64` and `static_cast<uint64_t>`

acelyc111, line 351:
```suggestion
    auto ans = in.GetUint64();
    auto act_min = static_cast<uint64_t>(dsn::ranger::access_type_min);
    auto act_max = static_cast<uint64_t>(dsn::ranger::access_type_max);
```

acelyc111 (approved):

src/runtime/ranger/access_type.h
```diff
@@ -0,0 +1,52 @@
...
+
+using act = std::underlying_type<access_type>::type;
+
+access_type operator|(access_type lhs, access_type rhs);
+
+access_type operator&(access_type lhs, access_type rhs);
+
+access_type &operator|=(access_type &lhs, access_type rhs);
+
+static access_type access_type_min = access_type::kInvalid;
```
empiredan, line 46:
```suggestion
const access_type kAccessTypeMin = access_type::kInvalid;
```

src/runtime/ranger/access_type.h
```diff
@@ -0,0 +1,52 @@
...
+static access_type access_type_max =
```
empiredan, line 47:
```suggestion
const access_type kAccessTypeMax =
```

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -36,6 +64,15 @@ void register_rpc_access_type(access_type ac_type,
         ac_type_of_rpc.emplace(code, ac_type);
     }
 }
+
+// Used to map access_type matched resources policies json string.
+std::map<std::string, access_type> access_type_maping({{"READ", access_type::kRead},
```
empiredan, line 69:
```suggestion
const std::map<std::string, access_type> kAccessTypeMaping({{"READ", access_type::kRead},
```

src/runtime/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -97,5 +134,30 @@ ranger_resource_policy_manager::ranger_resource_policy_manager(
...
+    for (const auto &item : data.GetArray()) {
+        CONTINUE_IF_MISSING_MEMBER(item, "accesses");
+        policy_item pi;
+        for (const auto &access : item["accesses"].GetArray()) {
+            CONTINUE_IF_MISSING_MEMBER(access, "isAllowed");
+            CONTINUE_IF_MISSING_MEMBER(access, "type");
+            if (access["isAllowed"].GetBool()) {
+                std::string type = access["type"].GetString();
+                std::transform(type.begin(), type.end(), type.begin(), toupper);
+                pi.access_types |= access_type_maping[type];
```
empiredan, line 152: Is it ensured that `type` must be in `access_type_maping` ?
```suggestion
                pi.access_types |= kAccessTypeMaping.at(type);
```

WHBANG: Oh, there may be other ways to define the Range Service. there may be different access_type. I will ignore them directly

src/common/json_helper.h
```diff
@@ -347,8 +347,8 @@ inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
 {
     dverify(in.IsUint64());
     auto ans = in.GetUint64();
-    auto act_min = static_cast<uint64_t>(dsn::ranger::access_type_min);
-    auto act_max = static_cast<uint64_t>(dsn::ranger::access_type_max);
+    auto act_min = static_cast<uint64_t>(dsn::ranger::kAccessTypeMin);
```
acelyc111, line 350: Maybe use 'empty'/'none' and 'full'/'all' instead of 'min' and 'max'? The first sight of min and max looks like the value of access_type, i.e. kInvalid and kControl.

src/common/json_helper.h
```diff
@@ -347,8 +347,8 @@ inline bool json_decode(const JsonObject &in, dsn::ranger::access_type &t)
 {
     dverify(in.IsUint64());
     auto ans = in.GetUint64();
-    auto act_min = static_cast<uint64_t>(dsn::ranger::kAccessTypeMin);
-    auto act_max = static_cast<uint64_t>(dsn::ranger::kAccessTypeMax);
+    auto act_min = static_cast<uint64_t>(dsn::ranger::kAccessTypeNone);
+    auto act_max = static_cast<uint64_t>(dsn::ranger::kAccessTypeAll);
     dverify(ans >= act_min && ans <= act_max);
```
acelyc111:
```suggestion
    auto act_none = static_cast<uint64_t>(dsn::ranger::kAccessTypeNone);
    auto act_all = static_cast<uint64_t>(dsn::ranger::kAccessTypeAll);
    dverify(ans >= act_none && ans <= act_all);
```

acelyc111 (approved):

empiredan (approved):

PR #1379, opened 2023-03-01
feat(Ranger): Compatible with the old ACL and the new ACL

WHBANG:

This patch is compatible with old and new acl.

- Modify some method names and parameter names to make them more accurate.
- Defines the configuration parameters that the new ACL needs to use.
- Two new 'allowed()' methods are provided for meta_server and replica_server.
- Some incompatible methods will be removed, (allowed&pre_check)commented.

### DIFF
--- a/src/runtime/security/access_controller.cpp
+++ b/src/runtime/security/access_controller.cpp
@@ -26,9 +26,10 @@
 namespace dsn {
 namespace security {
 DSN_DEFINE_bool(security, enable_acl, false, "whether enable access controller or not");
-
-DSN_DEFINE_bool(security, enable_ranger_acl, false, "whether enable access controller or not");
-
+DSN_DEFINE_bool(security,
+                enable_ranger_acl,
+                false,
+                "whether enable access controller integrate to Apache Ranger or not");
 DSN_DEFINE_string(security,
                   super_users,
                   "",
@@ -51,7 +52,6 @@ bool access_controller::pre_check(const std::string &user_name)
         return true;
     }
     return false;
-    return _super_users.find(user_name) != _super_users.end();
 }
 
 bool access_controller::is_enable_ranger_acl() { return FLAGS_enable_ranger_acl; }
--- a/src/runtime/security/access_controller.cpp
+++ b/src/runtime/security/access_controller.cpp
@@ -26,11 +26,22 @@
 namespace dsn {
 namespace security {
 DSN_DEFINE_bool(security, enable_acl, false, "whether enable access controller or not");
-DSN_TAG_VARIABLE(enable_acl, FT_MUTABLE);
 
-DSN_DEFINE_string(security, super_users, "", "super user for access controller");
+DSN_DEFINE_bool(security, enable_ranger_acl, false, "whether enable access controller or not");
 
-access_controller::access_controller() { utils::split_args(FLAGS_super_users, _super_users, ','); }
+DSN_DEFINE_string(security,
+                  super_users,
+                  "",
+                  "super users for access controller, comma-separated list of user names");
+
+access_controller::access_controller()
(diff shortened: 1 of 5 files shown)


src/runtime/security/access_controller.cpp
```diff
@@ -26,11 +26,22 @@
 namespace dsn {
 namespace security {
 DSN_DEFINE_bool(security, enable_acl, false, "whether enable access controller or not");
-DSN_TAG_VARIABLE(enable_acl, FT_MUTABLE);
```
acelyc111, line 29: nit: Remove the blank line.

src/runtime/security/access_controller.cpp
```diff
@@ -26,11 +26,22 @@
...
 
-DSN_DEFINE_string(security, super_users, "", "super user for access controller");
+DSN_DEFINE_bool(security, enable_ranger_acl, false, "whether enable access controller or not");
```
acelyc111, line 30:
```suggestion
DSN_DEFINE_bool(security, enable_ranger_acl, false, "whether enable access controller integrate to Apache Ranger or not");
```

src/runtime/security/access_controller.cpp
```diff
@@ -26,11 +26,22 @@
...
 
-access_controller::access_controller() { utils::split_args(FLAGS_super_users, _super_users, ','); }
+DSN_DEFINE_string(security,
+                  super_users,
+                  "",
```
acelyc111, line 34: What happen if enable acl but leave super_users as empty?

src/runtime/security/access_controller.cpp
```diff
@@ -40,6 +51,14 @@ bool access_controller::pre_check(const std::string &user_name)
         return true;
     }
     return false;
+    return _super_users.find(user_name) != _super_users.end();
+}
+
+bool access_controller::is_enable_ranger_acl() { return FLAGS_enable_ranger_acl; }
```
acelyc111, line 57: Why not use `FLAGS_enable_ranger_acl` directly?

src/runtime/security/access_controller.h
```diff
@@ -31,27 +33,41 @@ class access_controller
...
-    /**
-     * check if the message received is allowd to do something.
-     *   msg - the message received
-     **/
+    // Check if the message received is allowd to access the table.
+    // msg - the message received
+    // app_name - tables involved in ACL
+    virtual bool allowed(message_ex *msg, const std::string &app_name) { return false; }
+
+    // (TODO:wanghao) : this method will be deleted in the next patch.
```
acelyc111, line 52:
```suggestion
    // TODO(wanghao): this method will be deleted in the next patch.
```

src/runtime/security/access_controller.h
```diff
@@ -31,27 +33,41 @@ class access_controller
...
+    // check if the message received is allowd to do something.
+    // msg - the message received
     virtual bool allowed(message_ex *msg) = 0;
 
 protected:
+    // (TODO:wanghao) : this method will be deleted in the next patch.
```
acelyc111, line 58:
```suggestion
    // TODO(wanghao): this method will be deleted in the next patch.
```

src/runtime/security/access_controller.cpp
WHBANG: Both `dsn_meta_server` and `dsn_replica_server` will use it, so `FLAGS_enable_ranger_acl` cannot be used directly.

acelyc111 (approved):

src/runtime/security/access_controller.cpp
WHBANG: yes

empiredan (approved):

PR #1393, opened 2023-03-14
fix: jaas conf; make thread daemon

### LINKED ISSUE #1389: java client krb not work with  kinit
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.
use keytab and principal with java client. 

2. What did you expect to see?
use keytab not fail.

4. What did you see instead?
when i use kinit ,client will auth fail.

6. What version of Pegasus are you using?
2.3.4

shalk:

fix #1389 

##### Tests 

- Unit test
- Integration test
- No code

##### Code changes

- Has persistent data change

##### Side effects

- Increased code complexity

##### Related changes

### DIFF
--- a/java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
+++ b/java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import javax.security.auth.Subject;
 import javax.security.auth.kerberos.KerberosPrincipal;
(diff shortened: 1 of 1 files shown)


java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
```diff
@@ -53,7 +54,18 @@ class KerberosProtocol implements AuthProtocol {
   private String keyTab;
   private String principal;
   final int CHECK_TGT_INTEVAL_SECONDS = 10;
-  final ScheduledExecutorService service = Executors.newSingleThreadScheduledExecutor();
+  final ScheduledExecutorService service =
+      Executors.newSingleThreadScheduledExecutor(new TgtThreadFactory());
+
+  static class TgtThreadFactory implements ThreadFactory {
```
Apache9, line 60: Just create an anonymous class is enough?

java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
```diff
@@ -161,19 +173,16 @@ private static Configuration getLoginContextConfiguration(String keyTab, String
       @Override
       public AppConfigurationEntry[] getAppConfigurationEntry(String name) {
         Map<String, String> options = new HashMap<>();
-        // TGT is obtained from the ticket cache.
-        options.put("useTicketCache", "true");
```
Apache9, line 165: This is for keytab based login so we should remove useTicketCache, but do we support login use ticket cache? There is another method for ticker cache based login?

shalk: yes, we can use ticket cache without keytab.  these can be done in the same method like hadoop

Apache9 (approved):

empiredan (approved):

Issue #1342, opened 2023-02-08
Feature(new_metrics): migrate metrics for `replica` class


empiredan:
All of the metrics in `replica` class are replica-level. As is described in https://github.com/apache/incubator-pegasus/issues/1326, since all classes that hold replica-level metrics are sub-object to `replica` class, all of replica-level metrics are bound to `replica` class. Thus once `replica` class is destructed, *replica* entity will be retired (since reference count of entity will decrease to 1).

Following metrics are the members of `replica` ([replica.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/replica/replica.cpp)):

| Variables | Types/Computations |
| :-------: | :------------------: |
| _counter_private_log_size | Gauge |
| _counter_recent_write_throttling_delay_count | increase(Counter) |
| _counter_recent_write_throttling_reject_count | increase(Counter) |
| _counter_recent_read_throttling_delay_count | increase(Counter) |
| _counter_recent_read_throttling_reject_count | increase(Counter) |
| _counter_recent_backup_request_throttling_delay_count | increase(Counter) |
| _counter_recent_backup_request_throttling_reject_count | increase(Counter) |
| _counter_dup_disabled_non_idempotent_write_count | increase(Counter) |
| _counter_recent_read_splitting_reject_count | increase(Counter) |
| _counter_recent_write_splitting_reject_count | increase(Counter) |
| _counter_recent_write_bulk_load_ingestion_reject_count | increase(Counter) |

Note that `increment()` invocations for `_counter_recent_*_throttling_delay_count` and `_counter_recent_*_throttling_reject_count` are in [replica_throttle.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/replica/replica_throttle.cpp).

PR #1397, opened 2023-03-15
feat(new_metrics): migrate replica-level metrics for replica class

empiredan:
https://github.com/apache/incubator-pegasus/issues/1342

Migrate replica-level metrics in `replica` class to new framework, including the size of
private log, the number of backup requests, the number of read/write/backup requests
that are delayed/rejected due to throttling, the number of read/write requests that are
rejected due to splitting, the number of write requests that are rejected due to bulk load
ingestion, the number of non-idempotent write requests that are rejected due to duplication.

Note that the old metrics for the number of backup requests, the number of delayed/rejected
backup requests due to throttling and the number of rejected non-idempotent write requests
are table-level, which have been changed to replica-level in new metrics, since they could
be aggregated to table-level if necessary.


src/replica/replica_chkpt.cpp
```diff
@@ -168,9 +169,10 @@ void replica::on_checkpoint_timer()
                                  valid_start_offset,
                                  (int64_t)FLAGS_log_private_reserve_max_size_mb * 1024 * 1024,
                                  (int64_t)FLAGS_log_private_reserve_max_time_seconds);
-                             if (status() == partition_status::PS_PRIMARY)
-                                 _counter_private_log_size->set(_private_log->total_size() /
-                                                                1000000);
+                             if (status() == partition_status::PS_PRIMARY) {
+                                 METRIC_VAR_SET(private_log_size_mb,
+                                                _private_log->total_size() / 1000000);
```
acelyc111, line 174: how about using `_private_log->total_size() >> 20` ?

src/replica/replica.h
```diff
@@ -118,18 +119,20 @@ class test_checker;
...
-            response_client_##op_type(request, ERR_SPLITTING);                                     \
-            _counter_recent_##op_type##_splitting_reject_count->increment();                       \
-            return;                                                                                \
-        }                                                                                          \
-        if (!_split_mgr->check_partition_hash(                                                     \
-                ((dsn::message_ex *)request)->header->client.partition_hash, #op_type)) {          \
-            response_client_##op_type(request, ERR_PARENT_PARTITION_MISUSED);                      \
-            return;                                                                                \
+    do {                                                                                           \
+        if (_validate_partition_hash) {                                                            \
```
acelyc111, line 123: How about reduce indent like this? 
```
if (!_validate_partition_hash) {
    break;
}
...
```

empiredan: OK.

Issue #1383, opened 2023-03-06
RocksDB corruption leads to endless failure


acelyc111:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
Some RocksDB instances corrupted for some reasons, maybe disk driver IO error, data corruption, etc.

2. What did you expect to see?
The cluster can recover from error automatically.

3. What did you see instead?
The replica server will close the replica when see write errors, but will start the replica again in the same place whose data is still corrupted. And then, the error occured again and again.

For read requests, the replica will not handle the error instead, then read requests will fail again and again.

Write:
![image](https://user-images.githubusercontent.com/10775040/225831222-feba4822-9f91-4418-b322-2198fafd1c86.png)

Read:
![image](https://user-images.githubusercontent.com/10775040/225831579-687c2a0d-26ae-43e4-a19f-de8fa0d9ab97.png)

4. What version of Pegasus are you using?
1.12, 2.0

PR #1399, opened 2023-03-16
fix: Fault-tolerant storage engine errors for write operations

acelyc111:

Make Pegasus fault-tolerant storage engine (i.e. RocksDB) error
for write operations. The unrecoverable RocksDB instance
directory will be moved to a .err path and the replica will be
closed and removed from stub. Then meta server will detect the
missing replica and recover automatically.
This patch only deal with storage engine unrecoverable errors
for write operations.


empiredan:
Currently primary replica server respond to client after rocksdb has been written. However, rocksdb write interface may return `kCorruption` or `kIOError`, which will be returned to client and client will think this request has failed. In fact all of primary and secondary logs have been written successfully thus this request should be considered successful. Client will choose to write again and will lead to inconsistency for the non-idempotent writes such as `incr`, `check_and_set` and `check_and_mutate`.

To solve this problem, I think we can make rocksdb write asynchronous. Once fail to write rocksdb asynchronously, for example, `kCorruption` or `kIOError`, just remove the replica and move the rocksdb directory to `.err` and move this primary replica to other secondary replica. The consistency will be guaranteed if and only if logs are consistent. We can just write rocksdb asynchrously.

Issue #1343, opened 2023-02-08
Feature(new_metrics): migrate miscellaneous replica-level metrics


empiredan:
Some classes that have few of replica-level metrics are collected into one issue to be migrated.

<br/>

Following metrics are the members of `pegasus_event_listener` ([pegasus_event_listener.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/server/pegasus_event_listener.cpp)), which is created at the construction of `pegasus_server_impl`:
| Variables | Types/Computations |
| :--------: | :-------------------: |
| _pfc_recent_rdb_compaction_input_bytes | increase(Counter) |
| _pfc_recent_rdb_compaction_output_bytes | increase(Counter) |

<br/>

Following metrics are the members of `pegasus_mutation_duplicator` ([pegasus_mutation_duplicator.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/server/pegasus_mutation_duplicator.cpp)), which is created in the call chain `ship_mutation::ship_mutation()` ← `replica_duplicator::start_dup_log()` ← `replica_duplicator_manager::sync_duplication() ← replica::replica()` (actually the metrics in `pegasus_mutation_duplicator` of a replica share one metric instance):
| Variables | Types/Computations |
| :--------: | :-------------------: |
| _shipped_ops | rate(Counter) |
| _failed_shipping_ops | rate(Counter) |

PR #1407, opened 2023-03-22
feat(new_metrics): migrate replica-level metrics for pegasus_event_listener

empiredan:
https://github.com/apache/incubator-pegasus/issues/1343

Migrate replica-level metrics in pegasus_event_listener class to new framework, all of which
are rocksdb-related, including the number of completed flushes/compactions, the size of
flush output in bytes, the size of compaction input/output in bytes.

Note that in old perf counters there are just 2 replica-level metrics for pegasus_event_listener
while all of others are server-level. Migrated to new framework all of the metrics have become
replica-level; once server-level metrics are needed, just aggregate on replica-level ones.

### DIFF
--- a/src/utils/metrics.h
+++ b/src/utils/metrics.h
@@ -620,6 +620,9 @@ enum class metric_unit : size_t
     kKeys,
     kFiles,
     kAmplification,
+    kFlushes,
+    kCompactions,
+    kWrites,
     kInvalidUnit,
 };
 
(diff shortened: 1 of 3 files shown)


src/utils/metrics.h
```diff
@@ -620,6 +620,9 @@ enum class metric_unit : size_t
     kKeys,
     kFiles,
     kAmplification,
+    kFlushes,
```
acelyc111, line 623: How about keep them in dict order?

acelyc111 (approved):

Issue #1321, opened 2023-01-20
Feature(new_metrics): migrate built-in metrics


empiredan:
In the framework of `perf_counters`, built-in metrics are actually two metrics:

- *memused.res(MB)*, the usage of physical memory
- *memused.virt(MB)*, the size of virtual memory

They are collected in a timer each `FLAGS_perf_counter_update_interval_seconds` (typically 10 seconds).

In the framework of new metrics, we can also introduce the built-in metrics. During `start()` for both `pegasus_replication_service_app` and `pegasus_meta_service_app`, a timer will be launched to collect built-in metrics periodically. The built-in metrics will be attached to *server* entity, which is defined in `metrics.cpp`. Both of them will be *gauges*.

PR #1418, opened 2023-03-27
feat(new_metrics): migrate built-in server-level metrics

empiredan:
https://github.com/apache/incubator-pegasus/issues/1321

Migrate built-in server-level metrics to new framework, including the total
amount of virtual/physical memory usage in MB. Both metrics are wrapped
in `builtin_metrics` and updated periodically by a timer, which is started
or stopped along with meta/replica servers.

Since `dsn_perf_counter_test` involves both metrics which have been
removed from perf-counters, just disable it.


src/utils/builtin_metrics.h
```diff
@@ -15,19 +15,35 @@
...
+#include <memory>
+
+#include "utils/metrics.h"
+#include "utils/ports.h"
 #include "utils/singleton.h"
 
 namespace dsn {
-class builtin_counters : public dsn::utils::singleton<builtin_counters>
+
+class builtin_metrics : public utils::singleton<builtin_metrics>
```
acelyc111, line 28: Is it possible to avoid using singleton?
https://www.davidtanzer.net/david's%20blog/2016/03/14/6-reasons-why-you-should-avoid-singletons.html

src/utils/builtin_metrics.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+}
+
+builtin_metrics::~builtin_metrics() { stop(); }
+
+void builtin_metrics::on_close() {}
+
+void builtin_metrics::start()
+{
+    if (_timer) {
```
acelyc111, line 57: In which case `start()` will be called multiple times?

src/utils/builtin_metrics.h
empiredan: OK, I'll use member instead of singleton. Since in most cases an instance is started as a meta server or a replica server, in fact there tends to be an object for `builtin_metrics`.

src/utils/builtin_metrics.cpp
empiredan: Currently it is just called at the starting of meta/replica server.

src/server/pegasus_service_app.h
```diff
@@ -47,6 +48,7 @@ class pegasus_replication_service_app : public ::dsn::replication::replication_s
 
         if (ret == ::dsn::ERR_OK) {
             pegasus_counter_reporter::instance().start();
```
acelyc111, line 50: It's strange a singleton to explicit start, we can improve it later, define it as a common object too?

src/server/pegasus_service_app.h
```diff
@@ -57,11 +59,13 @@ class pegasus_replication_service_app : public ::dsn::replication::replication_s
...
         if (_updater_started) {
             pegasus_counter_reporter::instance().stop();
+            _builtin_metrics.stop();
         }
         return ret;
     }
 
 private:
+    dsn::builtin_metrics _builtin_metrics;
     bool _updater_started;
```
acelyc111, line 69: If pegasus_counter_reporter and builtin_metrics are both common objects, and they can stop automatically, _updater_started can be removed?

src/utils/builtin_metrics.cpp
acelyc111: So it must be a bug if start builtin_metrics multiple times, how about just CHECK `_timer` must be null when start?

src/utils/builtin_metrics.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    _timer.reset(new metric_timer(FLAGS_builtin_metrics_update_interval_ms,
+                                  std::bind(&builtin_metrics::update, this),
+                                  std::bind(&builtin_metrics::on_close, this)));
+}
+
+void builtin_metrics::stop()
+{
+    if (!_timer) {
```
acelyc111, line 68: Same, in which case builtin_metrics will be stopped multiple times?

src/utils/builtin_metrics.cpp
empiredan: OK, I'll do CHECK for the state before each operation is run.

src/server/pegasus_service_app.h
empiredan: Actually all codes about perf counters will be removed after all metrics are migrated. I'll add a TODO here to comment.

src/server/pegasus_service_app.h
empiredan: `service_app::start(...)` (such as `dsn::replication::replication_service_app::start(args_new)`) will return an error code. Once this error code is not `ERR_OK`, metric service will not be started. That's why `_updater_started ` is involved.

We can do another CHECK for the returned error code by `service_app::start(...)`. Actually the root caller, `start_app()` in service_control_task::exec() will also do CHECK for `ERR_OK`. Do CHECK here to guarantee that all following services (such as built-in metrics) are started, then `_updater_started ` could be removed.

src/utils/builtin_metrics.cpp
empiredan: OK.

Issue #1425, opened 2023-03-30
Feature(new_metrics): add disk-level metric entity and migrate disk-level metrics for `fs_manager`


empiredan:
Following metrics are the members of `fs_manager` ([fs_manager.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/common/fs_manager.cpp)), which is created at the construction of `replica_stub`:
| Variables | Types/Computations |
| :-------: | :------------------: |
| _counter_total_capacity_mb | Gauge |
| _counter_total_available_mb | Gauge |
| _counter_total_available_ratio | Gauge |
| _counter_min_available_ratio | Gauge |
| _counter_max_available_ratio | Gauge |

PR #1427, opened 2023-03-31
feat(new_metrics): add disk-level metric entity and migrate disk-level metrics for fs_manager

empiredan:
https://github.com/apache/incubator-pegasus/issues/1425

In perf counters, all metrics of `fs_manager` are server-level. For example,
the total capacity and the available capacity of all disks where there are
data of pegasus.

However, sometimes the capacity and the available capacity of each disk
seem more important: no space left on the disk will lead to serious problems.
Therefore, after being migrated to new framework, the server-level metrics
of perf counters become disk-level, including the capacity and the available
capacity of a disk. As for another disk-level metric -- the available percentage
of each disk used by a replica server, just use division operator.

Once server-level metrics are needed, just aggregate on the disk-level ones.
To compute another 2 server-level metrics -- the minimal/maximal available
percentage among all disks used by a replica server in a node, for example,
just use min/max operators over disk-level ones for Prometheus.

To implement disk-level metrics, disk-level metric entity are also added.

### DIFF
--- a/src/common/fs_manager.cpp
+++ b/src/common/fs_manager.cpp
@@ -41,7 +41,7 @@
 
 #include "common/gpid.h"
 #include "common/replication_enums.h"
-#include "perf_counter/perf_counter.h"
+#include "fmt/core.h"
 #include "runtime/api_layer1.h"
 #include "runtime/rpc/rpc_address.h"
 #include "utils/fail_point.h"
--- a/src/common/fs_manager.cpp
+++ b/src/common/fs_manager.cpp
@@ -80,25 +80,22 @@ DSN_TAG_VARIABLE(disk_min_available_space_ratio, FT_MUTABLE);
 
 namespace {
 
-metric_entity_ptr instantiate_disk_metric_entity(const std::string &tag, const std::string &data_dir)
+metric_entity_ptr instantiate_disk_metric_entity(const std::string &tag,
+                                                 const std::string &data_dir)
 {
     auto entity_id = fmt::format("disk_{}", tag);
 
-    return METRIC_ENTITY_disk.instantiate(
-        entity_id,
-        {{"tag", tag},
-         {"data_dir", data_dir}});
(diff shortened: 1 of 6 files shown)


src/common/fs_manager.cpp
```diff
@@ -297,17 +321,29 @@ bool fs_manager::for_each_dir_node(const std::function<bool(const dir_node &)> &
 
 void fs_manager::update_disk_stat(bool check_status_changed)
 {
-    reset_disk_stat();
+    _total_capacity_mb = 0;
+    _total_available_mb = 0;
```
acelyc111, line 325: Is there any other place except this function use it? If not, how about define it as a local variable?

empiredan: Yes, both are used in [replica_stub.cpp#L1156](https://github.com/apache/incubator-pegasus/blob/master/src/replica/replica_stub.cpp#L1156) as the response for `query_disk_info_rpc`.

src/common/fs_manager.cpp
```diff
@@ -108,6 +154,10 @@ bool dir_node::update_disk_stat(const bool update_disk_status)
     disk_available_ratio = static_cast<int>(
         disk_capacity_mb == 0 ? 0 : std::round(disk_available_mb * 100.0 / disk_capacity_mb));
 
+    METRIC_CALL_SET_METHOD(disk_capacity, total_disk_capacity_mb, disk_capacity_mb);
+    METRIC_CALL_SET_METHOD(disk_capacity, avail_disk_capacity_mb, disk_available_mb);
+    METRIC_CALL_SET_METHOD(disk_capacity, avail_disk_capacity_percentage, disk_available_ratio);
```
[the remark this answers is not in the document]
empiredan: OK, I'll drop this metric.

acelyc111 (approved):

src/common/fs_manager.cpp
acelyc111: OK

Issue #1426, opened 2023-03-31
New macro definitions for FQDN


GehaFearless:
relatd to [Support FQDN](https://github.com/apache/incubator-pegasus/issues/1403)

Add some new macro definitions.

PR #1429, opened 2023-04-03
feat: add new macro definitions & make the dockerfile compliant

GehaFearless:
add some macro definitions & fix a little docker build problem

issue: https://github.com/apache/incubator-pegasus/issues/1426


acelyc111 (approved):
LGTM
It would be great it use the new added macros to simplify more code in the codebase, it can be compeleted in another patch.

empiredan (approved):

Issue #1404, opened 2023-03-21
Add struct host_port & dns_resolver


GehaFearless:
A part of  [Support FQDN](https://github.com/apache/incubator-pegasus/issues/1403).

Implement of struct `rpc_host_port` instead of `rpc_address`. 
Also provide dns_resolver transfer `rpc_host_port` to `rpc_address.`

PR #1430, opened 2023-04-04
feat(FQDN): Implement of struct rpc_host_port

GehaFearless:
issue: https://github.com/apache/incubator-pegasus/issues/1404

Implement of struct rpc_host_port

### DIFF
--- a/src/runtime/rpc/rpc_host_port.h
+++ b/src/runtime/rpc/rpc_host_port.h
@@ -57,19 +57,24 @@ public:
     }
 
 private:
-    std::string _host = "";
+    std::string _host;
     uint16_t _port = 0;
     dsn_host_type_t _type = HOST_TYPE_INVALID;
 };
 
 inline bool operator<(const host_port &hp1, const host_port &hp2)
 {
-    if (hp1.type() != hp2.type())
+    if (hp1.type() != hp2.type()) {
         return hp1.type() < hp2.type();
+    }
 
     switch (hp1.type()) {
-    case HOST_TYPE_IPV4:
-        return hp1.host() < hp2.host() || (hp1.host() == hp2.host() && hp1.port() < hp2.port());
+    case HOST_TYPE_IPV4: {
+        int ret = hp1.host().compare(hp2.host());
+        return ret < 0 || (ret == 0 && hp1.port() < hp2.port());
+    }
+    case HOST_TYPE_GROUP:
+        CHECK(false, "type HOST_TYPE_GROUP not support!");
     default:
         return true;
     }
@@ -88,6 +93,8 @@ inline bool operator==(const host_port &hp1, const host_port &hp2)
     switch (hp1.type()) {
     case HOST_TYPE_IPV4:
         return hp1.host() == hp2.host() && hp1.port() == hp2.port();
+    case HOST_TYPE_GROUP:
+        CHECK(false, "type HOST_TYPE_GROUP not support!");
     default:
         return true;
     }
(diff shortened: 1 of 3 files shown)


acelyc111 (dismissed):
Good job!

acelyc111 (dismissed):

src/runtime/rpc/rpc_host_port.h
```diff
@@ -0,0 +1,124 @@
...
+        return os << hp.to_string();
+    }
+
+private:
+    std::string _host;
+    uint16_t _port = 0;
+    dsn_host_type_t _type = HOST_TYPE_INVALID;
+};
+
+inline bool operator<(const host_port &hp1, const host_port &hp2)
```
empiredan, line 66: When we would use `operator<` ?

src/runtime/rpc/rpc_host_port.h
```diff
@@ -0,0 +1,124 @@
...
+
+namespace std {
+template <>
+struct hash<::dsn::host_port>
+{
+    size_t operator()(const ::dsn::host_port &hp) const
+    {
+        switch (hp.type()) {
+        case HOST_TYPE_IPV4:
+            return std::hash<std::string>()(hp.host()) ^ hp.port();
```
empiredan, line 116:
```suggestion
            return std::hash<std::string>()(hp.host()) ^ std::hash<uint16_t>()(hp.port());
```

src/runtime/test/host_port_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+    host_port hp = host_port("localhost", 8080);
+    ASSERT_EQ(HOST_TYPE_IPV4, hp.type());
+    ASSERT_EQ(8080, hp.port());
+    ASSERT_EQ("localhost", hp.host());
+}
+
+TEST(host_port_test, operators)
+{
+    host_port hp("localhost", 8080);
+    ASSERT_EQ(hp, hp);
```
empiredan, line 54: Could add a case that compare with another object which is equal.

src/runtime/test/host_port_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    {
+        host_port new_hp("localhost", 8081);
+        ASSERT_NE(hp, new_hp);
+    }
+
+    host_port hp2;
+    ASSERT_NE(hp, hp2);
+    ASSERT_FALSE(hp.is_invalid());
+    ASSERT_TRUE(hp2.is_invalid());
```
empiredan, line 69: Is it necessary to test `operator<` ?

src/runtime/test/host_port_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+        ASSERT_EQ(std::string("invalid address"), hp.to_string());
+    }
+}
+
+TEST(host_port_test, host_port_build)
+{
+    host_port hp = host_port("localhost", 8080);
+    ASSERT_EQ(HOST_TYPE_IPV4, hp.type());
+    ASSERT_EQ(8080, hp.port());
+    ASSERT_EQ("localhost", hp.host());
```
empiredan, line 48: Could add a case to test ``host_port(rpc_address addr)`, and another case to test default object whose type is invalid.

src/runtime/test/host_port_test.cpp
GehaFearless: in host_port_build

src/runtime/test/host_port_test.cpp
GehaFearless: done

Issue #1331, opened 2023-01-29
Feature(new_metrics): migrate meta metrics


empiredan:
The meta-related metrics migrated to new framework will be attached to *server* entity. All involved classes are put as below.

<br/>

Following metrics are the members of `server_state` ([server_state.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/meta/server_state.cpp)), which is created at the construction of `meta_service`:
| Variables | Types/Computations |
| :-------: | :------------------: |
| _dead_partition_count | Gauge |
| _unreadable_partition_count | Gauge |
| _unwritable_partition_count | Gauge |
| _writable_ill_partition_count | Gauge |
| _healthy_partition_count | Gauge |
| _recent_update_config_count | increase(Counter) |
| _recent_partition_change_unwritable_count | increase(Counter) |
| _recent_partition_change_writable_count | increase(Counter) |

<br/>

Following metrics are the members of `greedy_load_balancer` ([greedy_load_balancer.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/meta/greedy_load_balancer.cpp)), which is created at `meta_service::start()`:

| Variables | Types/Computations |
| :-------: | :------------------: |
| _balance_operation_count | Gauge |
| _recent_balance_move_primary_count | increase(Counter) |
| _recent_balance_copy_primary_count | increase(Counter) |
| _recent_balance_copy_secondary_count | increase(Counter) |

<br/>

Following metrics are the member of `meta_service` ([meta_service.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_service.cpp)), which is created at the construction of `meta_service_app`:
| Variables | Types/Computations |
| :-------: | :------------------: |
| _recent_disconnect_count | increase(Counter) |
| _unalive_nodes_count | Gauge |
| _alive_nodes_count | Gauge |

<br/>

Following metrics are the member of `policy_context` and created at `policy_context::start()` ([meta_service.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_backup_service.cpp)). `policy_context` is created at `meta_service::start()` once cold backup is enabled:
| Variables | Types/Computations |
| :-------: | :------------------: |
| _counter_policy_recent_backup_duration_ms | Gauge |

<br/>

[...]

PR #1431, opened 2023-04-08
feat(new_metrics): add table-level metric entity and migrate table-level metrics for server_state of meta

empiredan:

In perf counters, all metrics of server_state are server-level, for example,
the number of healthy partitions among all tables of a pegasus cluster.

However, sometimes this is not enough. For example, the metric shows
that there are 4 unwritable partitions: the 4 unwritable partitions might
belong to different tables; or, they might belong to one table.

Therefore, these server-level metrics could be changed to table-level.
This will provide us with the status of each table. On the other hand,
once server-level metrics is needed, just aggregate on table-level ones.

The metrics of server_state that are migrated and changed to table-level
include: The number of dead, unreadable, unwritable, writable-ill, and
healthy partitions among all partitions of a table, the number of times
the configuration has been changed and the number of times the status
of partition has been changed to unwritable or writable for a table.

To implement table-level metrics, table-level metric entity is also added.


src/meta/table_metrics.cpp
```diff
@@ -0,0 +1,170 @@
...
+METRIC_DEFINE_entity(table);
+
+// The number of partitions in each status, see `health_status` and `partition_health_status()`
+// for details.
+
+METRIC_DEFINE_gauge_int64(
+    table,
+    dead_partitions,
+    dsn::metric_unit::kPartitions,
+    "The number of dead partitions among all tables, which means primary = 0 && secondary = 0");
```
acelyc111, line 35: `among all tables` -> `of this table`? Or just remove it because it is the table level metric.

src/meta/table_metrics.cpp
```diff
@@ -0,0 +1,170 @@
...
+    utils::auto_write_lock l(_lock);
+    _entities.clear();
+}
+
+bool operator==(const table_metric_entities &lhs, const table_metric_entities &rhs)
+{
+    utils::auto_read_lock l1(lhs._lock);
+    utils::auto_read_lock l2(rhs._lock);
+
+    if (lhs._entities.size() != rhs._entities.size()) {
```
acelyc111, line 148: How about short return true if their addreses are the same?

src/meta/table_metrics.cpp
empiredan: OK, I'll remove the redundant description.

src/meta/table_metrics.cpp
empiredan: Good idea !

acelyc111 (approved):

PR #1433, opened 2023-04-12
feat(Ranger): Use Apache Ranger for replica access controller

WHBANG:

This patch implements replica access controller using Ranger for ACL.

1. The Ranger policy info of the table is written to the app_envs of the table.
2. Support using the policy in the app_envs for ACL when the replica server
     reads and writes.
3. Modify some unit tests, and be compatible with old and new ACL.


src/replica/replica_stub.cpp
```diff
@@ -564,6 +566,7 @@ void replica_stub::initialize(bool clear /* = false*/)
     replication_options opts;
     opts.initialize();
     initialize(opts, clear);
+    _access_controller = dsn::make_unique<dsn::security::access_controller>();
```
acelyc111, line 569: Use std::make_unique

src/runtime/security/meta_access_controller.cpp
```diff
@@ -65,14 +65,10 @@ meta_access_controller::meta_access_controller(
                                         "RPC_CM_NOTIFY_STOP_SPLIT",
                                         "RPC_CM_QUERY_CHILD_STATE",
                                         "RPC_NEGOTIATION",
-                                        "RPC_CALL_RAW_MESSAGE",
```
acelyc111, line 68: Why remove these rpc codes?

src/runtime/security/replica_access_controller.h
```diff
@@ -30,16 +32,19 @@ namespace security {
...
+    void start_to_dump_and_sync_policies(const std::string &policies) override;
 
 private:
     utils::rw_lock_nr _lock; // [
-    std::unordered_set<std::string> _users;
+    std::unordered_set<std::string> _allowed_users;
     std::string _env_users;
     // ]
     std::string _name;
+    std::string _env_policies;
```
acelyc111, line 46: I guess it is used for fast judge to check if policies changed, could you add some necessary comments?

src/runtime/security/replica_access_controller.cpp
```diff
@@ -15,36 +15,61 @@
...
-    if (pre_check(user_name)) {
-        return true;
+    if (!FLAGS_enable_ranger_acl) {
+        if (!FLAGS_enable_acl || is_super_user(user_name)) {
+            return true;
+        }
+        {
+            utils::auto_read_lock l(_lock);
+            // If the user didn't specify any ACL, it means this table is publicly accessible to
+            // everyone. This is a backdoor to allow old-version clients to gracefully upgrade.After
```
acelyc111, line 58:
```suggestion
            // everyone. This is a backdoor to allow old-version clients to gracefully upgrade. After
```

src/runtime/security/replica_access_controller.cpp
```diff
@@ -15,36 +15,61 @@
...
+            // they are finally ensured to be fully upgraded, they can specify some usernames to ACL
+            // and the table will be truly protected.
+            if (!_allowed_users.empty() && _allowed_users.find(user_name) == _allowed_users.end()) {
```
acelyc111, line 61:
```suggestion
            if (_allowed_users.find(user_name) == _allowed_users.end()) {
```

src/runtime/security/meta_access_controller.cpp
WHBANG: coding error，duplicated in  input-parameter

src/runtime/security/replica_access_controller.cpp
WHBANG: The old ACL has left a backdoor for upgrade, and the `_allowed_users` will be cleared during the upgrade, so have to judge whether the list is empty, otherwise the back door will be gone. in other words, when the `_allowed_users` is empty, we want to return `true`

src/runtime/security/meta_access_controller.cpp
acelyc111: OK, let's adjust the rpc codes in lexicographical order to avoid these mistakes?

src/runtime/security/meta_access_controller.cpp
```diff
@@ -65,14 +65,10 @@ meta_access_controller::meta_access_controller(
...
                                         "RPC_CALL_RAW_SESSION_DISCONNECT",
-                                        "RPC_NFS_GET_FILE_SIZE",
```
acelyc111, line 70: How about this one? It's not duplicated.

src/runtime/security/replica_access_controller.cpp
```diff
@@ -62,10 +87,29 @@ void replica_access_controller::update_allowed_users(const std::string &users)
...
     {
         utils::auto_write_lock l(_lock);
-        // This swap operation is in constant time
-        _users.swap(users_set);
+        _allowed_users.swap(users_set);
         _env_users = users;
     }
 }
+
+void replica_access_controller::start_to_dump_and_sync_policies(const std::string &policies)
```
acelyc111, line 95: According to the function logic, how about just rename to `update_policies` ?

src/runtime/security/meta_access_controller.cpp
WHBANG: the RPC of nfs module is managed by the replica_access_controller, so remove it.

acelyc111 (dismissed):

foreverneverer (dismissed):

src/runtime/security/replica_access_controller.cpp
```diff
@@ -15,36 +15,61 @@
...
-replica_access_controller::replica_access_controller(const std::string &name) { _name = name; }
+DSN_DECLARE_bool(enable_acl);
+DSN_DECLARE_bool(enable_ranger_acl);
+replica_access_controller::replica_access_controller(const std::string &replica_name)
+{
+    _name = replica_name;
+}
 
-bool replica_access_controller::allowed(message_ex *msg)
+bool replica_access_controller::allowed(message_ex *msg, ranger::access_type req_type)
```
empiredan, line 48: Could this function be declared as `const` ? Could use `unordered_set::count` instead of `unordered_set::find`.

src/replica/replica.cpp
```diff
@@ -599,5 +599,11 @@ error_code replica::store_app_info(app_info &info, const std::string &path)
     return err;
 }
 
+bool replica::access_controller_allowed(message_ex *msg, ranger::access_type req_type)
```
empiredan, line 602: Declared as `const` ?

src/runtime/security/replica_access_controller.cpp
```diff
@@ -62,10 +88,37 @@ void replica_access_controller::update_allowed_users(const std::string &users)
...
+    dsn::json::json_forwarder<ranger::acl_policies>::decode(
+        dsn::blob::create_from_bytes(std::move(tmp_policies_str)), tmp_policies);
+    {
+        utils::auto_write_lock l(_lock);
+        _env_policies = policies;
+        _ranger_policies = std::move(tmp_policies);
     }
 }
+
+void replica_access_controller::check_allowed_users_valid()
```
empiredan, line 116:
```suggestion
void replica_access_controller::check_allowed_users_valid() const
```

src/runtime/security/replica_access_controller.h
```diff
@@ -30,14 +32,39 @@ namespace security {
...
+    // table changes
     void update_allowed_users(const std::string &users) override;
 
+    // Update '_ranger_policies' when the app_env(REPLICA_ACCESS_CONTROLLER_RANGER_POLICIES) of the
+    // table changes
+    void update_ranger_policies(const std::string &policies) override;
+
+private:
+    // Security check to avoid allowed_users is not empty in special scenarios.
+    void check_allowed_users_valid();
```
empiredan, line 51:
```suggestion
    void check_allowed_users_valid() const;
```

acelyc111 (approved):

empiredan (approved):

PR #1435, opened 2023-04-12
feat(new_metrics): add partition-level metric entity and migrate partition-level metrics for greedy_load_balancer of meta

empiredan:

In perf counters, all metrics of greedy_load_balancer are server-level, for
example, the number of each kind of operations by greedy balancer, including
moving primaries, copying primaries and copying secondaries.

For new metrics, it is hoped that they are fine-grained, since sometimes we
want to know which primaries are moved. Also, it is convenient to calculate
table-level or server-level metrics by just aggregate on partition-level ones.

The metrics of greedy_load_balancer that are changed to partition-level and
migrated to new framework include: the number of balance operations by
greedy balancer that are recently needed to be executed, move primaries,
copy primaries, and copy secondaries.

In addition to the metrics of greedy_load_balancer, we also change some
metrics of server_state again to partition-level which have been migrated
to table-level in https://github.com/apache/incubator-pegasus/pull/1431, 
for the reason that partition-level is considered more appropriate for them
than table-level.  The metrics changed to partition-level include the number
of times the configuration has been changed and the number of times the
status of partition has been changed to unwritable or writable for a partition.

To implement table-level metrics, partition-level metric entity is also added.


src/meta/table_metrics.h
```diff
@@ -21,35 +21,114 @@
...
+    __METRIC_DEFINE_INCREMENT_BY(greedy_move_primary_operations)
+    __METRIC_DEFINE_INCREMENT_BY(greedy_copy_primary_operations)
+    __METRIC_DEFINE_INCREMENT_BY(greedy_copy_secondary_operations)
+
+#undef __METRIC_DEFINE_INCREMENT_BY
+
+#define __METRIC_DEFINE_INCREMENT(name)                                                            \
+    void increment_##name(int32_t partition_id)                                                    \
+    {                                                                                              \
+        CHECK_LT(partition_id, _partition_metrics.size());                                         \
```
acelyc111, line 108: How about just `increment_##name##_by(partition_id, 1)`?

src/meta/table_metrics.cpp
```diff
@@ -17,12 +17,54 @@
...
+                      "The number of times the status of partition has been changed to unwritable");
+
+METRIC_DEFINE_counter(partition,
+                      writable_partition_changes,
+                      dsn::metric_unit::kChanges,
+                      "The number of times the status of partition has been changed to writable");
+
+METRIC_DEFINE_gauge_int64(
+    partition,
+    greedy_recent_balance_operations,
```
acelyc111, line 49: Is the `recent` word necessary?

src/meta/table_metrics.h
```diff
@@ -58,43 +137,55 @@ class table_metrics
...
+    struct partition_stats
+    {
+        int greedy_recent_balance_operations = 0;
+        int greedy_move_primary_operations = 0;
+        int greedy_copy_primary_operations = 0;
+        int greedy_copy_secondary_operations = 0;
+    };
 
-#define METRIC_CALL_TABLE_SET_METHOD(obj, name, table_id, value) (obj).set_##name(table_id, value)
+    using partition_map = std::unordered_map<gpid, partition_stats>;
```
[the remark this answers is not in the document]
empiredan: Means using 2-dimensional `std::unordered_map<int32_t, std::unordered_map<int32_t, partition_stats>` instead ?

src/meta/table_metrics.cpp
empiredan: Yes, actually this metric is updated after a batch of balance operations. Thus it is `recent`.

Also, it is somewhat special. It could be either the number of balance operations that have been executed, or the  number of balance operations that would be executed in the future. As a result it can only be declared as gauge rather than counter.

src/meta/table_metrics.h
```diff
@@ -21,35 +21,114 @@
...
+    void resize_partitions(int32_t partition_count);
+
+    METRIC_DEFINE_SET(dead_partitions, int64_t)
+    METRIC_DEFINE_SET(unreadable_partitions, int64_t)
+    METRIC_DEFINE_SET(unwritable_partitions, int64_t)
+    METRIC_DEFINE_SET(writable_ill_partitions, int64_t)
+    METRIC_DEFINE_SET(healthy_partitions, int64_t)
+
+#define __METRIC_DEFINE_INCREMENT_BY(name)                                                         \
+    void increment_by_##name(int32_t partition_id, int64_t x)                                      \
```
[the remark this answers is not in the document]
empiredan: OK.

acelyc111 (approved):

acelyc111 (approved):

PR #1438, opened 2023-04-14
feat(new_metrics): add backup-policy-level metric entity and migrate backup-policy-level metrics for meta_backup_service

empiredan:

In perf counters, there's only one metric for meta_backup_service, namely
the recent backup duration for each policy, which means this metric is
policy-level. Therefore policy-level entity would also be implemented in
new metrics.

### DIFF
--- a/src/meta/meta_backup_service.cpp
+++ b/src/meta/meta_backup_service.cpp
@@ -49,6 +49,7 @@
 #include "utils/chrono_literals.h"
 #include "utils/flags.h"
 #include "utils/fmt_logging.h"
+#include "utils/string_view.h"
 #include "utils/time_utils.h"
 
 METRIC_DEFINE_entity(backup_policy);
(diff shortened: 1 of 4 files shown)


src/meta/meta_backup_service.cpp
```diff
@@ -830,12 +862,8 @@ void policy_context::start()
         continue_current_backup_unlocked();
     }
 
-    std::string counter_name = _policy.policy_name + ".recent.backup.duration(ms)";
-    _counter_policy_recent_backup_duration_ms.init_app_counter(
-        "eon.meta.policy",
-        counter_name.c_str(),
-        COUNTER_TYPE_NUMBER,
-        "policy recent backup duration time");
+    CHECK(!(_policy.policy_name.empty()), "policy_name should has been initialized");
```
acelyc111, line 865:
```suggestion
    CHECK(!_policy.policy_name.empty(), "policy_name should has been initialized");
```

acelyc111 (approved):

Issue #1441, opened 2023-04-16
Feature(new_metrics): migrate replica-level metrics for pegasus_manual_compact_service


empiredan:
Following metrics are the members of `pegasus_manual_compact_service` ([pegasus_manual_compact_service.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/server/pegasus_manual_compact_service.cpp)), which is created at the construction of `pegasus_server_impl`:
| Variables | Types/Computations |
| :-------: | :------------------: |
| _pfc_manual_compact_enqueue_count | Gauge |
| _pfc_manual_compact_running_count | Gauge |

PR #1443, opened 2023-04-17
feat(new_metrics): migrate replica-level metrics for pegasus_manual_compact_service

empiredan:
https://github.com/apache/incubator-pegasus/issues/1441

In perf counters, all of the 2 metrics of `pegasus_manual_compact_service`
are about the numbers of tasks of rocksdb manual compaction: one is  the
number of current queued tasks, while another is the number of current
running tasks. Both metrics are server-level.

They would become replica-level after migrating to the new metrics, based
on which server-level ones could also be achieved. A convenient class
`auto_count` is also provided to increment gauge that will be decremented
automatically at the end of the scope.


src/utils/test/metrics_test.cpp
```diff
@@ -3195,4 +3232,6 @@ TEST_F(MetricVarTest, AutoLatencyMilliSeconds) { TEST_METRIC_VAR_AUTO_LATENCY(ms
 
 TEST_F(MetricVarTest, AutoLatencySeconds) { TEST_METRIC_VAR_AUTO_LATENCY(s, 1000 * 1000 * 1000); }
 
+TEST_F(MetricVarTest, AutoCount) { test_auto_count(); }
```
acelyc111, line 3235:
```suggestion
TEST_F(MetricVarTest, AutoCount) { ASSERT_NO_FATAL_FAILURE(test_auto_count()); }
```

PR #1444, opened 2023-04-17
fix(go-client): loopForRequest not return and retry forever

### LINKED ISSUE #1385: The client does not respond after the  replication  service is restarted
## Bug Report

集群配置: 
1. 共 4 台服务器
2. 其中 3 台部署  meta , 
3. 每台服务器都部署 1 个 replic

问题场景:
1. 客户端使用 golang
2. 端从 kafka 中读取数据, 写入 pegasus
3. 当重启一台 replic  服务的时候

4 golang  客户端打印
time="2023-03-07T07:56:34Z" level=info msg="session [10.218.73.124:34801(replica)] is closed by the peer" func="session.(*nodeSession).loopForResponse" file="session.go:287"

replic重启成功(1 分钟左右)

但是 client 程序一致僵死, 没有写的数据写入
拜读源码,
发现一旦发生 rpc 网络关闭问题, 
session 就关闭退出了, 
当前协程不再处理任何事情

问题出现在 session.go 的 286 行, 
```
if rpc.IsNetworkClosed(err) { // EOF
    n.logger.Printf("session %s is closed by the peer", n)
   return nil
}
```

littlepangdi:
fix #1385 

#### expected behavior: 
after replica server restart，The client should be able to reconnect with node without considering the client configuration update (reselecting the primary replica).

Actually, client configuration update is another issue we should talk about. I'd like to discuss that in another issue.

In this pr, we focus on why client cannot reconnect with restarted server.

1. SDK use several loop to monitor rpc.conn, when server is closed, `loopForDialing` will continue to retry dialling until connected with server(when server is restart),then,  two new loop `loopForRequest` & `loopForResponse`  will be created to handle request.
however,  `loopForRequest` will not return when correlative `loopForResponse`  returned because of IsNetworkClosed(EOF), since latter only return nil and will not shutdown tom.  thus, there will be more alive`loopForRequest` than `loopForResponse`  in this case.

2. SDK retried not on timeout err, however, we wrapped timeout err incorrectly in here https://github.com/apache/incubator-pegasus/blob/d16e65cd88b4fb6113d5e6a99234b5be527f28da/go-client/session/session.go#L338 ,thus every request like `multiset` will continue to do retry here since err we passed to outer is no longer timeout err
https://github.com/apache/incubator-pegasus/blob/d16e65cd88b4fb6113d5e6a99234b5be527f28da/go-client/pegasus/table_connector.go#L719

### Checklist 
monitor capture based on this pr:
timeout is eliminated after server restart.
image

### DIFF
--- a/go-client/pegasus/table_connector.go
+++ b/go-client/pegasus/table_connector.go
@@ -724,6 +724,7 @@ func (p *pegasusTableConnector) handleReplicaError(err error, replica *session.R
 			err = errors.New(err.Error() + " Rate of requests exceeds the throughput limit")
 		case base.ERR_INVALID_STATE:
 			err = errors.New(err.Error() + " The target replica is not primary")
+			retry = false
 		case base.ERR_OBJECT_NOT_FOUND:
 			err = errors.New(err.Error() + " The replica server doesn't serve this partition")
 		case base.ERR_SPLITTING:
--- a/go-client/pegasus/table_connector_test.go
+++ b/go-client/pegasus/table_connector_test.go
@@ -261,7 +261,7 @@ func TestPegasusTableConnector_TriggerSelfUpdate(t *testing.T) {
 	<-ptb.confUpdateCh
 	assert.Error(t, err)
 	assert.True(t, confUpdate)
-	assert.True(t, retry)
+	assert.False(t, retry)
 
 	confUpdate, retry, err = ptb.handleReplicaError(base.ERR_PARENT_PARTITION_MISUSED, nil)
 	<-ptb.confUpdateCh
--- a/go-client/session/session.go
+++ b/go-client/session/session.go
@@ -226,8 +226,16 @@ func (n *nodeSession) notifyCallerAndDrop(req *requestListener) {
 // single-routine worker used for sending requests.
 // Any error occurred will end up this goroutine as well as the connection.
 func (n *nodeSession) loopForRequest() error { // no error returned actually
+	//add ticker to trigger loop return
+	// since if correlative loopForResponse returned because of IsNetworkClosed(EOF),
+	// this loop will not receive any signal and runs forever
+	ticker := time.NewTicker(1 * time.Millisecond)
 	for {
 		select {
+		case <-ticker.C:
+			if n.conn.GetState() != rpc.ConnStateReady {
+				return nil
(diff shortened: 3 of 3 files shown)


littlepangdi:
During test, we found that there is still timeout exists, thus locate another problem--sdk should not retry on `replica is not primary error` after reconnect with restarted server ip.
In this case, configuration update should be processed and keep retrying means nothing.

littlepangdi:
Futhermore, if we remove user-side call of `TableConnector.Close()`， client is working perfectly even when server replica is stopped as if nothing happened. see for this issue #1446 
image

log:
triggers configuration update(ips are blurred):
```
time="2023-04-17T18:47:06+08:00" level=info msg="trigger configuration update of table [CL74485_loki_staging2_324] due to RPC failure [session unresponsive for long] to [xx(replica)]" func="pegasus.(*pegasusTableConnector).tryConfUpdate" file="table_connector.go:753"
time="2023-04-17T18:47:06+08:00" level=info msg="querying configuration of table(CL74485_loki_staging2_324) [metaList=[xxxx]]" func="session.(*MetaManager).QueryConfig" file="meta_session.go:110"
time="2023-04-17T18:47:06+08:00" level=info msg="querying configuration of table(CL74485_loki_staging2_324) from [xxxx(meta)]" func="session.(*metaSession).queryConfig" file="meta_session.go:45"
time="2023-04-17T18:47:06+08:00" level=info msg="create session with [xxxx(replica)]" func=session.newNodeSession file="session.go:141"
time="2023-04-17T18:47:13+08:00" level=info msg="failed to read response from [xxxx(replica)]: read tcp xxxx:46252->xxxx:15801: read: connection reset by peer" func="session.(*nodeSession).loopForResponse" file="session.go:298"
time="2023-04-17T18:47:14+08:00" level=info msg="dial to [xxxx(replica)]" func="session.(*nodeSession).dial" file="session.go:199"
time="2023-04-17T18:47:14+08:00" level=info msg="stop dialing for [xxxx(replica)], connection state: ConnStateReady" func="session.(*nodeSession).dial" file="session.go:211"
```

hycdong (approved):
Thanks for the contribution~

acelyc111 (approved):

PR #1447, opened 2023-04-17
fix: Fault-tolerant storage engine errors for read operations

acelyc111:

ReplicaServer doesn't handle the error returned from storage engine, thus
even if the storage engine is corrupted, the server doesn't recognize these
situactions, and still running happily. However, the client always gets an
error status.
This situaction will not recover automatically except stopping the server
and moving away the corrupted RocksDB directories manually.

This patch handle the kCorruption error returned from storage engine, then
close the replcia, move the directory to ".err" trash path. The replica is
able to recover automatically (if RF > 1).


src/runtime/serverlet.h
```diff
@@ -90,6 +90,8 @@ class rpc_replier
         return _response != nullptr ? _response->to_address : rpc_address::s_invalid_address;
     }
 
+    int error_code() const { return _response != nullptr ? _response->error() : 0; }
```
empiredan, line 93: For `rpc_replier`, the reason why we just need the error code in `message_ex`, rather than field `error` in thrift object `TResponse`, is that there is no read api (`on_get`, `on_multi_get`, ...) using `rpc_replier` to reply ?

acelyc111: Only simple_kv use this register function, simple_kv is just a simple demo of rdsn framework, so I'll update the code to limit the usage of function
```
template <typename TReq, typename TResp>
static bool register_async_rpc_handler(...)
```

levy5307 (approved):

empiredan (approved):

PR #1448, opened 2023-04-17
doc(go-client): add deprecated marker to TableConnector.Close()

### LINKED ISSUE #1446: Add deprecate to go-client TableConnector.Close() to avoid misleading users
## Feature Request

**Is your feature request related to a problem? Please describe:**

I'd like to propose we add Deprecated to `TableConnector.Close()` for two reasons:

1. Action by `pegasus.Client` indicates that users  don't have to call `TableConnector.Close()`.

We use golang-client to manage pegasus tables in our project(Loki). 
Normally, after create any connection, client should manully call `defer conn.Close()` to avoid any connection leakage  if `Close()` exists.

For example:
```
table, err := c.pegasusDB.GetClient().OpenTable(goCtx, tableName)
if err != nil {
	//close anyway to make sure session released
	if table != nil {
		table.Close()
	}
	errChan <- err
	return
}
defer table.Close()
//do something with table
table.MultiSet(timeoutCtx, []byte(hashKey), entry.SortKeys, entry.Values)

```
However, in golang-client sdk, pegasus.Client takeover this job for `TableConnector` and calls it in `Client.Close()`https://github.com/apache/incubator-pegasus/blob/1beb24a765ba009499a6348f82f1ea95ca2a663d/go-client/pegasus/client.go#L82

2. This method itself indicates `TableConnector.Close()` should not be called by users.

In golang-client sdk, `TableConnector.Close()` will  kill tom and destroy all related goroutine, including `loopForAutoUpdate()`.  This causes disaster in our case, config update will be triggered but has no time to be processed before it's killed. 
Overall, it makes `TableConnector.Close()` a danger operation since if error like server crushed or restarted occurred, we'd hope config be updated other than reopen an old connection to the wrong server.

**Describe the feature you'd like:**

Add Deprecated to `TableConnector.Close()`

**Describe alternatives you've considered:**

hide `TableConnector.Close()` by changing `Close()` to `close()` since it's in-package call only.

littlepangdi:
fix #1446 

##### Tests 

- Unit test
- Integration test
- No code

##### Code changes

- Has persistent data change

##### Side effects

- Increased code complexity

##### Related changes

### DIFF
--- a/go-client/pegasus/table_connector.go
+++ b/go-client/pegasus/table_connector.go
@@ -192,6 +192,8 @@ type TableConnector interface {
 	// NOTE: this operation is not guaranteed to be atomic
 	BatchGet(ctx context.Context, keys []CompositeKey) (values [][]byte, err error)
 
+	// Deprecated: danger operation: table is managed by client and should only be Close once when client is exiting.
+	// this operation will kill all table-related loops
 	Close() error
 }
 


littlepangdi:
This pr is also related to #1385 , normally, when replica server is restarted , client should update config and build connect with another server ip.
I suspect that the reason config is not updated successful is that `TableConnector.Close()` is called  and `loopForAutoUpdate()` is interrupted.

hycdong (approved):

acelyc111 (approved):

PR #1452, opened 2023-04-18
feat(Ranger): Use Apache Ranger for ACL when replica performs nfs copy

WHBANG:

This patch add ACL to the NFS copy of replica.

1. Added `gpid` info to the data structure defined in `nfs.thrift`
2. Perform ACL through the Ranger policy matched by `gpid`
3. The registration of nfs is moved to `replica_stub.cpp`, and the original registration 
    information is retained, which is convenient for testing

### DIFF
--- a/src/replica/replica_stub.h
+++ b/src/replica/replica_stub.h
@@ -277,7 +277,7 @@ public:
     template <typename TReqType, typename TRespType>
     bool check_status_and_authz_with_reply(const TReqType &request,
                                            ::dsn::rpc_replier<TRespType> &reply,
-                                           const ::dsn::ranger::access_type &ac_type)
+                                           const ::dsn::ranger::access_type &ac_type) const
     {
         if (!_access_controller->is_enable_ranger_acl()) {
             return true;
--- a/src/replica/replica_stub.h
+++ b/src/replica/replica_stub.h
@@ -55,6 +55,7 @@
 #include "dsn.layer2_types.h"
 #include "failure_detector/failure_detector_multimaster.h"
 #include "metadata_types.h"
+#include "nfs/nfs_node.h"
 #include "partition_split_types.h"
 #include "perf_counter/perf_counter_wrapper.h"
 #include "replica.h"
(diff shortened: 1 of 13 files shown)


src/replica/replica_stub.h
```diff
@@ -270,6 +274,38 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
+        replica_ptr rep = get_replica(pid);
+
+        if (!rep) {
+            TRespType resp;
+            resp.error = ERR_OBJECT_NOT_FOUND;
+            reply(resp);
+            return false;
+        }
+        dsn::message_ex *msg = reply.response_message();
+        if (!rep->access_controller_allowed(msg, ranger::access_type::kWrite)) {
```
acelyc111, line 294: According to the function name, we can't infer it is used for write or read RPCs, so it's not enough to judge it's write allowed only.

acelyc111 (dismissed):
Good job, thanks for the contribution！

src/replica/replica_stub.h
```diff
@@ -270,6 +274,39 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
 
     fs_manager *get_fs_manager() { return &_fs_manager; }
 
+    template <typename TReqType, typename TRespType>
+    bool check_status_and_authz_with_reply(const TReqType &request,
+                                           ::dsn::rpc_replier<TRespType> &reply,
+                                           const ::dsn::ranger::access_type &ac_type)
```
empiredan, line 280:
```suggestion
                                           const ::dsn::ranger::access_type &ac_type) const
```

src/replica/replica_stub.h
```diff
@@ -270,6 +274,39 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
+    {
+        if (!_access_controller->is_enable_ranger_acl()) {
```
empiredan, line 282: `is_enable_ranger_acl()` could also be declared as const ?

acelyc111: It is const.

acelyc111 (approved):

empiredan (approved):

Issue #1454, opened 2023-04-18
Feature(new_metrics): migrate metrics for replica_stub


empiredan:
Following metrics are the members of `replica_stub` ([replica_stub.cpp](https://github.com/apache/incubator-pegasus/blob/master/src/replica/replica_stub.cpp)), which is created at the construction of `replication_service_app`:
| Variables | Types/Computations |
| :-------: | :------------------: |
| _counter_replicas_count | Gauge |
| _counter_replicas_opening_count | Gauge |
| _counter_replicas_closing_count | Gauge |
| _counter_replicas_commit_qps | rate(Counter) |
| _counter_replicas_learning_count | Gauge |
| _counter_replicas_learning_max_duration_time_ms | Gauge |
| _counter_replicas_learning_max_copy_file_size | Gauge |
| _counter_replicas_learning_recent_start_count | increase(Counter) |
| _counter_replicas_learning_recent_round_start_count | increase(Counter) |
| _counter_replicas_learning_recent_copy_file_count | increase(Counter) |
| _counter_replicas_learning_recent_copy_file_size | increase(Counter) |
| _counter_replicas_learning_recent_copy_buffer_size | increase(Counter) |
| _counter_replicas_learning_recent_learn_cache_count | increase(Counter) |
| _counter_replicas_learning_recent_learn_app_count | increase(Counter) |
| _counter_replicas_learning_recent_learn_log_count | increase(Counter) |
| _counter_replicas_learning_recent_learn_reset_count | increase(Counter) |
| _counter_replicas_learning_recent_learn_fail_count | increase(Counter) |
| _counter_replicas_learning_recent_learn_succ_count | increase(Counter) |
| _counter_replicas_recent_prepare_fail_count | increase(Counter) |
| _counter_replicas_recent_replica_move_error_count | increase(Counter) |
| _counter_replicas_recent_replica_move_garbage_count | increase(Counter) |
| _counter_replicas_recent_replica_remove_dir_count | increase(Counter) |
| _counter_replicas_error_replica_dir_count | Gauge |
| _counter_replicas_garbage_replica_dir_count | Gauge |
| _counter_replicas_tmp_replica_dir_count | Gauge |
| _counter_replicas_origin_replica_dir_count | Gauge |
| _counter_replicas_recent_group_check_fail_count | increase(Counter) |
| _counter_shared_log_size | Gauge |
| _counter_shared_log_recent_write_size |increase(Counter) |
[...]

PR #1455, opened 2023-04-18
feat(new_metrics): migrate server-level metrics for replica_stub (part 1)

empiredan:

### DIFF
--- a/src/replica/replica.cpp
+++ b/src/replica/replica.cpp
@@ -230,9 +230,9 @@ void replica::update_last_checkpoint_generate_time()
 // Statistics //
 //            //
 
-void replica::update_commit_qps(int count)
+void replica::update_commit_requests(int count)
 {
-    _stub->_counter_replicas_commit_qps->add((uint64_t)count);
+    METRIC_INCREMENT_BY(*_stub, committed_requests, count);
 }
 
 void replica::init_state()
--- a/src/replica/replica.h
+++ b/src/replica/replica.h
@@ -284,7 +284,7 @@ public:
     //
     // Statistics
     //
-    void update_commit_qps(int count);
+    void update_commit_requests(int count);
 
     // routine for get extra envs from replica
     const std::map<std::string, std::string> &get_replica_extra_envs() const { return _extra_envs; }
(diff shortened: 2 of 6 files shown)


src/replica/replica.cpp
```diff
@@ -229,11 +228,7 @@ void replica::update_last_checkpoint_generate_time()
 //            //
 // Statistics //
 //            //
-
-void replica::update_commit_qps(int count)
-{
-    _stub->_counter_replicas_commit_qps->add((uint64_t)count);
-}
+METRIC_IMPL_INCREMENT_BY(*_stub, committed_requests, replica::)
```
acelyc111, line 231: It‘s a bit of strange to pass a namesapce as a "parameter", would it better to define functions manually?

src/replica/replica.h
```diff
@@ -284,7 +284,7 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
     //
     // Statistics
     //
-    void update_commit_qps(int count);
+    METRIC_DECLARE_INCREMENT_BY(committed_requests, );
```
acelyc111, line 287: It's strange too to pass such style of parameters.

src/replica/replica_stub.cpp
```diff
@@ -2032,10 +2042,10 @@ task_ptr replica_stub::begin_open_replica(
         if (rep->status() == partition_status::PS_INACTIVE && tsk->cancel(false)) {
             // reopen it
             _closing_replicas.erase(it);
-            _counter_replicas_closing_count->decrement();
+            METRIC_VAR_DECREMENT(closing_replicas);
```
acelyc111, line 2045: If needed, it's better to consider define a new metric type `FunctionGauge`, ref https://github.com/apache/kudu/blob/master/src/kudu/util/metrics.h#L1223

We can improve it in future refactoring.

src/replica/replica.h
empiredan: OK, `committed_requests` will be moved to replica-level and maintained in `replication_app_base`, which is more reasonable.

src/replica/replica_stub.cpp
empiredan: OK, I'll add another issue to track this.

acelyc111 (approved):

PR #1459, opened 2023-04-25
feat(new_metrics): migrate metrics for replica_stub (part 2)

empiredan:

This is the 2nd part of migrating metrics of `replica_stub` to new framework,
all of which are learn-related.

During this migration, there are 3 metrics still keeping server-level, including
the number of learning replicas, the max duration and the max size of files that
are copied from learnee among all learning replicas.

Another 11 metrics are changed from server-level to replica-level, since they
should be observed for each replica. All of them are observed from the view
of learners, namely the potential secondary replica. The learnee is the primary
replica. These metrics include: the number of learns launched by learner, the
number of learn rounds launched by learner (during a learn there might be
multiple rounds, the number of files that are copied from learnee, the size of
files that are copied from learnee, the size of data that are copied from learnee's
buffer, the number of learn responses of `LT_CACHE`, `LT_APP` and `LT_LOG`
type decided by learner with each learn response related to an `RPC_LEARN`
request, the number of times learner resets its local state (since its local state
is newer than learnee's) with each reset related to an learn response of an
`RPC_LEARN` request, the number of failed and successful learns launched by
learner.

### DIFF
--- a/src/replica/replica_stub.cpp
+++ b/src/replica/replica_stub.cpp
@@ -121,9 +121,9 @@ METRIC_DEFINE_gauge_int64(server,
                           "The max duration among all learning replicas");
 
 METRIC_DEFINE_gauge_int64(server,
-                          learning_replicas_max_duration_ms,
-                          dsn::metric_unit::kMilliSeconds,
-                          "The max duration among all learning replicas");
+                          learning_replicas_max_copy_bytes,
+                          dsn::metric_unit::kBytes,
+                          "The max size of data that are copied from remote among all learning replicas");
 
 namespace dsn {
 namespace replication {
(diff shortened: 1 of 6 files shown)


src/replica/replica_stub.cpp
```diff
@@ -1972,9 +1920,9 @@ void replica_stub::on_gc()
         }
     }
 
-    _counter_replicas_learning_count->set(learning_count);
-    _counter_replicas_learning_max_duration_time_ms->set(learning_max_duration_time_ms);
-    _counter_replicas_learning_max_copy_file_size->set(learning_max_copy_file_size);
+    METRIC_VAR_SET(learning_replicas, learning_count);
+    METRIC_VAR_SET(learning_replicas_max_duration_ms, learning_max_duration_time_ms);
```
acelyc111, line 1924: How about use percentile type of metric for 'duration', then it's able to provide more distribution details.

acelyc111 (approved):

PR #1463, opened 2023-04-27
feat(new_metrics): migrate metrics for replica_stub (part 4)

empiredan:

This is the 4th part of migrating metrics of replica_stub to new framework.

### DIFF
--- a/src/replica/duplication/duplication_sync_timer.cpp
+++ b/src/replica/duplication/duplication_sync_timer.cpp
@@ -69,7 +69,7 @@ void duplication_sync_timer::run()
         if (!confirmed.empty()) {
             req->confirm_list[r->get_gpid()] = std::move(confirmed);
         }
-        METRIC_SET(*(r->get_duplication_manager()), dup_pending_mutations);
+        METRIC_SET(*r, dup_pending_mutations);
     }
 
     duplication_sync_rpc rpc(std::move(req), RPC_CM_DUPLICATION_SYNC, 3_s);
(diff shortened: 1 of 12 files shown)


src/replica/duplication/duplication_sync_timer.cpp
```diff
@@ -65,15 +64,13 @@ void duplication_sync_timer::run()
...
 
     // collects confirm points from all primaries on this server
-    uint64_t pending_muts_cnt = 0;
     for (const replica_ptr &r : get_all_primaries()) {
         auto confirmed = r->get_duplication_manager()->get_duplication_confirms_to_update();
         if (!confirmed.empty()) {
             req->confirm_list[r->get_gpid()] = std::move(confirmed);
         }
-        pending_muts_cnt += r->get_duplication_manager()->get_pending_mutations_count();
+        METRIC_SET(*(r->get_duplication_manager()), dup_pending_mutations);
```
acelyc111, line 72: Now that we can get the replica the metric belongs to, why not set the metric via `replica_ptr` directly?

empiredan: OK, I'll add a method for `replica` class.

src/replica/replica_stub.cpp
```diff
@@ -2901,7 +2895,7 @@ uint64_t replica_stub::gc_tcmalloc_memory(bool release_all)
             release_bytes -= 1024 * 1024;
         }
     }
-    _counter_tcmalloc_release_memory_size->set(tcmalloc_released_bytes);
+    METRIC_VAR_SET(tcmalloc_released_bytes, tcmalloc_released_bytes);
```
acelyc111, line 2898: @empiredan As we discuss offline, `tcmalloc_released_bytes` is better to be a counter type, we can improve it in next patch.

acelyc111 (approved):

src/replica/replica_stub.cpp
empiredan: OK, I'll fix this in the next PR.

PR #1474, opened 2023-05-11
feat(new_metrics): migrate metrics for replica_stub (part 6)

empiredan:

This is the 6th part of migrating metrics of replica_stub to new framework,
all of which are bulk-load-related.

During this migration, there are 7 metrics which are changed from server-level
to replica-level, including the number of downloading/ingesting bulk loads, the
number of successful/failed bulk loads, the number of files that have been
downloaded successfully or have failed to be downloaded for bulk loads, the
size of files that have been downloaded successfully for bulk loads.

Another 3 metrics are still kept server-level, including the number of current
running bulk loads, the max duration of ingestions or the whole bulk loads.

### DIFF
--- a/src/replica/bulk_load/replica_bulk_loader.cpp
+++ b/src/replica/bulk_load/replica_bulk_loader.cpp
@@ -93,14 +93,16 @@ class block_filesystem;
 namespace replication {
 
 replica_bulk_loader::replica_bulk_loader(replica *r)
-    : replica_base(r), _replica(r), _stub(r->get_replica_stub()),
-    METRIC_VAR_INIT_replica(bulk_load_downloading_count),
-    METRIC_VAR_INIT_replica(bulk_load_ingestion_count),
-    METRIC_VAR_INIT_replica(bulk_load_successful_count),
-    METRIC_VAR_INIT_replica(bulk_load_failed_count),
-    METRIC_VAR_INIT_replica(bulk_load_download_file_successful_count),
-    METRIC_VAR_INIT_replica(bulk_load_download_file_failed_count),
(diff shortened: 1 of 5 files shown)


src/replica/replica_stub.cpp
```diff
@@ -191,6 +191,21 @@ METRIC_DEFINE_counter(server,
                       dsn::metric_unit::kRequests,
                       "The number of busy write requests");
 
+METRIC_DEFINE_gauge_int64(server,
+                          bulk_load_running_count,
+                          dsn::metric_unit::kBulkLoads,
+                          "The number of current running bulk loads");
+
+METRIC_DEFINE_gauge_int64(server,
```
acelyc111, line 199: For the xxx_max_xxx mastrics, how about use percentile? we can do that in next patches.

src/replica/bulk_load/replica_bulk_loader.cpp
```diff
@@ -50,6 +48,41 @@
 #include "utils/string_view.h"
 #include "utils/thread_access_checker.h"
 
+METRIC_DEFINE_counter(replica,
+                      bulk_load_download_count,
```
acelyc111, line 52: Like `bulk_load_running_count`, how about rename the following two metrics to `bulk_load_downloading_count` and `bulk_load_ingesting_count` ?

And in fact they are gauge type?

empiredan: OK. Actually in perf counter they are accumulative counter. We could observe their increase. We could also change them to gauge, by decreasing them while finished. Is this OK ?

acelyc111 (approved):

PR #1478, opened 2023-05-17
feat(Ranger): global resource removes LIST access control and create a default database policy for the legacy table

WHBANG:

1. The access control management of RPC `RPC_CM_LIST_APPS` is removed from the global level resource, which is managed by the database resource.
2. Provide a default database-level policy name for access control to legacy tables.
3. Modified some ut.
4. Refactored Ranger policy access logic, as shown below:
企业微信截图_b1d5aedf-0e8c-40d0-92d9-cad386e6ab6e


acelyc111:
Could you please draw the flow-process diagram in raw text? It's a part of the code to understand the backgroud, ref: https://github.com/apache/incubator-pegasus/blob/master/src/server/hotkey_collector.h#L61

src/runtime/ranger/ranger_resource_policy.cpp
```diff
@@ -27,44 +27,155 @@ bool policy_item::match(const access_type &ac_type, const std::string &user_name
...
+    }
+    return do_policies_check(check_type, ac_type, user_name, deny_policies, deny_policies_exclude);
+}
+
+policy_check_status
+acl_policies::do_policies_check(const policy_check_type &check_type,
+                                const access_type &ac_type,
+                                const std::string &user_name,
+                                const std::vector<policy_item> &policies,
+                                const std::vector<policy_item> &policies_exclude) const
```
acelyc111, line 46: How about renaming to `exclude_policies` to make it as a noun ?

src/runtime/ranger/ranger_resource_policy.cpp
```diff
@@ -27,44 +27,155 @@ bool policy_item::match(const access_type &ac_type, const std::string &user_name
...
+}
 
-    // 2. Check if it is allowed.
-    for (const auto &allow_policy : allow_policies) {
-        // 2.1. In 'allow_policies'.
-        if (!allow_policy.match(ac_type, user_name)) {
+bool check_ranger_resource_policy_allowed(const std::vector<ranger_resource_policy> &policies,
+                                          const access_type &ac_type,
+                                          const std::string &user_name,
+                                          bool is_need_match_database,
```
acelyc111, line 81:
```suggestion
                                          bool need_match_database,
```

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -68,8 +100,36 @@ struct ranger_resource_policy
...
+struct matched_database_table_policy
+{
+    std::string matched_database_name;
+    std::string matched_table_name;
+    acl_policies policies;
+
+    DEFINE_JSON_SERIALIZATION(matched_database_name, matched_table_name, policies);
+};
+
+// Returns true if 'policies' allows "user_name" to access "database_name" via "rpc_code".
```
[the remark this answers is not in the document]
acelyc111: Where is `rpc_code`?

src/runtime/ranger/ranger_resource_policy.h
```diff
@@ -68,8 +100,36 @@ struct ranger_resource_policy
...
+// 'is_need_match_database' being true means that the 'policies' needs to be matched to the database
+// first, false means not
+bool check_ranger_resource_policy_allowed(const std::vector<ranger_resource_policy> &policies,
+                                          const access_type &ac_type,
+                                          const std::string &user_name,
+                                          bool is_need_match_database,
```
acelyc111, line 123: As I've ever mentioned, it would be better to use enum rather than boolean.

Further more, add some comments to explain when needs to match database, and when doesn't need.

acelyc111:
implemented by other PRs

PR #1480, opened 2023-05-17
refactor: remove some useless code

acelyc111:

Remove some useless and duplicate code.


src/server/test/pegasus_server_test_base.h
```diff
@@ -77,14 +76,14 @@ class pegasus_server_test_base : public ::testing::Test
...
+        delete _replica_stub;
+        delete _replica;
     }
 
 protected:
     std::unique_ptr<mock_pegasus_server_impl> _server;
-    dsn::replication::replica *_replica;
-    dsn::replication::replica_stub *_replica_stub;
+    dsn::replication::replica *_replica = nullptr;
+    dsn::replication::replica_stub *_replica_stub = nullptr;
```
levy5307, line 86: could you change tye type of `_replica_stub` to `unique_ptr<replica_stub>`?

acelyc111: Good idea, but it will introduce more code changes if using unique_ptr. std::unique_ptr is not able to access the class replica's private constructor member.
Because this is a patch to "remove codes", so we can do it in another PR.

empiredan (approved):

Issue #1481, opened 2023-05-18
Feature(new_metrics): migrate metrics for some duplication class


empiredan:
Migrate some duplication-related classes, including `ship_mutation`, `mutation_buffer`, `load_from_private_log`, etc.

PR #1482, opened 2023-05-18
feat(new_metrics): migrate metrics for some duplication class

empiredan:
https://github.com/apache/incubator-pegasus/issues/1481

Some duplication-related classes, including `ship_mutation`, `mutation_buffer`,
`load_from_private_log`, are migrated to the new framework.

During this migration, there are 6 metrics which are changed from server-level
to replica-level, all of which are duplication-related, including the shipped size of
private log, the number of times private log files have failed to be loaded, the bytes
of mutations that have been skipped due to failed loadings of private log files, the
size read from private log, the number of mutations read from private log, the number
of lost mutations recently.

Another 2 metrics, the numbers of failed/successful DUPLICATE requests sent from
client, are renamed according to the new style of naming for duplication.

### DIFF
--- a/src/replica/duplication/mutation_batch.h
+++ b/src/replica/duplication/mutation_batch.h
@@ -27,6 +27,7 @@
 #include "replica/prepare_list.h"
 #include "replica/replica_base.h"
 #include "utils/errors.h"
+#include "utils/metrics.h"
 
 namespace dsn {
 namespace replication {
--- a/src/replica/duplication/mutation_batch.h
+++ b/src/replica/duplication/mutation_batch.h
@@ -22,7 +22,6 @@
 #include <memory>
 
 #include "common/replication_other_types.h"
-#include "perf_counter/perf_counter_wrapper.h"
 #include "replica/duplication/mutation_duplicator.h"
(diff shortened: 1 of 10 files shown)


src/replica/duplication/mutation_batch.h
```diff
@@ -45,7 +45,7 @@ class mutation_buffer : public prepare_list
     void commit(decree d, commit_type ct) override;
 
 private:
-    perf_counter_wrapper _counter_dulication_mutation_loss_count;
+    METRIC_VAR_DECLARE_gauge_int64(dup_recent_lost_mutations);
```
acelyc111, line 48: Is the "recent" word necesary?

empiredan: Yes, actually this metric is `set` with the number of lost mutations, thus it's `recent`.

acelyc111: To record all change values, will would be great to use counter? we can improve it in the future.

empiredan: OK.

Issue #1483, opened 2023-05-22
Feature(new_metrics): migrate metrics for task queue


empiredan:
Migrate all metrics of task queues to the new framework.

PR #1484, opened 2023-05-22
feat(new_metrics): feat(new_metrics): migrate metrics for task queue

empiredan:
https://github.com/apache/incubator-pegasus/issues/1483

### DIFF
--- a/src/runtime/task/task_queue.cpp
+++ b/src/runtime/task/task_queue.cpp
@@ -49,14 +49,14 @@ METRIC_DEFINE_gauge_int64(queue,
                           "The length of task queue");
 
 METRIC_DEFINE_counter(queue,
-                          queue_delayed_tasks,
-                          dsn::metric_unit::kTasks,
-                          "The accumulative number of delayed tasks by throttling before enqueue");
+                      queue_delayed_tasks,
+                      dsn::metric_unit::kTasks,
+                      "The accumulative number of delayed tasks by throttling before enqueue");
 
 METRIC_DEFINE_counter(queue,
-                          queue_rejected_tasks,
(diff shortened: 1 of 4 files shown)


src/runtime/task/task_queue.cpp
```diff
@@ -114,4 +128,14 @@ void task_queue::enqueue_internal(task *task)
     tls_dsn.last_worker_queue_size = increase_count();
     enqueue(task);
 }
+
+const metric_entity_ptr &task_queue::queue_metric_entity() const
+{
+    CHECK_NOTNULL(_queue_metric_entity,
```
acelyc111, line 134: How about outputing the name of the queue to distinguish which queue miss initialization?

.github/workflows/lint_and_test_cpp.yaml
```diff
@@ -147,9 +147,13 @@ jobs:
           ./run.sh build --test --skip_thirdparty -j $(nproc) -t release
           ccache -s
       - name: Pack Server
-        run: ./run.sh pack_server
+        run: |
+          ./run.sh pack_server
+          rm -rf pegasus-server-*
```
acelyc111, line 152: How about submitting this update to the master branch?

src/runtime/task/task_queue.cpp
empiredan: OK, I'll fix the checks for all related entities.

.github/workflows/lint_and_test_cpp.yaml
empiredan: OK.

acelyc111 (approved):

PR #1473, opened 2023-05-11
refactor: fs manager

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1383

PR #1487, opened 2023-05-25
refactor: update replica's dir_node status (part1)

acelyc111:

In prior implemention, every replica has a "dir_node status", if a dir_node has
some abnormal status (e.g. in space insufficient), we have to update all replicas'
referenced "dir_node status", it is implemented in `replica_stub::update_disks_status`.
This make the "dir_node status" updating path too long, and a bit of duplicate.

A new implemention is completed in https://github.com/apache/incubator-pegasus/pull/1473,
every replica has a reference of dir_node directly, so it would be easy to update replcia's
"dir_node status" by updating the referenced dir_node's status once.

Before the new implemention, this patch submit a minor refactor to remove
`replica_stub::update_disks_status` and related functions and variables. Also some unit
tests have been updated.


src/common/fs_manager.cpp
```diff
@@ -110,29 +110,26 @@ uint64_t dir_node::remove(const gpid &pid)
...
-        LOG_ERROR("update disk space failed: dir = {}", full_dir);
-        return false;
+    FAIL_POINT_INJECT_F("update_disk_stat", [](string_view) { return; });
+
+    // Get disk space info.
+    dsn::utils::filesystem::disk_space_info dsi;
+    if (!dsn::utils::filesystem::get_disk_space_info(full_dir, dsi)) {
+        // TODO(yingchun): it may encounter some IO errors when get_disk_space_info() failed, deal
+        //  with it.
+        LOG_ERROR("update disk space failed, dir = {}", full_dir);
```
levy5307, line 122: LOG_ERROR("get disk space info failed, dir = {}", full_dir);

src/common/fs_manager.cpp
```diff
@@ -110,29 +110,26 @@ uint64_t dir_node::remove(const gpid &pid)
...
-    if (!update_disk_status) {
-        LOG_INFO("update disk space succeed: dir = {}, capacity_mb = {}, available_mb = {}, "
-                 "available_ratio = {}%",
-                 full_dir,
-                 disk_capacity_mb,
-                 disk_available_mb,
-                 disk_available_ratio);
-        return false;
-    }
+    // Update status.
```
levy5307, line 132: In my opinion, the comment is useless

src/common/fs_manager.cpp
```diff
@@ -110,29 +110,26 @@ uint64_t dir_node::remove(const gpid &pid)
...
+void dir_node::update_disk_stat()
 {
-    FAIL_POINT_INJECT_F("update_disk_stat", [](string_view) { return false; });
-    dsn::utils::filesystem::disk_space_info info;
-    if (!dsn::utils::filesystem::get_disk_space_info(full_dir, info)) {
-        LOG_ERROR("update disk space failed: dir = {}", full_dir);
-        return false;
+    FAIL_POINT_INJECT_F("update_disk_stat", [](string_view) { return; });
+
+    // Get disk space info.
```
levy5307, line 117: remove this comment

src/common/fs_manager.cpp
```diff
@@ -341,17 +337,16 @@ bool fs_manager::for_each_dir_node(const std::function<bool(const dir_node &)> &
     return true;
 }
 
-void fs_manager::update_disk_stat(bool check_status_changed)
+void fs_manager::update_disk_stat()
 {
+    zauto_read_lock l(_lock);
```
levy5307, line 342: Why did you add a read lock here?

acelyc111: It should be a write lock.
Because there is a gap between reset_disk_stat() and the follwoing updating. If a reader get the disks info in the gap, zeros which are incorrect will be returned.

levy5307 (approved):

empiredan (approved):

Issue #1490, opened 2023-05-26
Units showed "Unknown" for some metrics of new framework


empiredan:
While querying the metrics of new framework, some metric units showed "Unknown" as follows:
```json
    {
        "type": "replica",
        "id": "replica_6.1",
        "attributes": {
            "partition_id": "1",
            "table_id": "6"
        },
        "metrics": [
            {
                "type": "counter",
                "name": "corrupt_writes",
                "unit": "requests",
                "desc": "The number of corrupt writes for each replica",
                "value": 0
            },
            {
                "type": "counter",
                "name": "backup_request_bytes",
                "unit": "Unknown",
                "desc": "The number of bytes for backup requests",
                "value": 0
            },
            {
                "type": "counter",
                "name": "check_and_mutate_bytes",
                "unit": "Unknown",
                "desc": "The number of bytes for CHECK_AND_MUTATE requests",
                "value": 0
            },
            {
                "type": "counter",
                "name": "dup_lagging_writes",
                "unit": "requests",
                "desc": "the number of lagging writes (time lag larger than `dup_lagging_write_threshold_ms`)",
                "value": 0
            },
            {
                "type": "counter",
                "name": "check_and_set_bytes",
                "unit": "Unknown",
                "desc": "The number of bytes for CHECK_AND_SET requests",
                "value": 0
            },
            {
                "type": "percentile",
                "name": "dup_time_lag_ms",
                "unit": "milliseconds",
                "desc": "the time lag (in ms) between master and slave in the duplication",
                "p50": 0,
                "p90": 0,
                "p95": 0,
                "p99": 0,
                "p999": 0
            },
            {
                "type": "counter",
                "name": "multi_put_bytes",
                "unit": "Unknown",
                "desc": "The number of bytes for MULTI_PUT requests",
                "value": 0
            },
            ...
        ],
        ...
    }
```

The reason is that `enum_to_string` are forgotten to be added for some metric units.

PR #1491, opened 2023-05-26
refactor(new_metrics): refactor enum definition for metric types and units

empiredan:
https://github.com/apache/incubator-pegasus/issues/1490

For the enum classes, such as metric type and unit, in addition to the declaration
of all enumerators, extra code is needed to implement `enum_to_string` for each
enumerator. However, the "extra code" tended to be forgotten, just as what has
been said in the above issue #1490.

Therefore, we should find a way that could share all the enumerators between
the declaration and `enum_to_string`. Each enumerator would be written only
once, and there is no need to remember to add `enum_to_string` for each
enumerator.

We could implement the "sharing" by function-like macros. They have the names
of all the enumerators, and accept an argument to perform custom actions over
the names to implement the "sharing". Both metric types and units would be
refactored in this way.

While building ASAN Github actions failed due to running out of disk space.
This problem is resolved by dropping all directories of `CMakeFiles`, which is
also tracked another issue: 
        https://github.com/apache/incubator-pegasus/issues/1497

### DIFF
--- a/src/replica/test/replica_disk_test_base.h
+++ b/src/replica/test/replica_disk_test_base.h
@@ -57,7 +57,6 @@ public:
         generate_mock_app_info();
 
         stub->_fs_manager._dir_nodes.clear();
-        // stub->_fs_manager.reset_disk_stat();
         generate_mock_dir_nodes(dir_nodes_count);
         generate_mock_empty_dir_node(empty_dir_nodes_count);
 
--- a/src/replica/test/replica_disk_test_base.h
+++ b/src/replica/test/replica_disk_test_base.h
@@ -57,7 +57,7 @@ public:
         generate_mock_app_info();
 
         stub->_fs_manager._dir_nodes.clear();
-        stub->_fs_manager.reset_disk_stat();
+        // stub->_fs_manager.reset_disk_stat();
         generate_mock_dir_nodes(dir_nodes_count);
         generate_mock_empty_dir_node(empty_dir_nodes_count);
 
--- a/src/utils/enum_helper.h
+++ b/src/utils/enum_helper.h
@@ -66,6 +66,10 @@
 
 #define ENUM_END(type) ENUM_END2(type, type)
 
+#define ENUM_CONST(str) k##str
+#define ENUM_CONST_DEF(str) ENUM_CONST(str),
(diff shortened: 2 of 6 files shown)


src/replica/test/replica_disk_test_base.h
```diff
@@ -57,7 +57,6 @@ class replica_disk_test_base : public replica_test_base
         generate_mock_app_info();
 
         stub->_fs_manager._dir_nodes.clear();
-        stub->_fs_manager.reset_disk_stat();
```
acelyc111, line 60: Is it related to this patch?

src/utils/enum_helper.h
```diff
@@ -66,6 +66,11 @@
 
 #define ENUM_END(type) ENUM_END2(type, type)
 
+#define ENUM_CONST(str) k##str
+#define ENUM_CONST_DEF(str) ENUM_CONST(str),
+#define ENUM_CONST_REG_STR(enum_class, str)                                                        \
+    helper->register_enum(#str, enum_class::ENUM_CONST(str));
```
acelyc111, line 72: Could you please add some tests to check the enum_helper_xxx works well as expect? Use the existing `metric_type` is OK.

acelyc111, line 72: It would be great if adding a sample of how to use the serial of macros, and the related enum helpers, e.g. convert from/to std::string.

src/replica/test/replica_disk_test_base.h
empiredan: Actually this resulted from rebasing `master` branch. In [master](https://github.com/apache/incubator-pegasus/blob/master/src/replica/test/replica_disk_test_base.h#L60) branch, this line was newly added, while it has been removed in `migrate-metrics-dev` branch. After rebasing, `migrate-metrics-dev` has this line which should have been dropped.

src/utils/enum_helper.h
empiredan: OK, I'll add some test cases.

empiredan: OK, I'll add some sample comments to describe how to use these macros.

acelyc111 (approved):

Issue #1493, opened 2023-05-27
Start onebox failed


acelyc111:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
Start onebox after Pegasus has been built.

2. What did you expect to see?
The onebox start failed, the error message is:
```
/home/laiyingchun/data/pegasus_2.4/onebox/meta1/pegasus_server: error while loading shared libraries: libjvm.so: cannot open shared object file: No such file or directory
```

3. What did you see instead?
The cluster should start normally.

4. What version of Pegasus are you using?
Branch v2.4/master

PR #1499, opened 2023-05-30
fix: fix start_onebox failed because of libjvm.so not found (#1494)

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1493

This is a cherry-pick commit to branch v2.4.
Fix rhe bug of "pegasus_server: error while loading shared libraries: libdsn_replica_server.so: cannot open shared object file: No such file or directory". It's because that the libjvm.so is not added to $LD_LIBRARY_PATH.


levy5307 (approved):
LGTM

PR #1512, opened 2023-06-04
feat: skip IO_ERROR dir_node when assign replicas

acelyc111:

A disk (a.k.a node_dir in Pegasus) is possible to become SPACE_INSUFFICIENT or
IO_ERROR from NORMAL, meanwhile, it's possible to recovery from SPACE_INSUFFICIENT
to NORMAL. So we can keep all node_dirs in system, but only reject to assign
replicas on abnormal node_dirs, reject to do write type of operations on abnormal
node_dirs.

This patch also update some unit tests.


src/replica/replica_stub.cpp
```diff
@@ -618,6 +618,10 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
     LOG_INFO("start to load replicas");
     std::map<dir_node *, std::vector<std::string>> dirs_by_dn;
     for (const auto &dn : _fs_manager.get_dir_nodes()) {
+        // Skip IO error dir_node.
+        if (dn->status == disk_status::IO_ERROR) {
```
empiredan, line 622: The reason why do not check `dn->status != disk_status::NORMAL` is that even if the status is `disk_status::SPACE_INSUFFICIENT` we could still just read this disk rather than writing into it ? Once current status is `disk_status::SPACE_INSUFFICIENT` should we reject any write ?

acelyc111: Yes.
It is how we do in replica::on_client_write.

src/replica/replica_2pc.cpp
```diff
@@ -179,8 +180,7 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
     }
 
     if (FLAGS_reject_write_when_disk_insufficient &&
-        (_dir_node->status == disk_status::SPACE_INSUFFICIENT ||
-         _primary_states.secondary_disk_space_insufficient())) {
+        (_dir_node->status != disk_status::NORMAL || _primary_states.secondary_disk_abnormal())) {
         response_client_write(request, ERR_DISK_INSUFFICIENT);
```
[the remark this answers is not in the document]
acelyc111: done

empiredan (approved):

src/replica/replica_2pc.cpp
```diff
@@ -116,6 +117,21 @@ DSN_DEFINE_uint64(
...
 
+namespace {
+error_code disk_status_to_error_code(disk_status::type ds)
+{
+    switch (ds) {
+    case disk_status::SPACE_INSUFFICIENT:
+        return dsn::ERR_DISK_INSUFFICIENT;
+    case disk_status::IO_ERROR:
+        return dsn::ERR_DISK_IO_ERROR;
+    default:
```
levy5307, line 128: It's better to move `disk_status::NORMAL` in front of all other cases to improve cache efficiency

src/replica/replica_stub.cpp
levy5307, line 622: dsn_unlikely

src/replica/replica_2pc.cpp
acelyc111: Now the only caller will not pass a disk_status::NORMAL parameter to this function.

levy5307 (approved):

empiredan (approved):

Issue #1206, opened 2022-10-26
Feature(new_metrics): provide a RESTful service to query immediate values of metrics from new framework


empiredan:
# 1. Motivation

The new framework should provide a service to collect metrics. Since it's hoped that the metrics are collected very conveniently, the service can be **RESTful**. With *http* protocol, the metrics can be viewed just by any browser or a `curl` command.

# 2. Data types

## 2.1 Request

Common RESTful semantics will be adopted for the service. For example, metrics can queried by `GET` method with parameters passed in [query string](https://en.wikipedia.org/wiki/Query_string).

However, there are just field pairs (name and value) in query string. In our scenario, complex data structures such as array should be supported. For example, multiple metric names may be contained in a query.

To solve this problem, the field value can be designed as such format to represent an array: comma("`,`") can be used as the delimiter to separate each element of the array, for example "`put_count,put_latency,alive_node_count`"; also, comma("`,`") can be used safely in query component in URL according to [possible side effect using comma in query string](https://stackoverflow.com/questions/45686595/possible-side-effect-using-comma-in-querystring) and [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.4).

## 2.2 Response

The collected metrics can just be organized in *json* format. The detailed implementations will be described in the following sections.

# 3. API

#### URI

```url
/metrics
```

#### Parameters

| Parameters | Data Types |Description|
|:----------:|:-----------:|:----------|
| types | array | entity types, such as `server`, `table`, `replica`, etc. |
| ids | array | entity IDs. |
| attributes | array | attributes for entity, such as table name and partition id for entity `replica`, and "`attr_key_1,attr_val_1,attr_key_2,attr_val_2`" means there are 2 pairs of attributes: `attr_key_1:attr_val_1` and `attr_key_2:attr_val_2`. |
| metrics | array | metric names. |

#### Headers

```http
Accept: application/json
```

#### Headers

```http
Content-Type: application/json
```

#### Content

```json
[
    { // entity 1
        "type": "<entity_type>",
        "id": "<entity_id>",
        "attributes": {
            "<attr_key_1>": "<attr_val_1>",
            "<attr_key_2>": "<attr_val_2>",
            "<attr_key_3>": "<attr_val_3>",
            ...
        },
        "metrics": [ // metrics that belong to entity 1
            // gauge
            {
                "name": "<gauge_name>",
                "value": 100
            },
            // counter
            {
                "name": "<counter_name>",
[...]

PR #1517, opened 2023-06-07
feat: Aggregate table/server level metrics

xinghuayu007:
Related issue: https://github.com/apache/incubator-pegasus/issues/1206

This patch collects all raw metric data from pegasus, aggregate them into table/server/cluster level and sinks them into prometheus. If the type of the metric is replica partition, it will be aggregated into cluster level table metric.
If the type of the metric is table, it will be aggregated into cluster level table metric, it also will be aggregated into server level table metric.
If the type of the metric is server, it will be aggregated into cluster level table metric, it also will be sinked into prometheus directly.

Currently, this patch supports 3 type of prometheus metric types: counter, gauge, summary.

##### Tests 

- Unit test
- Integration test
- No code

##### Code changes

- Has persistent data change

##### Side effects

- Increased code complexity

##### Related changes

### DIFF
--- a/collector/avail/detector.go
+++ b/collector/avail/detector.go
@@ -19,106 +19,134 @@ package avail
 
 import (
 	"context"
-	"sync/atomic"
+	"math/rand"
 	"time"
 
+	"github.com/apache/incubator-pegasus/go-client/admin"
 	"github.com/apache/incubator-pegasus/go-client/pegasus"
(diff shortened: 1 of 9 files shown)


acelyc111:
We can keep a more rapid development iteration for this develop branch, especially for collector. We will fix these remaining issues in following patches.

Issue #1529, opened 2023-06-12
`throttle_test` failed frequently insufficient disk space


empiredan:
`throttle_test` failed frequently as following error:

```
start test, on set test / throttle by size / 50kb value size
/__w/incubator-pegasus/incubator-pegasus/src/test/function_test/throttle_test/test_throttle.cpp:309: Failure
Value of: last_error == PERR_OK || last_error == PERR_APP_BUSY
  Actual: false
Expected: true
-107
/__w/incubator-pegasus/incubator-pegasus/src/test_util/test_util.cpp:92: Failure
Failed
Timed out waiting for assertion to pass.
/__w/incubator-pegasus/incubator-pegasus/src/test/function_test/throttle_test/test_throttle.cpp:523: Failure
Expected: start_test(plan) doesn't generate new fatal failures in the current thread.
  Actual: it does.
[  FAILED  ] throttle_test.test (625101 ms)
[----------] 1 test from throttle_test (625101 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (625101 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] throttle_test.test

 1 FAILED TEST
dsn exit with code 1
```

`-107` is defined in `src/include/pegasus/error_def.h`, meaning `disk insufficient`:
```c++
PEGASUS_ERR_CODE(PERR_DISK_INSUFFICIENT, -107, "disk insufficient");
```

This resulted from insufficient disk space:
```
2023-06-12 20:21:38.28 (1686572498028082022 60602) replica.default0.0000ec3900010001: fs_manager.cpp:188:update_disk_stat(): update disk space succeed: dir = /data/onebox/replica1/data/replica/reps, capacity_mb = 204700, available_mb = 7750, available_ratio = 4%, disk_status = replication::disk_status::SPACE_INSUFFICIENT
```

PR #1530, opened 2023-06-12
fix(github): throttle_test failed frequently due to insufficient disk space

empiredan:
https://github.com/apache/incubator-pegasus/issues/1529

Options used by onebox and function tests are made configurable. The default
values for these options are unchanged. Once custom options are provided as
the arguments of `run.sh`, these custom options would be used to substitute
for the current configurations. Github workflow for cpp has used custom options
to reduce the consumption of disk space.

On the other hand, to solve the problem that `throttle_test` failed frequently
fundamentally, `throttle_test` should be refactored: at the end of each test
case, onebox for each replica server should be stopped and the data should
be cleared to release the disk space. This is tracked by another issue:
        https://github.com/apache/incubator-pegasus/issues/1534

### DIFF
--- a/.github/workflows/lint_and_test_cpp.yaml
+++ b/.github/workflows/lint_and_test_cpp.yaml
@@ -47,6 +47,7 @@ on:
   workflow_dispatch:
 
 env:
+  # Update the options to reduce the consumption of the disk space
   ONEBOX_OPTS: disk_min_available_space_ratio=5
   TEST_OPTS: throttle_test_medium_value_kb=10,throttle_test_large_value_kb=25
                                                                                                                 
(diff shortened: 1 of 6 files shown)


.github/workflows/lint_and_test_cpp.yaml
```diff
@@ -231,7 +231,7 @@ jobs:
           export LD_LIBRARY_PATH=`pwd`/thirdparty/output/lib:/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/server
           ulimit -s unlimited
           . ./scripts/config_hdfs.sh
-          ./run.sh test -m ${{ matrix.test_module }}
+          ./run.sh test --onebox_opts "disk_min_available_space_ratio=5" -m ${{ matrix.test_module }}
```
acelyc111, line 234: Set it to 5 may still be flaky, how about adjust the threadholds in throttle_test to make it easier to trigger 'throttle' and make the tests passed?

acelyc111 (approved):

levy5307 (approved):

PR #1561, opened 2023-07-06
fix: Add a flag "rocksdb_write_global_seqno" and set it to false by default.

ruojieranyishen:
https://github.com/apache/incubator-pegasus/issues/1562

Add rocksdb_write_global_seqno, and set its default value to false.


ruojieranyishen:
I changed the type of issue. Now reference: https://github.com/apache/incubator-pegasus/issues/1562

Issue #1564, opened 2023-07-10
Bug:manual compact unexpected start


ninsmiracle:
## Bug Report

  In online usage, Pegasus users often use Spark for offline data bulk loading. In the configuration of Spark tasks, users can set the `manual_compact.periodic.trigger_time parameter` to define the desired time for manual compaction, avoiding peak traffic periods. For example, a user generates a batch of data during the day (2 PM) and imports it using Spark, expecting to perform a manual compaction to reduce disk usage at 1 AM the next day.

  However, in the existing code logic now, when a user performs the Bulkload operation and sets the `manual_compact.periodic.trigger_time`, if the specified time (e.g., 01:00) is earlier than the current time (14:00), an immediate manual compaction is triggered. This unexpected behavior leads to two problems:

1. The user expects to perform two bulk loading operations in a short time. The unexpected manual compaction may prevent the second bulk load from starting.

2. In pegasus duplication, when using full-app duplicate, all parameters of the primary cluster's original app are copied to the corresponding app in the backup cluster. If the primary cluster's table has undergone manual compaction using Spark, the parameters related to manual compaction are also copied. **This leads to the backup cluster's RocksDB instances performing manual compaction before they are fully initialized, resulting in undefined behavior and ultimately causing all nodes in the backup cluster to crash.** Although the SRE can delete all manual compaction parameters of the original app before enabling full duplicate, forgetting to do so will result in serious issues.

PR #1565, opened 2023-07-11
fix: manual compact unexpected start

ninsmiracle:
#1564 

  My modifications mainly focus on `pegasus_manual_compact_service::check_periodic_compact`. When the user sets `MANUAL_COMPACT_PERIODIC_TRIGGER_TIME_KEY` for the first time in `app_envs`, the midnight of the date on which the time is set is recorded.

  If the user sets the trigger time to 02:00 and the current time is 15:00, they want to perform manual_compact again at 02:00 the next day. `_manual_compact_first_day_s.load() == cur_day_midnight && t_ms < now` effectively restricts manual_compact on the same day for the user.

  However, if the user sets the trigger time to 16:00 and the current time is still 15:00, it is obvious that the user wants to do manual_compact at 16:00 on the same day. The `_time_natural_increase` member variable resolves this situation by indicating that the time has naturally progressed under the current logic, rather than comparing it with the initial `MANUAL_COMPACT_PERIODIC_TRIGGER_TIME_KEY` setting.

Lastly, if the user deletes the `MANUAL_COMPACT_PERIODIC_TRIGGER_TIME_KEY` parameter and then sets it again, it will be treated as the first-time setting.

##### Tests 

- Unit test
- Cluster test


src/server/pegasus_manual_compact_service.h
```diff
@@ -91,7 +91,9 @@ class pegasus_manual_compact_service : public dsn::replication::replica_base
 
     // manual compact state
     std::atomic<bool> _disabled;
+    std::atomic<bool> _time_natural_increase;
     std::atomic<int> _max_concurrent_running_count;
+    std::atomic<uint64_t> _manual_compact_first_day_s;
```
acelyc111, line 96: Could you please add some comments for the newly added variable?

The logic is too complex if adding these variables, if the current "periodic", "max_concurrent_running" features are not well implemented, I don't mind you to rewrite them or even remove them.

src/server/pegasus_manual_compact_service.cpp
```diff
@@ -204,8 +214,16 @@ bool pegasus_manual_compact_service::check_periodic_compact(
     }
 
     auto now = static_cast<int64_t>(now_timestamp());
+    int64_t cur_day_midnight = dsn::utils::get_unix_sec_today_midnight();
     for (auto t : trigger_time) {
         auto t_ms = t * 1000;
+
+        if (_manual_compact_first_day_s.load() == cur_day_midnight && t_ms < now &&
```
acelyc111, line 221: It's allowed to set multiple periodic compact tasks on a table, for example there are 2 tasks set to be executed at 02:00 and 16:00, now real world time is 15:00, will the second task (16:00 one) to be executed normally?

Issue #1576, opened 2023-08-01
Free more disk space for workflows on Github


empiredan:
We've reduce the usage of disk space for workflows on Github, it's not enough:

- https://github.com/apache/incubator-pegasus/pull/1486
- https://github.com/apache/incubator-pegasus/pull/1498

However, as more and more workflows on Github have failed for the reason "No space left on device", we have to free more disk space to get them to work normally again:
![image](https://github.com/apache/incubator-pegasus/assets/743379/7fb4b3b6-44d3-4faf-9f3d-504ef3a57f96)

We could reference what we've done for branch [migrate-metrics-dev](https://github.com/apache/incubator-pegasus/tree/migrate-metrics-dev): https://github.com/apache/incubator-pegasus/pull/1537

Also, some tests failed due to SPACE_INSUFFICIENT:
![image](https://github.com/apache/incubator-pegasus/assets/743379/10685de8-b8ae-4922-91a4-53257087e232)

PR #1577, opened 2023-08-01
chore(github): free disk space for github workflows

empiredan:
https://github.com/apache/incubator-pegasus/issues/1576

Free more disk space for github workflows in the following ways:

- remove unnecessary packages and directories with reduced disk space up
   to 11GB;
- support updating options such as `disk_min_available_space_ratio` by env
   `TEST_OPTS` for some tests;
- reduce `disk_min_available_space_ratio` from 10 to 5 to some failed tests
   due to `SPACE_INSUFFICIENT`.

### DIFF
--- a/src/common/test/run.sh
+++ b/src/common/test/run.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # The MIT License (MIT)
 #
 # Copyright (c) 2015 Microsoft Corporation
--- a/src/common/test/run.sh
+++ b/src/common/test/run.sh
@@ -31,6 +31,25 @@ exit_if_fail() {
     fi
 }
 
+if [ -n ${TEST_OPTS} ]; then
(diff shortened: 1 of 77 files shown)


src/common/test/run.sh
```diff
@@ -31,6 +31,25 @@ exit_if_fail() {
     fi
 }
 
+if [ -n ${TEST_OPTS} ]; then
```
acelyc111, line 34: I‘m planning to remove the `run.sh` scripts for unit tests in a long term, this change may be a small block for that.
How about define the default value as
```
DSN_DEFINE_int32(replication,
                 disk_min_available_space_ratio,
#ifdef PEGASUS_UNIT_TEST
                 5,
#else
                 10,
#endif
                 "if disk available space ratio "
```

acelyc111 (approved):

levy5307 (approved):

Issue #1589, opened 2023-08-28
Bug:The server crash when balance with duplication


ninsmiracle:
## Bug Report
Encountering the same situation as described in issue #693. 
However, I can reproduce this bug to occur 100%.

Please answer these questions before submitting your issue. Thanks!

This bug appear in online server when the clueter doing duplication and try to do load balance at the same time. This bug make a lot of machine coredump.

1. What did you do?
-Created a large-scale application with a substantial replica size (greater than 10GB).
-Shutdown two replica servers.
-Waited for the cluster to achieve complete health.
-Initiated data streaming (and waited for 30 minutes).
-Enabled duplication function (and waited for 30 minutes).
-Restarted the two replica servers.
-Set the meta level to lively, initiating the balancing process.

2. What did you see instead?
server crush with coredump info:
```
#0  dsn::zlock::lock (this=this@entry=0x98)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/runtime/zlocks.cpp:89
89      /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/runtime/zlocks.cpp: No such file or directory.
(gdb) #0  dsn::zlock::lock (this=this@entry=0x98)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/runtime/zlocks.cpp:89
#1  0x00007f7d063eb9af in zauto_lock (lock=..., this=<synthetic pointer>)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/include/dsn/tool-api/zlocks.h:121
#2  dsn::replication::mutation_log::max_commit_on_disk (this=this@entry=0x0)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/replica/mutation_log.cpp:877
#3  0x00007f7d064bcf1b in dsn::replication::load_mutation::run (1678968919
    this=0x2fe2ce380)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/replica/duplication/duplication_pipeline.cpp:45
#4  0x00007f7d06625631 in dsn::task::exec_internal (
    this=this@entry=0xcc81a0870)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/runtime/task/task.cpp:176
#5  0x00007f7d0663ace2 in dsn::task_worker::loop (this=0x2e2e580)
[...]

PR #1590, opened 2023-08-29
fix:dup conflict with balance

ninsmiracle:
#1589 

There are two conflict may occure when opening duplication and balance:
1.load(one stage of duplication) with replica close
  Add a atomic parameter and check replica still have doing log load when it's closed. 
  So how to end load and let replica close continue?There are two way could make it happen:
-duplication step to next stage(shipping)
-wait duplication do **update_duplication_map**,and it will `remove_all_duplications` when replica loss primary identity.

In remove_all_duplications,duplication will use a map named `_replica` to check the replica identity,so I protected it when replica close. 

2.gc useless replica with replica close (close one replica twice)
  After replica connected with meta, meta will request for config to replica every 500ms. And replica server update itself's config when meta reply.In this logic ,on_node_query_reply_scatter2 will gc useless replica(which status is PS_INACTIVE),to be precise will set status from PS_INACTIVE TO PS_ERROR.
  When replica doing `update_local_configuration`,it will exec close action when status changed and new status is PS_INACTIVE or PS_ERROR.
  So I judge replica is already closed or doing close when replica exec close replica to deal with above problem.

##### Tests 

- Cluster test(mentioned in issue#1589)

### In summary:
  The earliest zlock coredump was caused by the necessity of acquiring a lock during the 'dup' load stage, which requires a lock from a member variable of the 'replica' class. However, at this point, the 'replica' had already been closed, related members are destructedresulting in obtaining,resulting zlock get an error `_lock` value.

  The root cause of the replica being closed actually stems from the logic where the meta requests 'replica' for configuration, triggering a 'replica gc' process that sets the replica itself to 'PS_ERROR'. Based on the existing logic, both 'PS_INACTIVE' and 'PS_ERROR' trigger the 'begin_close' logic, with 'inactive' having a 10-minute delay and 'error' being executed immediately.

  A double safeguard was implemented for the modification. Firstly, when enqueuing the close operation, it is checked whether the replica has already been closed or is in the process of closing. Secondly, during the execution of the close operation, in case it encounters a 'dup load', it is configured to enter the task queue with a delay of one minute.

### DIFF
--- a/src/replica/duplication/duplication_pipeline.cpp
+++ b/src/replica/duplication/duplication_pipeline.cpp
@@ -53,6 +53,7 @@ void load_mutation::run()
 {
     decree last_decree = _duplicator->progress().last_decree;
     _start_decree = last_decree + 1;
+    _duplicator->set_is_loading(true);
     if (_replica->private_log()->max_commit_on_disk() < _start_decree) {
         // wait 100ms for next try if no mutation was added.
         repeat(100_ms);
--- a/src/replica/duplication/load_from_private_log.cpp
+++ b/src/replica/duplication/load_from_private_log.cpp
@@ -246,7 +246,7 @@ void load_from_private_log::replay_log_block()
     // mutations()
     step_down_next_stage(_mutation_batch.last_decree(), _mutation_batch.move_all_mutations());
 
-    //step to ship meaning load has been finished
+    // step to ship meaning load has been finished
     _duplicator->set_is_loading(false);
 }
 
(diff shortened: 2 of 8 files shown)


src/replica/duplication/duplication_sync_timer.h
```diff
@@ -73,6 +73,11 @@ class duplication_sync_timer
 
     std::vector<replica_ptr> get_all_replicas();
 
+    // duplication_sync_timer is async with replica close,so replica_duplicator_manager may already
+    // release
+    // dup should stop right now
+    bool replica_is_cloing_or_closed(gpid id);
```
acelyc111, line 79: The same to https://github.com/apache/incubator-pegasus/pull/1608, you can judge the state of the replica by itself.

Issue #1593, opened 2023-08-31
Feature: apply and remove all shared logs


empiredan:
In https://github.com/XiaoMi/rdsn/pull/1019 we've written private logs as WAL instead of shared logs.

However, obsolete shared logs that had been applied to rocksdb were not removed since then. Since we've kept at least 1 shared log file which would never be removed, we should change policy of garbage collection to delete all shared logs.

PR #1594, opened 2023-08-31
feat(slog): flush and remove all shared logs for garbage collection

empiredan:
https://github.com/apache/incubator-pegasus/issues/1593

In https://github.com/XiaoMi/rdsn/pull/1019 we've written private logs as
WAL instead of shared logs, which also means shared log files would never
be appended with new mutations.

However, obsolete shared logs that had been applied to rocksdb were not
removed since then. There is at least 1 shared log file which is never removed.
We should change this policy of garbage collection, to delete all obsolete
shared log files.

To facilitate the garbage collection of shared log files, we also trigger checkpoints
which would flush rocksdb data to disk for each replica that has prevented shared
log files from being removed for garbage collection. It is necessary to limit the
number of submitted replicas that would be flushed at a time to avoid too much
consumption of I/O resources which might affect the processing of read/write
requests from clients.

### DIFF
--- a/src/replica/mutation_log.h
+++ b/src/replica/mutation_log.h
@@ -285,8 +285,10 @@ public:
     decree max_gced_decree(gpid gpid) const;
     decree max_gced_decree_no_lock(gpid gpid) const;
 
+    using log_file_map_by_index = std::map<int, log_file_ptr>;
+
     // thread-safe
-    std::map<int, log_file_ptr> get_log_file_map() const;
+    log_file_map_by_index get_log_file_map() const;
 
     // check the consistence of valid_start_offset
     // thread safe
@@ -299,6 +301,58 @@ public:
     void demand_switch_file() { _switch_file_demand = true; }
 
     task_tracker *tracker() { return &_tracker; }
+
+    struct reserved_slog_info
+    {
+        size_t file_count = 0;
+        int64_t log_size = 0;
+        int min_file_index = 0;
+        int max_file_index = 0;
+
+        std::string to_string() const
+        {
+            return fmt::format("reserved_slog_info = [file_count = {}, log_size = {}, "
+                               "min_file_index = {}, max_file_index = {}]",
+                               file_count,
+                               log_size,
+                               min_file_index,
+                               max_file_index);
+        }
+
+        friend std::ostream &operator<<(std::ostream &os, const reserved_slog_info &reserved_log)
+        {
+            return os << reserved_log.to_string();
+        }
+    };
+
+    struct slog_deletion_info
+    {
+        int to_delete_file_count = 0;
+        int64_t to_delete_log_size = 0;
+        int deleted_file_count = 0;
+        int64_t deleted_log_size = 0;
+        int deleted_min_file_index = 0;
+        int deleted_max_file_index = 0;
+
+        std::string to_string() const
+        {
+            return fmt::format("slog_deletion_info = [to_delete_file_count = {}, "
+                               "to_delete_log_size = {}, deleted_file_count = {}, "
+                               "deleted_log_size = {}, deleted_min_file_index = {}, "
+                               "deleted_max_file_index = {}]",
+                               to_delete_file_count,
+                               to_delete_log_size,
+                               deleted_file_count,
+                               deleted_log_size,
+                               deleted_min_file_index,
+                               deleted_max_file_index);
+        }
+
+        friend std::ostream &operator<<(std::ostream &os, const slog_deletion_info &log_deletion)
+        {
+            return os << log_deletion.to_string();
+        }
+    };
 
 protected:
     // thread-safe
@@ -325,7 +379,7 @@ private:
                              replay_callback callback,
                              /*out*/ int64_t &end_offset);
 
-    static error_code replay(std::map<int, log_file_ptr> &log_files,
+    static error_code replay(log_file_map_by_index &log_files,
                              replay_callback callback,
                              /*out*/ int64_t &end_offset);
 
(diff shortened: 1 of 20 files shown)


src/replica/mutation_log.h
```diff
@@ -345,6 +345,67 @@ class mutation_log : public ref_counter
     // get total size ithout lock.
     int64_t total_size_no_lock() const;
 
+    struct reserved_slog_info
+    {
+        size_t file_count;
+        int64_t log_size;
+        int smallest_file_index;
+        int largest_file_index;
```
acelyc111, line 353: How about renaming "min" and "max"?

src/replica/mutation_log.h
```diff
@@ -345,6 +345,67 @@ class mutation_log : public ref_counter
...
+                               log_size,
+                               smallest_file_index,
+                               largest_file_index);
+        }
+
+        friend std::ostream &operator<<(std::ostream &os, const reserved_slog_info &reserved_log)
+        {
+            return os << reserved_log.to_string();
+        }
+    };
```
acelyc111, line 369: It's needed to add USER_DEFINED_STRUCTURE_FORMATTER for reserved_slog_info because we have upgraded libfmt.

src/replica/mutation_log.h
```diff
@@ -345,6 +345,67 @@ class mutation_log : public ref_counter
...
+                               deleted_log_size,
+                               deleted_smallest_file_index,
+                               deleted_largest_file_index);
+        }
+
+        friend std::ostream &operator<<(std::ostream &os, const slog_deletion_info &log_deletion)
+        {
+            return os << log_deletion.to_string();
+        }
+    };
```
acelyc111, line 398: Same.

src/replica/mutation_log.h
acelyc111, line 353: Initialize the variables to 0 ?

src/replica/mutation_log.h
```diff
@@ -345,6 +345,67 @@ class mutation_log : public ref_counter
...
+
+    struct slog_deletion_info
+    {
+        int to_delete_file_count = 0;
+        int64_t to_delete_log_size = 0;
+        int deleted_file_count = 0;
+        int64_t deleted_log_size = 0;
+        int deleted_smallest_file_index = 0;
+        int deleted_largest_file_index = 0;
```
acelyc111, line 378: Same, rename to "min" and "max"?

src/replica/mutation_log.h
```diff
@@ -345,6 +345,67 @@ class mutation_log : public ref_counter
...
+                               deleted_largest_file_index);
+        }
+
+        friend std::ostream &operator<<(std::ostream &os, const slog_deletion_info &log_deletion)
+        {
+            return os << log_deletion.to_string();
+        }
+    };
+
+    using log_file_map = std::map<int, log_file_ptr>;
```
acelyc111, line 400: How about renaming to `log_file_by_index` to indicate the key and value?

src/replica/replica_stub.h
```diff
@@ -361,6 +364,30 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
+    using replica_gc_info_map = std::unordered_map<gpid, replica_gc_info>;
+
+    // Try to remove obsolete files of shared log for garbage collection according to the provided
+    // states of all replicas. The purpose is to remove all of the files of shared log, since it
+    // has been deprecated, and would not be appended any more.
+    void gc_slog(const replica_gc_info_map &replica_gc_map);
+
+    // The number of flushed replicas for the garbage collection of shared log at a time should be
+    // limited.
+    void limit_flush_replicas_for_slog_gc(size_t prevent_gc_replica_count);
```
acelyc111, line 384: Why add this limitation, what will happen if not limit it?

src/replica/replica_stub.h
```diff
@@ -361,6 +364,30 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
+
+    // Flush rocksdb data to sst files for replicas to facilitate garbage collection of more files
+    // of shared log.
+    void flush_replicas_for_slog_gc(const replica_gc_info_map &replica_gc_map,
```
acelyc111, line 388: What does the second parameter `prevent_gc_replicas` means?

src/replica/replica_stub.h
```diff
@@ -436,6 +464,15 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
     closing_replicas _closing_replicas;
     closed_replicas _closed_replicas;
 
+    // The number of replicas that prevent slog files from being removed for gc at last round.
```
acelyc111, line 467:
```suggestion
    // The number of replicas that prevent slog files from being removed for gc at the last round.
```

src/replica/replica_stub.h
empiredan: The 2nd parameter is just the replicas which prevent shared files from being removed for garbage collection. Thus their rocksdb data would be flushed to disk to facilitate the garbage collection for the shared logs.

src/replica/replica_stub.h
empiredan: Both flushing replicas to the disk and processing read/write requests would consume much more IO resources. Too much flushes would affect read/write requests. Thus it should be limited.

src/replica/replica_stub.cpp
```diff
@@ -1767,138 +1777,211 @@ void replica_stub::on_gc_replica(replica_stub_ptr this_, gpid id)
...
+
+    std::ostringstream oss;
+    size_t i = 0;
+    for (const auto &pid : prevent_gc_replicas) {
+        if (i != 0) {
+            oss << ", ";
+        }
+        oss << pid.to_string();
+        ++i;
+    }
```
acelyc111, line 1921: How about using ` fmt::join()` instead?

acelyc111 (approved):

levy5307 (approved):

Issue #1596, opened 2023-09-01
Bug:(duplication) master cluster down when duplication checking a private log to load


ninsmiracle:
## Bug Report

Please answer these questions before submitting your issue. Thanks!
Master cluster core dump after duplication task running some times.

1. What did you do?
This coredump occured in an online cluster. We begin running a duplication task and after a few days, and some nodes begin coredump.

2. What did you see instead?
We can see that some error happened when duplication checking a plog to load(a staget of duplication).
```
Program terminated with signal 6, Aborted.
#0  0x00007fbfd14e21d7 in raise () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install bzip2-libs-1.0.6-13.el7.x86_64 elfutils-libelf-0.166-2.el7.x86_64 elfutils-libs-0.166-2.el7.x86_64 glibc-2.17-157.el7_3.1.x86_64 keyutils-libs-1.5.8-3.el7.x86_64 krb5-libs-1.14.1-27.el7_3.x86_64 libattr-2.4.46-12.el7.x86_64 libcap-2.22-8.el7.x86_64 libcom_err-1.42.9-9.el7.x86_64 libgcc-4.8.5-28.el7_5.1.x86_64 libselinux-2.5-6.el7.x86_64 pcre-8.32-15.el7_2.1.x86_64 systemd-libs-219-30.el7_3.8.x86_64 xz-libs-5.2.2-1.el7.x86_64 zlib-1.2.7-17.el7.x86_64
(gdb) #0  0x00007fbfd14e21d7 in raise () from /lib64/libc.so.6
#1  0x00007fbfd14e38c8 in abort () from /lib64/libc.so.6
#2  0x00007fbfd65c963e in dsn_coredump ()
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/runtime/service_api_c.cpp:93
#3  0x00007fbfd63c47c3 in dsn::replication::log_file::log_file (
    this=0x52b031600,
    path=0x672f88718 "/home/work/ssd2/pegasus/alsgsrv-monetization-master/replica/reps/8.79.pegasus/plog/log.2889.96902391528", handle=<optimized out>,
    index=<optimized out>, start_offset=96902391528, is_read=<optimized out>)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/replica/log_file.cpp:166
#4  0x00007fbfd63c630e in dsn::replication::log_file::open_read (
[...]

PR #1597, opened 2023-09-01
fix(duplication): prevent plog files from being removed by GC while they are being checked by duplication

ninsmiracle:
https://github.com/apache/incubator-pegasus/issues/1596

Using an atomic member to prevent plog files from being removed by GC
when the plog files are being checked by duplication.

##### Tests 
- Unit test
-cluster test


src/replica/replica.h
```diff
@@ -633,6 +639,7 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
     bool _is_manual_emergency_checkpointing{false};
     bool _is_duplication_master{false};
     bool _is_duplication_follower{false};
+    std::atomic<bool> _is_duplication_plog_checking{false};
```
acelyc111, line 642: Where will the variable be used?

acelyc111:
@ninsmiracle Thanks for the contribution!

This pull request mention "slog" but I didn't see how does slog been operated in the patch, could you please describe the relationship? Thanks!

ninsmiracle:
> @ninsmiracle Thanks for the contribution!
> 
> This pull request mention "slog" but I didn't see how does slog been operated in the patch, could you please describe the relationship? Thanks!

I think this PR should called "plog gc crash",however in our inside gitlab repository,this commit called this name.I guess it's may be related to old version Slog ? But I'm not familiar with old version pegasus.

acelyc111:
> > @ninsmiracle Thanks for the contribution!
> > This pull request mention "slog" but I didn't see how does slog been operated in the patch, could you please describe the relationship? Thanks!
> 
> In my opinion, this PR should called "plog gc crash",however in our inside gitlab repository,this commit called this name.I guess it's may be related to old version Slog ? But I'm not familiar with old version pegasus. So could I change the PR's name?

Feel free to rename.

src/replica/replica.h
```diff
@@ -633,6 +638,7 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
...
```
acelyc111, line 641: Could you please to add some comments about this variable?

acelyc111 (dismissed):

src/replica/replica.h
```diff
@@ -625,6 +630,8 @@ class replica : public serverlet<replica>, public ref_counter, public replica_ba
     bool _is_manual_emergency_checkpointing{false};
     bool _is_duplication_master{false};
     bool _is_duplication_follower{false};
+    // replica is finding some private logs to load for duplication,avoid unexpected plog gc
```
acelyc111, line 633:
```suggestion
    // Indicate whether the replica is during finding out some private logs to
    // load for duplication. It useful to prevent plog GCed unexpectedly.
```

acelyc111 (approved):

src/replica/duplication/load_from_private_log.cpp
```diff
@@ -146,6 +146,8 @@ void load_from_private_log::run()
 
 void load_from_private_log::find_log_file_to_start()
 {
+    _duplicator->set_duplication_plog_checking(true);
```
empiredan, line 149: Would it be fine to return from the middle of the function without roll back by `_duplicator->set_duplication_plog_checking(false);` ?

src/replica/duplication/load_from_private_log.cpp
```diff
@@ -157,6 +159,7 @@ void load_from_private_log::find_log_file_to_start()
         error_s es = log_utils::open_read(pr.second->path(), file);
         if (!es.is_ok()) {
             LOG_ERROR_PREFIX("{}", es);
+            _duplicator->set_duplication_plog_checking(false);
```
empiredan, line 162: There's `return` statement in function `find_log_file_to_start(log_file_map)`, which should be also set false:

```c++
void load_from_private_log::find_log_file_to_start(
    const mutation_log::log_file_map_by_index &log_file_map)
{
    _current = nullptr;
    if (dsn_unlikely(log_file_map.empty())) {
        LOG_ERROR_PREFIX("unable to start duplication since no log file is available");
        return;
    }
```
```

File: src/replica/duplication/load_from_private_log.cpp (hunk 4 above)
Comment (empiredan) [inline src/replica/duplication/load_from_private_log.cpp:149]:
```suggestion
    _duplicator->set_duplication_plog_checking(true);
    auto cleanup = dsn::defer([this]() { _duplicator->set_duplication_plog_checking(false); });
```

src/replica/duplication/load_from_private_log.cpp
```diff
@@ -168,6 +171,8 @@ void load_from_private_log::find_log_file_to_start()
 void load_from_private_log::find_log_file_to_start(
     const mutation_log::log_file_map_by_index &log_file_map)
 {
+    auto cleanup = dsn::defer([this]() { _duplicator->set_duplication_plog_checking(false); });
```
empiredan, line 174: Is this line necessary ? `_duplicator->set_duplication_plog_checking(false);` would be called twice ?

empiredan (dismissed):

acelyc111 (dismissed):

empiredan (approved):

acelyc111 (approved):

Issue #1595, opened 2023-08-31
Bug:(duplication) master cluster verify start decree error when restart a pause dup


ninsmiracle:
## Bug Report

When I restart a paused dup app on a recovered servers , the start decree of the duplication looking for will been loss,and those servers coredump agian.
It also has been occured online.

1. What did you do?
- Set meta level to lively firstly.
- Begin a duplication task firstly and wait for 5 minutes.
- Shut down two nodes of the master cluster, and then pause the duplication task immediately.
- Wait for one minutes,and restart the previously shut down nodes.
- Wait for ten minutes.(Give replica enough time to GC plog)
- Resume the duplication task 10 minutes later.
- Nodes which stopped just now core dump.

3. What did you see instead?
There are many coredump with the message "replica_duplicator::verify_start_decree" in master cluster.
```
#0  0x00007f968e6a01d7 in raise () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install glibc-2.17-157.el7_3.1.x86_64 keyutils-libs-1.5.8-3.el7.x86_64 krb5-libs-1.14.1-27.el7_3.x86_64 libcom_err-1.42.9-9.el7.x86_64 libgcc-4.8.5-28.el7_5.1.x86_64 libselinux-2.5-6.el7.x86_64 pcre-8.32-15.el7_2.1.x86_64 zlib-1.2.7-17.el7.x86_64
(gdb) #0  0x00007f968e6a01d7 in raise () from /lib64/libc.so.6
#1  0x00007f968e6a18c8 in abort () from /lib64/libc.so.6
#2  0x00007f969378763e in dsn_coredump ()
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/runtime/service_api_c.cpp:93
#3  0x00007f969364dc54 in dsn::replication::replica_duplicator::verify_start_decree (this=<optimized out>, start_decree=64525717)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/replica/duplication/replica_duplicator.cpp:218   
#4  0x00007f969365ea05 in dsn::replication::load_from_private_log::run (
    this=0xa3016f040)
    at /home/jiashuo1/work/incubator-pegasus/src/rdsn/src/replica/duplication/load_from_private_log.cpp:71
#5  0x00007f96937c4631 in dsn::task::exec_internal (
    this=this@entry=0x90e7ce0f0)
[...]

PR #1608, opened 2023-09-14
fix(duplication): failed to start duplication after source replica servers exit due to missing duplicating status in .app-info

ninsmiracle:
https://github.com/apache/incubator-pegasus/issues/1595

Currently, after a duplication was added, `duplicating` status would be only updated
for in-memory state. However, once some replica servers exited and were started again,
`duplicating` status would be lost since it had not been persisted into `.app-info` before.
Consequently, plog files would be removed by GC, which lead to failed assertion for decree
in `replica_duplicator::verify_start_decree()`.

To resolve this problem, `duplicating` should be updated for both in-memory and on-disk
states. Then, `duplicating` status would be loaded correctly from `.app-info` file at the loading
stage of replicas.

##### Tests 
- Manual test
- Cluster test


src/replica/duplication/duplication_sync_timer.cpp
```diff
@@ -107,11 +107,23 @@ void duplication_sync_timer::update_duplication_map(
...
         } else {
             r->get_duplication_manager()->update_duplication_map(it->second);
+            doing_duplication = true;
+        }
+
+        if (_stub->_closed_replicas.find(id) == _stub->_closed_replicas.end() &&
+            _stub->_closing_replicas.find(id) == _stub->_closing_replicas.end()) {
+            doing_duplication ? info->__set_duplicating(true) : info->__set_duplicating(false);
+            r->store_app_info(*info, "");
         }
```
acelyc111, line 127: How about updating the app info in `update_duplication_map()`?

src/replica/duplication/duplication_sync_timer.cpp
```diff
@@ -211,5 +232,14 @@ duplication_sync_timer::get_dup_states(int app_id, /*out*/ bool *app_found)
     return result;
 }
 
+bool duplication_sync_timer::replica_is_cloing_or_closed(gpid id)
+{
+    if (_stub->_closing_replicas.find(id) != _stub->_closing_replicas.end() ||
```
acelyc111, line 237: Move this function to replica_stub is more meaningful, and it's needed to protected by `_replicas_lock` to avoid data race.

src/replica/duplication/duplication_sync_timer.h
```diff
@@ -73,6 +73,10 @@ class duplication_sync_timer
 
     std::vector<replica_ptr> get_all_replicas();
 
+    // duplication_sync_timer is async with replica close,so replica_duplicator_manager may already
+    // release dup should stop right now
```
acelyc111, line 77: Just describe what does this function do is OK, you can describe why invoke it when in the caller.

src/replica/duplication/duplication_sync_timer.cpp
acelyc111: I think it's possible to judge the replica's state by itself, maybe by `status()`, it's not needed to access replica_stub.

src/replica/duplication/duplication_sync_timer.cpp
```diff
@@ -113,12 +117,28 @@ void duplication_sync_timer::update_duplication_map(
...
             // no duplication is assigned to this app
             r->get_duplication_manager()->update_duplication_map({});
+            doing_duplication = false;
         } else {
             r->get_duplication_manager()->update_duplication_map(it->second);
+            doing_duplication = true;
+        }
+
+        if (doing_duplication) {
+            info->__set_duplicating(true);
```
acelyc111, line 137: It would be better to update the info in replica class to avoid concurrent write, see replica::update_app_max_replica_count() for example.

src/replica/duplication/duplication_sync_timer.cpp
acelyc111: > I think it's possible to judge the replica's state by itself, maybe by `status()`, it's not needed to access replica_stub.

@ninsmiracle take a look at this.

acelyc111 (approved):

acelyc111 (approved):

empiredan (approved):

Issue #1575, opened 2023-08-01
Data at rest encryption


acelyc111:
# Motivation

There are some Pegasus users that store privacy data in Pegasus, it’s important to protect the data against unauthorized access by persons who gain access to the storage media used by Pegasus.

It's possible to support transparent data at rest encryption to provide a way to protect users’ data, which is transparent to users and straightforward to set up for operators.

Data at rest encryption refers to encrypting data for storage and decrypting it when reading the stored data. It uses symmetric encryption where the same key is used to encrypt and to decrypt the data. Keys need to be stored and handled securely as anyone with access to a key will be able to decrypt any data encrypted with it.

# Cloud disk encryption

If your Pegasus clusters are deployed on public cloud service storages, it’s possible to use their own encryption solutions. See:

- [Amazon EBS Encryption](https://docs.aws.amazon.com/zh_cn/AWSEC2/latest/UserGuide/EBSEncryption.html)
- [Aliyun ECS Encryption](https://help.aliyun.com/document_detail/59643.html?spm=a2c4g.59643.0.0)
- [Tencent cloud CBS Encryption](https://cloud.tencent.com/document/product/362/38946)
- [Huawei cloud EVS Encryption](https://support.huaweicloud.com/productdesc-evs/evs_01_0001.html)

It’s not needed to enable Pegasus Data at rest encryption to avoid encrypting/decrypting data twice, which may lead to poor performance.

# Goals

- Data at rest encryption of all user data (key-values) on a fresh Pegasus cluster.
- Pluggable key management to enable interfacing with existing key management systems, such as [Hadoop KMS](https://hadoop.apache.org/docs/stable/hadoop-kms/index.html#KMS_HTTP_REST_API).
- Cluster key architecture (see Key management).
- User data in logs will be redacted.

# Non-Goals

- Enabling (or disabling) data at rest encryption on an existing cluster
  TODO: It's possible to implement this, after all the data been full compacted, the data could transfer to plaintext/ciphertext.
- Multiple tenants, or table granularity encryption.
  TODO: It's possible to implement this after the cluster granularity encryption been implemented.
- Selective encryption (certain tables are encrypted, others are not)
  TODO: Same to the above.
- Encrypt data of shell-tools output.
- Transport layer encryption.
  TODO: Use TLS libs.
- Core dump encryption.
- [pegasus-spark](https://github.com/pegasus-kv/pegasus-spark)
[...]

PR #1637, opened 2023-10-12
feature(aio): use rocksdb APIs to re-implement the aio module

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1575

This is a dependent work to implement encryption at rest, we can use the capacity of
rocksdb encryption after this patch.

- Use rocksdb APIs to implement class `native_linux_aio_provider`. Both of the
  implementations are using `pread()` and `pwrite()` system calls, so there isn't
  significant performance changes, see the newly added simple benchmark performance
  comparation below.
- Separate the file read and write operations for class `aio_provider`

### DIFF
--- a/src/aio/file_io.cpp
+++ b/src/aio/file_io.cpp
@@ -26,32 +26,51 @@
 
 #include "aio/file_io.h"
 
+#include <memory>
 // IWYU pragma: no_include <algorithm>
 #include <vector>
 
 #include "aio/aio_provider.h"
 #include "disk_engine.h"
+#include "rocksdb/env.h"
+#include "utils/fmt_logging.h"
 
 namespace dsn {
 class task_tracker;
 
 namespace file {
 
-/*extern*/ disk_file *open(const char *file_name, int flag, int pmode)
+/*extern*/ disk_file *open(const std::string &fname, FileOpenType type)
 {
-    auto fd = disk_engine::provider().open(file_name, flag, pmode);
-    if (fd.is_invalid()) {
-        return nullptr;
+    switch (type) {
+    case FileOpenType::kReadOnly: {
+        auto sf = disk_engine::provider().open_read_file(fname);
+        if (!sf) {
+            return nullptr;
+        }
+        return new disk_file(std::move(sf));
     }
-
-    return new disk_file(fd);
+    case FileOpenType::kWriteOnly: {
+        auto wf = disk_engine::provider().open_write_file(fname);
+        if (!wf) {
+            return nullptr;
+        }
+        return new disk_file(std::move(wf));
+    }
+    default:
+        CHECK(false, "");
+    }
+    return nullptr;
 }
 
 /*extern*/ error_code close(disk_file *file)
 {
-    error_code result = ERR_INVALID_HANDLE;
+    error_code result = ERR_OK;
     if (file != nullptr) {
-        result = disk_engine::provider().close(file->native_handle());
+        // A read file is not needed to close.
+        if (file->wfile()) {
+            result = disk_engine::provider().close(file->wfile());
+        }
         delete file;
         file = nullptr;
     }
@@ -60,11 +79,11 @@ namespace file {
 
 /*extern*/ error_code flush(disk_file *file)
 {
-    if (nullptr != file) {
-        return disk_engine::provider().flush(file->native_handle());
-    } else {
+    if (file == nullptr || file->wfile() == nullptr) {
         return ERR_INVALID_HANDLE;
     }
+
+    return disk_engine::provider().flush(file->wfile());
 }
 
 /*extern*/ aio_task_ptr read(disk_file *file,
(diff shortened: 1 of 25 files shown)


acelyc111:
Old implementation:
```
[ RUN      ] aio_test.basic/0
sequential write, cost 0.358 ms
sequential read, cost 6.853 ms
concurrent read, cost 0.609 ms
un-sequential write, cost 0.389 ms
sequential read, cost 6.722 ms
concurrent read, cost 0.606 ms
overwrite, cost 0.244 ms
sequential read, cost 6.899 ms
concurrent read, cost 0.615 ms
vector write, cost 0.064 ms
sequential read, cost 6.866 ms
concurrent read, cost 0.617 ms
[       OK ] aio_test.basic/0 (31289 ms)
```

New implementation:
```
[ RUN      ] aio_test.basic/0
sequential write, cost 0.333 ms
sequential read, cost 6.687 ms
concurrent read, cost 0.662 ms
un-sequential write, cost 0.385 ms
sequential read, cost 6.976 ms
concurrent read, cost 0.654 ms
overwrite, cost 0.261 ms
sequential read, cost 6.966 ms
concurrent read, cost 0.612 ms
vector write, cost 0.074 ms
sequential read, cost 6.972 ms
concurrent read, cost 0.641 ms
[       OK ] aio_test.basic/0 (31670 ms)
```

foreverneverer (approved):

src/aio/native_linux_aio_provider.cpp
```diff
@@ -26,107 +26,119 @@
...
 
-linux_fd_t native_linux_aio_provider::open(const char *file_name, int flag, int pmode)
+std::unique_ptr<rocksdb::RandomAccessFile>
+native_linux_aio_provider::open_read_file(const std::string &fname)
 {
-    auto fd = ::open(file_name, flag, pmode);
-    if (fd == DSN_INVALID_FILE_HANDLE) {
-        LOG_ERROR("create file '{}' failed, err = {}", file_name, utils::safe_strerror(errno));
+    std::unique_ptr<rocksdb::RandomAccessFile> rfile;
+    auto s = dsn::utils::PegasusEnv(dsn::utils::FileDataType::kSensitive)
```
empiredan, line 51: means all files would be read or written as encrypted files ?

src/aio/native_linux_aio_provider.cpp
```diff
@@ -26,107 +26,119 @@
...
-                "and will retry it.",
-                aio_ctx.buffer_size,
-                buffer_offset,
-                ret);
-        }
-    } while (dsn_unlikely(buffer_offset < aio_ctx.buffer_size));
+    rocksdb::Slice data((const char *)(aio_ctx.buffer), aio_ctx.buffer_size);
+    auto s = aio_ctx.dfile->wfile()->Write(aio_ctx.file_offset, data);
+    if (!s.ok()) {
+        LOG_ERROR("write file failed, err = {}", s.ToString());
```
empiredan, line 119: Is it necessary to specify which file fails to be written ?

src/aio/native_linux_aio_provider.cpp
```diff
@@ -26,107 +26,119 @@
...
-                       aio_ctx.buffer,
-                       aio_ctx.buffer_size,
-                       aio_ctx.file_offset);
-    if (dsn_unlikely(ret < 0)) {
-        LOG_WARNING("write failed with errno={} and will retry it.", utils::safe_strerror(errno));
+    rocksdb::Slice result;
+    auto s = aio_ctx.dfile->rfile()->Read(
+        aio_ctx.file_offset, aio_ctx.buffer_size, &result, (char *)(aio_ctx.buffer));
+    if (!s.ok()) {
+        LOG_ERROR("read file failed, err = {}", s.ToString());
```
empiredan, line 134: Ditto.

src/aio/file_io.cpp
```diff
@@ -26,32 +26,51 @@
...
+    return nullptr;
 }
 
 /*extern*/ error_code close(disk_file *file)
 {
-    error_code result = ERR_INVALID_HANDLE;
+    error_code result = ERR_OK;
     if (file != nullptr) {
-        result = disk_engine::provider().close(file->native_handle());
+        // A read file is not needed to close.
```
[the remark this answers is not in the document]
acelyc111: Yes, there isn't even a `close()` member function for `rocksdb::RandomAccessFile` class. The file will be closed automatically it its derive classes' destructor.

src/aio/native_linux_aio_provider.cpp
acelyc111: Yes. We can improve to distinguish files which are not necessary to encrypt, however, it's not a problem if all files are encrypted.

src/aio/native_linux_aio_provider.cpp
acelyc111: Not necessary, the callers have the responsibility to log the fikle name.

empiredan (approved):

PR #1641, opened 2023-10-16
chore(CI): unify to use Ubuntu 22.04 for CI

empiredan:
https://github.com/apache/incubator-pegasus/issues/1604

Prior to this patch, the `thirdparties-bin-test-${OS}-${branch}` docker images are
only built for Ubuntu 18.04 for CI purpose, but some CI jobs use Ubuntu 22.04 image,
it may cause the CI failed because the Ubuntu 22.04 image is outdated (for example,
lack of or outdated of some thirdparties).
This patch unify to build and use `thirdparties-bin-test-2204-${branch}` docker image.

### DIFF
--- a/.github/workflows/lint_and_test_cpp.yaml
+++ b/.github/workflows/lint_and_test_cpp.yaml
@@ -259,7 +259,7 @@ jobs:
     needs: cpp_clang_format_linter
     runs-on: ubuntu-latest
     container:
-      image: apache/pegasus:thirdparties-bin-test-ubuntu1804-${{ github.base_ref }}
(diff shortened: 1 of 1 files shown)


acelyc111 (approved):

acelyc111:
Merged although some CIs failed, they are not caused by this patch, we'll resolve the issues later.

empiredan:
> Merged although some CIs failed, they are not caused by this patch, we'll resolve the issues later.

Yes, we introduce this patch since `thirdparties-bin-test-ubuntu1804-v2.5` would not be generated:
image

Issue #1647, opened 2023-10-18
Compilation error "no member named 'make_unique' in namespace 'std'" on MacOS for v2.5


empiredan:
According to https://github.com/apache/incubator-pegasus/pull/1640#issuecomment-1767531204, for v2.5, MacOS compilation failed due to "no member named 'make_unique' in namespace 'std'" as below:
```
>>> compiling thrift file dsn.layer2.thrift ...
mkdir /Users/runner/work/incubator-pegasus/incubator-pegasus/idl/output
/Users/runner/work/incubator-pegasus/incubator-pegasus/thirdparty/output/bin/thrift -gen cpp:moveable_types -out output dsn.layer2.thrift
mv output/dsn.layer2_types.h /Users/runner/work/incubator-pegasus/incubator-pegasus/src/common/serialization_helper
mv output/dsn.layer2_types.cpp /Users/runner/work/incubator-pegasus/incubator-pegasus/src/runtime
rm -rf /Users/runner/work/incubator-pegasus/incubator-pegasus/idl/output

done
Running cmake Pegasus...
~/work/incubator-pegasus/incubator-pegasus/build/release ~/work/incubator-pegasus/incubator-pegasus ~/work/incubator-pegasus/incubator-pegasus
CMake Warning (dev) at CMakeLists.txt:18 (project):
  cmake_minimum_required() should be called prior to this top-level project()
  call.  Please see the cmake-commands(7) manual for usage documentation of
  both commands.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- The C compiler identification is AppleClang 14.0.0.14000029
-- The CXX compiler identification is AppleClang 14.0.0.14000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
Running CompilerInfo.cmake
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Applications/Xcode_14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

Selected compiler clang 14.0
[...]

PR #1648, opened 2023-10-19
fix(build): fix compilation error on MacOS for C++14

empiredan:
https://github.com/apache/incubator-pegasus/issues/1647
 
The standard followed by [v2.5](https://github.com/apache/incubator-pegasus/tree/v2.5)
would be C++14. However, Compilation on MacOS failed due to some errors, such as:
- no member named 'make_unique' in namespace 'std'
- initialized lambda captures are a C++14 extension
- 'ext/alloc_traits.h' file not found

Most of the errors are about C++14. `cmake` directives should be fixed to make C++14
compilation on MacOS work well.

### DIFF
--- a/cmake_modules/BaseFunctions.cmake
+++ b/cmake_modules/BaseFunctions.cmake
@@ -204,7 +204,7 @@ function(dsn_setup_compiler_flags)
   # We want access to the PRI* print format macros.
   add_definitions(-D__STDC_FORMAT_MACROS)
 
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y -gdwarf-4" CACHE STRING "" FORCE)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -gdwarf-4" CACHE STRING "" FORCE)
 
   #  -Wall: Enable all warnings.
   add_compile_options(-Wall)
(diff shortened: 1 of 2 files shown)


empiredan:
`#include <ext/alloc_traits.h>` should be declared as `no_include` with IWYU due to following compilation error on MacOS:
```
[ 43%] Building CXX object src/meta/CMakeFiles/dsn_meta_server.dir/meta_service_app.cpp.o
/Users/runner/work/incubator-pegasus/incubator-pegasus/src/meta/meta_service.cpp:29:10: fatal error: 'ext/alloc_traits.h' file not found
#include <ext/alloc_traits.h>
         ^~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [src/meta/CMakeFiles/dsn_meta_server.dir/meta_service.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

acelyc111 (approved):

empiredan:
Use `target_compile_features` by referring to the following docs:

- https://stackoverflow.com/questions/40452287/let-cmake-with-clang-use-c11-c14
- https://stackoverflow.com/questions/41220265/check-if-a-c11-feature-is-enabled-in-compiler-with-cmake/41221098#41221098
- https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html

Issue #1665, opened 2023-10-31
Bug(manual_compact):replica lose manual compact finished status after replica migrate


ninsmiracle:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
-Set a suitable  time and let  a table finish manual compact
-View the progress of manual compaction,now is 100
-Stop a node
-View the progress of manual compaction again ,now is 79
-Go to the alive nodes and view their logs,some replicas begin manual compact
-View the progress of manual compaction again ,now is 91
-However some replica DO NOT manual compaction this day manual compaction progress stack in 91

So that , in our online envs,some pegasus users can not finish manual compact after replica migrate. 

2. What did you expect to see?
Replicas which finished manual compact can hold the status after replica migrate.

3. What version of Pegasus are you using?
[pegasus2.4](https://github.com/apache/incubator-pegasus/tree/v2.4)

abdraheem98:
Is the issue still open? I'm interested to solve this issue. Could please tell some more details. Thanks in Advance

acelyc111:
@ninsmiracle Each replica of a partition will do manual compaction independently, when one node shutdown, the other replicas on other nodes may not start or in progress manual compaction.
What tools you were using to check the progress, could you leave some details? Thanks!

ninsmiracle:
I used `admin-cli` to check the manaul compaction progress. And this is the result I use `manual-compaction query -a TABLE_NAME` commond after manual compact begin or node stop:
 
![image](https://github.com/apache/incubator-pegasus/assets/110282526/d819feb0-884d-4b3d-9840-047d60aa27a3)

And we can use pegasus http interface to check the replicas compact detail of target table.
![image](https://github.com/apache/incubator-pegasus/assets/110282526/85d70398-73ac-4874-a054-53d22d208cdf)

ninsmiracle:
> Is the issue still open? I'm interested to solve this issue. Could please tell some more details. Thanks in Advance

This is my way to recurrent the issus:

ninsmiracle:
> Is the issue still open? I'm interested to solve this issue. Could please tell some more details. Thanks in Advance

This is my way to recurrent this bug:

1. Create a Pegasus table and input some data.
[...]

PR #1666, opened 2023-10-31
Fix(manual_compact):fix replica lose manual compact finished status after replica migrate bug

ninsmiracle:
#1665 

I add another string in `meta_store` (meta_store is a column families which persist pegasus_last_manual_compact_finish_time and so on).
So that we can read the value from meta_store when we recover a replica after replica migrate.

##### Tests 
- Manual test 
1.Create table,and put some data in it.
2.Let table begin to do manual compaction
3.Finish manual compaction and view the replica relationship of this table.
4.Stop a node,and check the progress of compaction.
5.Wait a few minutes,and check if still have some replica can not show a Finish status.

Test Result
```
Pegasus-AdminCli-1.2.0 » manual-compaction query -a gns
Table gns manual compaction progress is 100
Pegasus-AdminCli-1.2.0 » manual-compaction query -a gns
Table gns manual compaction progress is 83
Pegasus-AdminCli-1.2.0 » manual-compaction query -a gns
Table gns manual compaction progress is 83
Pegasus-AdminCli-1.2.0 » manual-compaction query -a gns
Table gns manual compaction progress is 100
```

##### Code changes

- try to read `LAST_MANUAL_COMPACT_USED_TIME` from meta_store when DB exist.
- Set `LAST_MANUAL_COMPACT_USED_TIME` to zero when DB not exist.
- Set `LAST_MANUAL_COMPACT_USED_TIME` to the time replica last used.
- Generate a new checkpoint after we change LAST_MANUAL_COMPACT_USED_TIME in func  `do_manual_compact`

### DIFF
--- a/collector/avail/detector.go
+++ b/collector/avail/detector.go
@@ -34,7 +34,7 @@ import (
 
 // Detector periodically checks the service availability of the Pegasus cluster.
 type Detector interface {
-	Start(tom *tomb.Tomb) error
+	Run(tom *tomb.Tomb) error
 }
 
 // NewDetector returns a service-availability detector.
@@ -96,7 +96,7 @@ type pegasusDetector struct {
 	partitionCount int
 }
 
-func (d *pegasusDetector) Start(tom *tomb.Tomb) error {
+func (d *pegasusDetector) Run(tom *tomb.Tomb) error {
 	var err error
 	// Open the detect table.
 	d.detectTable, err = d.client.OpenTable(context.Background(), d.detectTableName)
--- a/collector/config.yml
+++ b/collector/config.yml
@@ -44,3 +44,6 @@ falcon_agent:
 
 availablity_detect:
   table_name : test
+
+hotspot:
+  partition_detect_interval : 10s
--- a/collector/hotspot/partition_detector.go
+++ b/collector/hotspot/partition_detector.go
@@ -16,3 +16,43 @@
 // under the License.
 
 package hotspot
+
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"gopkg.in/tomb.v2"
+)
+
+type PartitionDetector interface {
+	Run(tom *tomb.Tomb) error
+}
+
+type PartitionDetectorConfig struct {
+	DetectInterval time.Duration
+}
+
+func NewPartitionDetector(conf PartitionDetectorConfig) PartitionDetector {
+	return &partitionDetector{
+		detectInterval: conf.DetectInterval,
+	}
+}
+
+type partitionDetector struct {
+	detectInterval time.Duration
+}
+
+func (d *partitionDetector) Run(tom *tomb.Tomb) error {
+	for {
+		select {
+		case <-time.After(d.detectInterval):
+			d.detect()
+		case <-tom.Dying():
+			log.Info("Hotspot partition detector exited.")
+			return nil
+		}
+	}
+}
+
+func (d *partitionDetector) detect() {
+}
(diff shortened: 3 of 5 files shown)


acelyc111:
I’m confused how can this patch fix this issue, I guess the reason the compaction progress not update is caused by the [serving](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_data.h#L217C28-L217C28) and the [progress](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_data.cpp#L543) is not well updated after a node shutdown.

ninsmiracle:
> I’m confused how can this patch fix this issue, I guess the reason the compaction progress not update is caused by the [serving](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_data.h#L217C28-L217C28) and the [progress](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_data.cpp#L543) is not well updated after a node shutdown.

Yes. And We can see the current logic how a replica judge itself finish compact this round or not:
```
if (start_time_ms > 0) {
    return dsn::replication::manual_compaction_status::RUNNING;
} else if (enqueue_time_ms > 0) {
    return dsn::replication::manual_compaction_status::QUEUING;
} else if (last_time_used_ms > 0 && last_finish_time_ms > 0) {
    return dsn::replication::manual_compaction_status::FINISHED;
} else {
    return dsn::replication::manual_compaction_status::IDLE;
}
```
When replica restart or migrate,`last_time_used_ms` will loss.
And this replica will try to enter task queue again,however compaction's time checker prevent doing it util next day.
So I add `LAST_MANUAL_COMPACT_USED_TIME` and persist this parameter to colum families. So that this replica can be judged status `finish` after replica migrate.

acelyc111:
> > I’m confused how can this patch fix this issue, I guess the reason the compaction progress not update is caused by the [serving](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_data.h#L217C28-L217C28) and the [progress](https://github.com/apache/incubator-pegasus/blob/master/src/meta/meta_data.cpp#L543) is not well updated after a node shutdown.
> 
> Yes,the progress is not well updated after replica restart. And We can see the current logic how a replica judge itself finish compact this round or not:
> 
> ```
> if (start_time_ms > 0) {
>     return dsn::replication::manual_compaction_status::RUNNING;
> } else if (enqueue_time_ms > 0) {
>     return dsn::replication::manual_compaction_status::QUEUING;
> } else if (last_time_used_ms > 0 && last_finish_time_ms > 0) {
>     return dsn::replication::manual_compaction_status::FINISHED;
> } else {
>     return dsn::replication::manual_compaction_status::IDLE;
> }
> ```
> 
> When replica restart or migrate,`last_time_used_ms` will loss. And this replica will try to enter task queue again,however compaction's time checker prevent doing it util next day. So I add `LAST_MANUAL_COMPACT_USED_TIME` and persist this parameter to colum families. So that this replica can be judged status `finish` after replica migrate.

How to comprehend `compaction's time checker prevent doing it util next day`, do you mean the logic in `bool pegasus_manual_compact_service::check_manual_compact_state()`, in this case, transfer the status from `manual_compaction_status::IDLE` to `manual_compaction_status::FINISHED` may help?

ninsmiracle:
> pegasus_manual_compact_service::check_manual_compact_state

  Yes, `pegasus_manual_compact_service::check_manual_compact_state()` will prevent compaction log doing with this replica util next day. 
  So, if we have the parameter `last_finish_time_ms`, `pegasus_manual_compact_service::query_compact_status` will response the caller that 'this replica is FINISH manual compaction'.
  In other words, I do not think we should transfer the `status`,just need tell the caller when it send a query request.

PR #1667, opened 2023-10-31
fix: parse timestamp from the name of data dir for gc instead of the last update time

empiredan:
https://github.com/apache/incubator-pegasus/issues/1673

It's found that sometimes the data directories of replicas are removed
immediately after they are renamed with postfixes `.err/.gar`, though
actually both of `gc_disk_error_replica_interval_seconds` and
`gc_disk_garbage_replica_interval_seconds` have been configured
with at least one day.

The reason is that the base time for expiration time is **the last write
time**, that is, `st_mtime` within `struct stat` returned by `stat()`.
Once a long time has passed since the last write time, the data directory
will be removed immediately after it is renamed with postfixes `.err/.gar`.

To fix this problem, just use the timestamp within the directory name as
the base time that is generated when the data directory is renamed with
postfixes `.err/.gar`. The last update time would be used iff the timestamp
is NOT found within the directory name.


src/replica/disk_cleaner.cpp
```diff
@@ -70,6 +75,91 @@ const std::string kFolderSuffixBak = ".bak";
...
+// This timestamp could be used as the minimum, since it's far earlier than the time when
+// Pegasus was born.
+#define MIN_TIMESTAMP_US 1262275200000000
+#define MIN_TIMESTAMP_US_LENGTH (sizeof(STRINGIFY(MIN_TIMESTAMP_US)) - 1)
+
+bool parse_timestamp_us(const std::string &name, size_t suffix_size, uint64_t &timestamp_us)
+{
+    // Examples of the directory names of faulty or dropped replicas could be:
+    // 1.1.pegasus.1698843209235962.err
+    // 2.1.pegasus.1698843214240709.gar
```
acelyc111, line 104: It would better to describe the parameters above the line of function declaration.

Besides, give describe what the `suffix` is (to understand what `suffix_size` would be), and example of `timestamp_us`

src/replica/disk_cleaner.cpp
```diff
@@ -70,6 +75,91 @@ const std::string kFolderSuffixBak = ".bak";
 const std::string kFolderSuffixOri = ".ori";
 const std::string kFolderSuffixTmp = ".tmp";
 
+namespace {
+
+bool get_expiration_seconds_by_last_write_time(const std::string &path,
```
acelyc111, line 80: Instead of `seconds`, maybe `timestamp` would be more meaningful.

src/replica/disk_cleaner.cpp
```diff
@@ -87,59 +177,85 @@ error_s disk_remove_useless_dirs(const std::vector<std::shared_ptr<dir_node>> &d
...
-                     fpath,
-                     last_write_time + remove_interval_seconds - current_time_ms / 1000);
+                     path,
+                     expiration_seconds - current_time_ms / 1000);
         }
     }
     return error_s::ok();
 }
 
+bool is_data_dir_removable(const std::string &dir)
```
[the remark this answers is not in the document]
empiredan: OK, I'll add test cases for both `is_data_dir_removable(bool)` and `is_data_dir_invalid(bool)`.

src/replica/disk_cleaner.cpp
empiredan: OK.

src/replica/disk_cleaner.cpp
empiredan: OK, I'll replace all "expiration seconds” with "timestamp".

src/replica/disk_cleaner.cpp
```diff
@@ -70,6 +75,108 @@ const std::string kFolderSuffixBak = ".bak";
...
+    if (!dsn::utils::filesystem::last_write_time(path, last_write_time_s)) {
+        LOG_WARNING("gc_disk: failed to get last write time of {}", path);
+        return false;
+    }
+
+    expiration_timestamp_s = static_cast<uint64_t>(last_write_time_s) + delay_seconds;
+    return true;
+}
+
+// Unix timestamp in microseconds for 2010-01-01 00:00:00.
```
acelyc111, line 98: nit: note which time zone (CST) it is.

src/replica/disk_cleaner.cpp
```diff
@@ -70,6 +75,108 @@ const std::string kFolderSuffixBak = ".bak";
...
+// There are only 2 kinds of directory names that could include timestamp: one is the faulty
+// replicas whose name has suffix ".err"; another is the dropped replicas whose name has
+// suffix ".gar". The examples for both kinds of directory names:
+//     1.1.pegasus.1698843209235962.err
+//     1.2.pegasus.1698843214240709.gar
+//
+// Specify the size of suffix by `suffix_size`. For both kinds of names (.err and .gar),
+// `suffix_size` is 4.
+//
+// The timestamp is the number the number just before the suffix, between the 2 dots. For
```
acelyc111, line 115: Duplicate `the number`?

acelyc111 (approved):

acelyc111 (approved):

GehaFearless (approved):

PR #1668, opened 2023-11-01
fix(log): not redacted hotkey log

Samunroyu:
The hotkey log should not redacted when encrytion are at set.

Modify ‘c_escape_sensitive_string’ back to ‘c_escape_string’

- Unit test

### DIFF
--- a/src/server/hotkey_collector.cpp
+++ b/src/server/hotkey_collector.cpp
@@ -196,7 +196,7 @@ inline void hotkey_collector::change_state_by_result()
         if (!_result.hot_hash_key.empty()) {
             change_state_to_finished();
             LOG_ERROR_PREFIX("Find the hotkey: {}",
-                             pegasus::utils::c_escape_sensitive_string(_result.hot_hash_key));
+                             pegasus::utils::c_escape_string(_result.hot_hash_key));
         }
         break;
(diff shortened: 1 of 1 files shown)


acelyc111 (approved):

src/server/hotkey_collector.cpp
```diff
@@ -314,7 +314,7 @@ void hotkey_collector::query_result(dsn::replication::detect_hotkey_response &re
         LOG_INFO_PREFIX(hint);
     } else {
         resp.err = dsn::ERR_OK;
-        resp.__set_hotkey_result(pegasus::utils::c_escape_sensitive_string(_result.hot_hash_key));
+        resp.__set_hotkey_result(pegasus::utils::c_escape_string(_result.hot_hash_key));
```
empiredan, line 317:
```suggestion
        // Hot key should not be encrypted, thus use `c_escape_string` instead of
        // `c_escape_sensitive_string` (otherwise it would be overwritten with
        // "<redacted>").
        resp.__set_hotkey_result(pegasus::utils::c_escape_string(_result.hot_hash_key));
```

acelyc111 (approved):

empiredan (approved):

PR #1701, opened 2023-11-29
feat(ddl_client): add failure logs when backup or restore commands failed

acelyc111:
- Print `hint_message` when `add_backup_policy`, `disable_backup_policy`,
  `enable_backup_policy` or `update_backup_policy` failed. The hint messages
  are currently used in admin-cli only.
- Return corresponding error code when `query_restore` failed.

### DIFF
--- a/src/client/replication_ddl_client.cpp
+++ b/src/client/replication_ddl_client.cpp
@@ -1048,6 +1048,7 @@ dsn::error_code replication_ddl_client::add_backup_policy(const std::string &pol
     ::dsn::unmarshall(resp_task->get_response(), resp);
 
     if (resp.err != ERR_OK) {
+        std::cout << "add backup policy failed: " << resp.hint_message << std::endl;
         return resp.err;
     } else {
         std::cout << "add backup policy succeed, policy_name = " << policy_name << std::endl;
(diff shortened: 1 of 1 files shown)


GehaFearless (approved):

src/client/replication_ddl_client.cpp
```diff
@@ -1350,6 +1354,8 @@ dsn::error_code replication_ddl_client::query_restore(int32_t restore_app_id, bo
     } else if (response.err == ERR_APP_DROPPED) {
         std::cout << "restore failed, because some partition's data is damaged on cold backup media"
                   << std::endl;
+    } else {
```
empiredan, line 1357: Is it necessary to print some error message here ?

acelyc111: `configuration_query_restore_response` type of struct doesn't have a `hint_message` like field, and the caller (in `https://github.com/apache/incubator-pegasus/blob/master/src/shell/commands/cold_backup.cpp#L493`) will print the error message if failed.
So it's not necessary to print error message here.

empiredan: OK.

empiredan (approved):

PR #1755, opened 2023-12-06
fix(java-client): use junit5 instead of junit3/junit4

Samunroyu:
The test of java client are using junit3 and junit4 in the same time. It should change to use junit5 uniformly.

1. Change the java client test import and usage for uniformly junit5. Such as deprecated 'Assert' from usage of junit3. 
2. upgrader junit version to 5.9.3，use junit.api  and junit.jupiter.engine 
3. introduce org.hamcrest for some test
4. fix slf4j-api dependency prombles

#### Tests 

- Unit test 

#### Side effects
only java client unit test

#### Related changes
pom.xml have some jar dependency problems. Exclude slf4j-api from libthrift、zookeeper、metrics-core

### DIFF
--- a/java-client/pom.xml
+++ b/java-client/pom.xml
@@ -58,8 +58,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <compileSource>1.8</compileSource>
     <releaseTarget>8</releaseTarget>
-    <junit.engine.version>5.3.2</junit.engine.version>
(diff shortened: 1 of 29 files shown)


GehaFearless (approved):

acelyc111 (approved):
Thanks for the contribution!

acelyc111 (approved):

acelyc111:
The failed tests are not related to this patch, we can fix it later.

empiredan (approved):

PR #1760, opened 2023-12-08
feat(cpp-shell): make mlog_dump dump plog

acelyc111:
Since Pegasus 2.5.0, the slog has been deprecated entirely, the plog is left.
The 'mlog_dump' tool in cpp-shell is used to dump slog ever, this patch transfer
it to dump plog.

### DIFF
--- a/src/shell/commands/debugger.cpp
+++ b/src/shell/commands/debugger.cpp
@@ -64,6 +64,7 @@
 #include "utils/blob.h"
 #include "utils/filesystem.h"
 #include "utils/fmt_logging.h"
+#include "utils/strings.h"
 
 bool sst_dump(command_executor *e, shell_context *sc, arguments args)
 {
--- a/src/shell/commands/debugger.cpp
+++ b/src/shell/commands/debugger.cpp
@@ -81,7 +81,7 @@ bool mlog_dump(command_executor *e, shell_context *sc, arguments args)
                                            {0, 0, 0, 0}};
 
     bool detailed = false;
-    std::string slog_dir;
+    std::string plog_dir;
     std::string output;
     optind = 0;
     while (true) {
@@ -95,7 +95,7 @@ bool mlog_dump(command_executor *e, shell_context *sc, arguments args)
             detailed = true;
             break;
         case 'i':
-            slog_dir = optarg;
+            plog_dir = optarg;
             break;
         case 'o':
             output = optarg;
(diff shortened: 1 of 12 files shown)


src/shell/commands/debugger.cpp
```diff
@@ -78,7 +81,7 @@ bool mlog_dump(command_executor *e, shell_context *sc, arguments args)
                                            {0, 0, 0, 0}};
 
     bool detailed = false;
-    std::string input;
+    std::string slog_dir;
```
empiredan, line 84: Should be `plog_dir`?

src/shell/commands/debugger.cpp
```diff
@@ -101,12 +104,28 @@ bool mlog_dump(command_executor *e, shell_context *sc, arguments args)
...
+        fmt::print(stderr, "ERROR: '{}' is not a directory\n", plog_dir);
+        return false;
+    }
+
+    char splitters[] = {'\\', '/', 0};
+    auto slog_dir_tmp = plog_dir;
+    std::string name =
+        dsn::utils::get_last_component(dirname((char *)slog_dir_tmp.c_str()), splitters);
+    if (name.empty()) {
+        fmt::print(stderr, "ERROR: '{}' is not a valid slog directory\n", plog_dir);
```
empiredan, line 121:
```suggestion
        fmt::print(stderr, "ERROR: '{}' is not a valid plog directory\n", plog_dir);
```

src/shell/commands/debugger.cpp
```diff
@@ -101,12 +104,28 @@ bool mlog_dump(command_executor *e, shell_context *sc, arguments args)
...
+        return false;
+    }
+
+    char app_type[128];
+    int32_t app_id, pidx;
+    if (3 != sscanf(name.c_str(), "%d.%d.%s", &app_id, &pidx, app_type)) {
+        fmt::print(stderr, "ERROR: '{}' is not a valid slog directory\n", plog_dir);
```
empiredan, line 128:
```suggestion
        fmt::print(stderr, "ERROR: '{}' is not a valid plog directory\n", plog_dir);
```

src/shell/commands/debugger.cpp
```diff
@@ -101,12 +104,28 @@ bool mlog_dump(command_executor *e, shell_context *sc, arguments args)
...
-        fprintf(stderr, "ERROR: input %s is not a directory\n", input.c_str());
+    if (!dsn::utils::filesystem::directory_exists(plog_dir)) {
+        fmt::print(stderr, "ERROR: '{}' is not a directory\n", plog_dir);
+        return false;
+    }
+
+    char splitters[] = {'\\', '/', 0};
+    auto slog_dir_tmp = plog_dir;
+    std::string name =
+        dsn::utils::get_last_component(dirname((char *)slog_dir_tmp.c_str()), splitters);
```
empiredan, line 119:
```suggestion
    const auto replica_path = std::filesystem::path(plog_dir).parent_path();
    const auto name = replica_path.filename().string();
```

GehaFearless (approved):

PR #1763, opened 2023-12-11
build(images): append branch name to the tag of each build-env-* image to support building image for each branch

empiredan:
Since build-env-* images should be built for released versions and master branch, we could append branch name to the tag to create image for each branch.

### DIFF
--- a/.github/workflows/build-push-env-docker.yml
+++ b/.github/workflows/build-push-env-docker.yml
@@ -62,6 +62,6 @@ jobs:
           file: ./docker/pegasus-build-env/${{ matrix.dockertag }}/Dockerfile
           push: true
           tags: |
-            apache/pegasus:build-env-${{ matrix.dockertag }}
+            apache/pegasus:build-env-${{ matrix.dockertag }}-${{ github.ref_name }}
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
--- a/.github/workflows/build-push-env-docker.yml
+++ b/.github/workflows/build-push-env-docker.yml
@@ -27,7 +27,6 @@ on:
     paths:
       - 'docker/pegasus-build-env/**'
       - '.github/workflows/build-push-env-docker.yml'
-      - 'thirdparty/**'
 
   # for manually triggering workflow
   workflow_dispatch:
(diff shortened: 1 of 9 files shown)


empiredan:
Manual test has been run to test branch [build-env-image-tag-add-branch](https://github.com/apache/incubator-pegasus/tree/build-env-image-tag-add-branch):
https://github.com/apache/incubator-pegasus/actions/workflows/build-push-env-docker.yml
image

Images have been built on dockerhub:
https://hub.docker.com/r/apache/pegasus/tags?page=1&name=build-env
image

.github/workflows/build-push-env-docker.yml
```diff
@@ -62,6 +62,6 @@ jobs:
           file: ./docker/pegasus-build-env/${{ matrix.dockertag }}/Dockerfile
           push: true
           tags: |
-            apache/pegasus:build-env-${{ matrix.dockertag }}
+            apache/pegasus:build-env-${{ matrix.dockertag }}-${{ github.ref_name }}
```
acelyc111, line 65: https://github.com/apache/incubator-pegasus/blob/71eae1b7a2ec37cb88ee0f8530859fbb4a1c0380/docker/thirdparties-src/Dockerfile
and https://github.com/apache/incubator-pegasus/blob/71eae1b7a2ec37cb88ee0f8530859fbb4a1c0380/docker/thirdparties-bin/Dockerfile are using images like apache/pegasus:build-env-centos7, you should also consider this.

empiredan: OK, the images that are depending on build-env-* have also been updated.

docker/ci-env/Dockerfile
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-FROM apache/pegasus:build-env-centos7
+FROM apache/pegasus:build-env-centos7-master
```
acelyc111, line 18: This Dockerfile seems not used any more, it can be safely removed later.

acelyc111 (approved):

docker/ci-env/Dockerfile
empiredan: That's true, I've also not found where this image is referenced. I could create another pull requests to remove it.

GehaFearless (approved):

empiredan:
https://github.com/apache/incubator-pegasus/pull/1763#discussion_r1423802199 has been resolved in https://github.com/apache/incubator-pegasus/pull/1773.

PR #1776, opened 2023-12-18
feat(http): use scoped enum http_status_code instead of magic numbers

empiredan:
Previously `http_server` has used scoped enum `http_status_code` to
represent the status codes for http responses, while `http_client` has
still been using magic numbers directly.

This PR is to make all http APIs for both http client and server use the
unique type of status code:

- change the name of enumerators of `http_status_code` to Google Style,
i.e., from `ok` and `bad_request` to `kOk` and `kBadRequest`
- use `http_status_code` instead of magic numbers for http client
- support the conversion between enumerators and values of any type


src/http/http_status_code.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include "utils/enum_helper.h"
+#include "utils/fmt_logging.h"
+#include "utils/ports.h"
+
+namespace dsn {
+
+enum class http_status_code : size_t
+{
+    ok,                    // 200
```
acelyc111, line 33: How about naming as `kOk`?
https://google.github.io/styleguide/cppguide.html#Enumerator_Names

src/http/http_status_code.h
```diff
@@ -0,0 +1,66 @@
...
+    bad_request,           // 400
+    not_found,             // 404
+    internal_server_error, // 500
+    invalid,
+};
+
+std::string http_status_code_to_string(http_status_code code);
+
+template <typename TInt, typename = std::enable_if_t<std::is_integral_v<TInt>>>
+http_status_code http_status_code_from_int(TInt val)
```
acelyc111, line 44: Is the template class enum_helper_xxx help?

src/http/http_status_code.h
empiredan: OK.

src/http/http_status_code.h
empiredan: OK, I'll move this into enum helper.

acelyc111 (approved):

foreverneverer (approved):

Issue #1779, opened 2023-12-18
Improve the client libraries CI


acelyc111:
Now the various client libraries CI procedure will download old versions (e.g. 2.0, 2.1) of Pegasus server binaries (even from external repository).
In the case of adding new interface for both client and server side, the current implementation dosen't meet the requirement.
We should imptove this situation.

PR #1790, opened 2023-12-19
chore(go-client-ci): build the server instead of downloading it from external site

acelyc111:
https://github.com/apache/incubator-pegasus/issues/1779

- Separate format and lint jobs
- Build Pegasus server binaries in the CI procedure instead of downloading it from any site
- Fix lint and test failures
- improve readme


GehaFearless (dismissed):

go-client/session/radmin_rpc_types.go
```diff
@@ -15,7 +15,7 @@
 * KIND, either express or implied.  See the License for the
 * specific language governing permissions and limitations
 * under the License.
```
empiredan, line 17: Whether all of these lines should also be indented by one space ?

go-client/session/admin_rpc_types.go
```diff
@@ -15,7 +15,7 @@
 * KIND, either express or implied.  See the License for the
 * specific language governing permissions and limitations
 * under the License.
```
empiredan, line 17: Whether all of these lines should also be indented by one space ?

go-client/test.sh
```diff
@@ -1,46 +0,0 @@
...
-PEGASUS_PKG_URL="https://github.com/apache/incubator-pegasus/releases/download/v2.0.0/pegasus-tools-2.0.0-5d969e8-glibc2.12-release.tar.gz"
-
-# start pegasus onebox environment
-if [ ! -f $PEGASUS_PKG.tar.gz ]; then
-    wget --quiet $PEGASUS_PKG_URL
-    tar xf $PEGASUS_PKG.tar.gz
-fi
-cd $PEGASUS_PKG
-
-./run.sh start_onebox -m 2 -r 3 -w
```
empiredan, line 35: Shall we keep a script allowing the developers to run their tests locally, or just add some instructions in README ?

acelyc111: Added to README

go-client/session/radmin_rpc_types.go
acelyc111: It is generated by go-client/generator/main.go, I've updated the files.

go-client/session/admin_rpc_types.go
acelyc111: Same.

empiredan (approved):

GehaFearless (approved):

PR #1798, opened 2023-12-19
feat(run.sh): make it possible to build partial objects

acelyc111:
Now it's possible to build partial objects by specifying "-m obj1,obj2" parameter to `./run.sh build`.

### DIFF
--- a/.github/workflows/lint_and_test_cpp.yaml
+++ b/.github/workflows/lint_and_test_cpp.yaml
@@ -605,10 +605,10 @@ jobs:
           rm -rf hadoop-bin/share/doc
           rm -rf zookeeper-bin/docs
       - name: Compilation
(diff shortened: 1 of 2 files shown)


.github/workflows/lint_and_test_cpp.yaml
```diff
@@ -606,6 +606,7 @@ jobs:
         run: |
           ccache -p
           ccache -z
+          // TODO(yingchun): Append "-m dsn_utils_tests" to the command if not needed to pack server or tools, for example, the dependencies are static linked.
```
acelyc111, line 609:
```suggestion
        # TODO(yingchun): Append "-m dsn_utils_tests" to the command if not needed to pack server or tools, for example, the dependencies are static linked.
        run: |
          ccache -p
          ccache -z
```

GehaFearless (approved):

empiredan (approved):

PR #1808, opened 2023-12-21
refactor(ranger): Move ranger out of runtime

acelyc111:
This is a pure refactor patch without any function changes.
The patch is aim to reduce the size and dependencies of runtime library,
mainly move the "ranger" out of "runtime", including:
- Add a test named "dsn_ranger_tests"
- Move src/runtime/test/test_utils.h to src/runtime/test_utils.h so the
  other tests out of "runtime" is able to inlcude it easily
- Remove duplicate comments from **/CMakeLists.txt
- Move concat_path_unix_style() from meta_options to filesystem where is
  more reasonable, and  also reduce some unnecessary couple
- Make supported_mechanisms as a static member variable of negotiation
  class instead of a global variable which is more modernized


GehaFearless (dismissed):

src/runtime/security/negotiation.cpp
```diff
@@ -40,6 +37,8 @@ DSN_DEFINE_bool(security,
 DSN_DEFINE_bool(security, mandatory_auth, false, "wheter to do authertication mandatorily");
 DSN_TAG_VARIABLE(mandatory_auth, FT_MUTABLE);
 
+const std::set<std::string> negotiation::supported_mechanisms{"GSSAPI"};
```
empiredan, line 40:
```suggestion
const std::set<std::string> negotiation::kSupportedMechanisms{"GSSAPI"};
```

src/runtime/security/negotiation.h
```diff
@@ -54,6 +55,10 @@ class negotiation
     bool check_status(negotiation_status::type status, negotiation_status::type expected_status);
 
 protected:
+    // TODO(zlw):we can't get string list from cflags now,
+    //  so we should get supported mechanisms from config in the later
+    static const std::set<std::string> supported_mechanisms;
```
empiredan, line 60:
```suggestion
    static const std::set<std::string> kSupportedMechanisms;
```

empiredan (approved):

empiredan (approved):

foreverneverer (approved):

PR #1809, opened 2023-12-21
refactor(logging): drop the unnecessary interface, use `log` instead of `dsn_log` and improve the tests

empiredan:
Refactor the logging system, including:
- drop the unnecessary interface `logging_provider::dsn_logv()`
- change `dsn_log` or `dlog` to `log` for the naming of logging-related
macros, functions, classes, types and variables
- improve the tests for the logging and fix the bug that the whole
utils test directory is removed after utils tests are finished.

### DIFF
--- a/src/utils/test/logger.cpp
+++ b/src/utils/test/logger.cpp
@@ -25,7 +25,7 @@
  */
 
 #include <errno.h>
-#include <stdarg.h>
+#include <fmt/core.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <algorithm>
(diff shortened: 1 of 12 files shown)


src/utils/test/logger.cpp
```diff
@@ -36,99 +37,95 @@
...
 
-static void clear_files(vector<int> &log_index)
-{
-    char file[256] = {};
-    for (auto i : log_index) {
-        snprintf_p(file, 256, "log.%d.txt", i);
-        dsn::utils::filesystem::remove_path(string(file));
-    }
-}
+// Don't name the dir with "./test", otherwise the whole utils test dir would be removed.
```
acelyc111, line 75: 👍

acelyc111 (approved):

GehaFearless (approved):

PR #1816, opened 2023-12-22
fix: standby meta server exits abnormally with core dump after receiving the http request `/meta/cluster`

empiredan:
https://github.com/apache/incubator-pegasus/issues/1817

If meta server is built without testing(i.e. `./run.sh build` without `--test`),
the macro `MOCK_TEST` dedicated to test code should not be enabled,
which might lead to unexpected behaviors.

On the other hand, meta server should not exit abnormally with core
dump even if it is built for testing(i.e. `./run.sh build --test`). Once it
finds that balancer is not initialized by mocking, it should go through
the normal process to prevent from core dump.

### DIFF
--- a/src/meta/meta_http_service.cpp
+++ b/src/meta/meta_http_service.cpp
@@ -829,7 +829,10 @@ void meta_http_service::update_scenario_handler(const http_request &req, http_re
 bool meta_http_service::redirect_if_not_primary(const http_request &req, http_response &resp)
 {
 #ifdef MOCK_TEST
-    // Once MOCK_TEST is defined, the meta server must has been built with `./run.sh build --test`.
+    // To enable MOCK_TEST, the option BUILD_TEST for cmake should be opened by:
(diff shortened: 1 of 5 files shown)


src/meta/meta_http_service.cpp
```diff
@@ -827,12 +828,26 @@ void meta_http_service::update_scenario_handler(const http_request &req, http_re
 
 bool meta_http_service::redirect_if_not_primary(const http_request &req, http_response &resp)
 {
-#ifdef DSN_MOCK_TEST
-    return true;
+#ifdef MOCK_TEST
+    // Once MOCK_TEST is defined, the meta server must has been built with `./run.sh build --test`.
```
acelyc111, line 832: It would be better to say `BUILD_TEST` or `MOCK_TEST` is enabled, because it's possible to build by cmake directly, not using ./run.sh build.

acelyc111 (approved):

GehaFearless (approved):

PR #1822, opened 2024-01-02
feat(java-client): Introduce exponentialBackoff retry mechanism

Samunroyu:
Introduces a retry mechanism into the Java client to enhance user
convenience and robustness.

It can be enabled by calling `retryPolicy()`, `retryBaseInterval()`
and `retryMaxTimes()` in when construct ClientOptions objects.

### DIFF
--- a/java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
+++ b/java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
@@ -38,10 +38,11 @@ public class DefaultRetryPolicy implements RetryPolicy {
     if (now >= deadlineNanos) {
       return new RetryAction(RetryDecision.FAIL, Duration.ZERO, "request deadline reached");
     }
-    long timeoutMs = timeout.toMillis();
-    long retryDelayMs =
+    long timeoutNanos = timeout.toNanos();
+    long retryDelayNanos =
         Math.min(
-            timeoutMs > 3 ? timeoutMs / 3 : 1, TimeUnit.NANOSECONDS.toMillis(deadlineNanos - now));
-    return new RetryAction(RetryDecision.RETRY, Duration.ofNanos(retryDelayMs), "");
+            timeoutNanos > 3000000 ? timeoutNanos / 3000000 : 1,
+            TimeUnit.NANOSECONDS.toNanos(deadlineNanos - now));
+    return new RetryAction(RetryDecision.RETRY, Duration.ofNanos(retryDelayNanos), "");
   }
 }
--- a/java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
+++ b/java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package org.apache.pegasus.client.retry;
 
 import java.time.Duration;
--- a/java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
+++ b/java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
@@ -39,9 +39,9 @@ public class DefaultRetryPolicy implements RetryPolicy {
       return new RetryAction(RetryDecision.FAIL, Duration.ZERO, "request deadline reached");
     }
     long timeoutMs = timeout.toMillis();
-    long retryDelayMs =
+    long retryDelayNanos =
         Math.min(
-            timeoutMs > 3 ? timeoutMs / 3 : 1, TimeUnit.NANOSECONDS.toMillis(deadlineNanos - now));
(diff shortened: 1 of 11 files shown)


acelyc111:
Some suggestions to improve the commit message:
https://cbea.ms/git-commit/
https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,29 @@
+package org.apache.pegasus.client.retry;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import org.apache.pegasus.client.ClientOptions;
+
+/**
+ * The default retry policy, which is the only policy before we introduce the retry policy
+ * mechanism. It only considers the timeout value to calculate the retry delay, that's why\ we need
```
Apache9, line 9: Why there is '\' after the 'why' word? Typo?

java-client/src/main/java/org/apache/pegasus/client/retry/ExponentialBackoffRetryPolicy.java
```diff
@@ -0,0 +1,40 @@
...
+public class ExponentialBackoffRetryPolicy implements RetryPolicy {
+
+  private static final int[] RETRY_BACKOFF = {1, 2, 3, 5, 10, 20, 40, 100, 100, 100, 100, 200, 200};
+
+  private final long retryBaseIntervalMs;
+
+  private final int retryMaxTimes;
+
+  public ExponentialBackoffRetryPolicy(ClientOptions opts) {
+    this.retryBaseIntervalMs = opts.getRetryBaseInterval().toMillis();
```
Apache9, line 18: Better to add some comments and also some checks when setting retry base interval, as here we will convert the timeout to milliseconds, if user specify an interval less than 1 millisecond, we will not perform as expected...

Apache9:
I tried rerun the failed job but it still failed...

java-client/src/main/java/org/apache/pegasus/client/retry/ExponentialBackoffRetryPolicy.java
Samunroyu: > Better to add some comments and also some checks when setting retry base interval, as here we will convert the timeout to milliseconds, if user specify an interval less than 1 millisecond, we will not perform as expected...

sure. if user specify an interval less than 1 millisecond, it will set to defaultintervalms.

java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
Samunroyu: > Why there is '' after the 'why' word? Typo?

it typo. i fix it.

Samunroyu:
> I tried rerun the failed job but it still failed...

the failed job is failed at start pegasus cluster. not java client unit test. Rerun one more time the job successful.

acelyc111 (dismissed):

java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,47 @@
...
+  public RetryAction shouldRetry(int retries, long deadlineNanos, Duration timeout) {
+    long now = System.nanoTime();
+    if (now >= deadlineNanos) {
+      return new RetryAction(RetryDecision.FAIL, Duration.ZERO, "request deadline reached");
+    }
+    long timeoutMs = timeout.toMillis();
+    long retryDelayMs =
+        Math.min(
+            timeoutMs > 3 ? timeoutMs / 3 : 1, TimeUnit.NANOSECONDS.toMillis(deadlineNanos - now));
+    return new RetryAction(RetryDecision.RETRY, Duration.ofMillis(retryDelayMs), "");
```
Apache9, line 45: Since here we also pass a Duration in, let's use nanos instead of millis?

java-client/src/main/java/org/apache/pegasus/client/retry/ExponentialBackoffRetryPolicy.java
```diff
@@ -0,0 +1,64 @@
...
+  public RetryAction shouldRetry(int retries, long deadlineNanos, Duration timeout) {
+    if (retries >= retryMaxTimes) {
+      return new RetryAction(
+          RetryDecision.FAIL, Duration.ZERO, "max retry times " + retryMaxTimes + " reached");
+    }
+    long now = System.nanoTime();
+    if (now >= deadlineNanos) {
+      return new RetryAction(RetryDecision.FAIL, Duration.ZERO, "request deadline reached");
+    }
+    long normalIntervalMs =
```
Apache9, line 56: Use nanos too.

java-client/src/main/java/org/apache/pegasus/client/retry/DefaultRetryPolicy.java
Samunroyu: > Since here we also pass a Duration in, let's use nanos instead of millis?

So we should allow the user to specify the time interval for nano units?

Apache9: If we change to use nanos here, then we do not need to change the code here if we allow users to specify timeouts with smaller unit than millis, like micros and nanos in the future.
Whether to allow users to specify timeout with unit smaller than millis can be discussed later.

java-client/src/main/java/org/apache/pegasus/client/retry/ExponentialBackoffRetryPolicy.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import org.apache.pegasus.client.ClientOptions;
+
+/** A retry policy which generates retry interval with exponential backoff policy. */
+public class ExponentialBackoffRetryPolicy implements RetryPolicy {
+
+  private static final int[] RETRY_BACKOFF = {1, 2, 3, 5, 10, 20, 40, 100, 100, 100, 100, 200, 200};
+
+  private final long retryBaseIntervalMs;
```
Apache9, line 31: Just change this to retryBaseIntervalNanos...

Apache9 (dismissed):

acelyc111 (approved):

empiredan (approved):

empiredan (approved):

acelyc111 (approved):

Issue #1824, opened 2024-01-03
Complement the implementation of class host_port


GehaFearless:
A part of https://github.com/apache/incubator-pegasus/issues/1403.

Complement the implementation of class `host_port` for immediate use in the future.

PR #1825, opened 2024-01-03
feat(fqdn): add operator< and use shared_ptr for group_host_port() instead of raw pointer for host_port

GehaFearless:
Issue: https://github.com/apache/incubator-pegasus/issues/1824

Complement the implementation of class host_port for immediate use in the future.

### DIFF
--- a/src/runtime/rpc/rpc_host_port.cpp
+++ b/src/runtime/rpc/rpc_host_port.cpp
@@ -20,8 +20,8 @@
 #include <errno.h>
 #include <netdb.h>
 #include <netinet/in.h>
-#include <string.h>
 #include <sys/socket.h>
+#include <cstring>
 #include <memory>
 #include <unordered_set>
 #include <utility>
@@ -79,7 +79,7 @@ host_port::host_port(rpc_address addr)
         _port = addr.port();
     } break;
     case HOST_TYPE_GROUP: {
-        _group_host_port = new rpc_group_host_port(addr.group_address());
+        _group_host_port = std::make_shared<rpc_group_host_port>(addr.group_address());
     } break;
     default:
         break;
@@ -120,7 +120,7 @@ void host_port::reset()
         _port = 0;
         break;
     case HOST_TYPE_GROUP:
-        group_host_port()->release_ref();
+        _group_host_port = nullptr;
         break;
     default:
         break;
@@ -142,7 +142,6 @@ host_port &host_port::operator=(const host_port &other)
         break;
     case HOST_TYPE_GROUP:
         _group_host_port = other._group_host_port;
-        group_host_port()->add_ref();
         break;
     default:
         break;
@@ -157,9 +156,9 @@ std::string host_port::to_string() const
     case HOST_TYPE_IPV4:
         return fmt::format("{}:{}", _host, _port);
     case HOST_TYPE_GROUP:
-        return fmt::format("address group {}", group_host_port()->name());
+        return fmt::format("host_port group {}", group_host_port()->name());
     default:
-        return "invalid address";
+        return "invalid host_port";
     }
 }
 
@@ -167,9 +166,7 @@ void host_port::assign_group(const char *name)
 {
     reset();
     _type = HOST_TYPE_GROUP;
-    _group_host_port = new rpc_group_host_port(name);
-    // take the lifetime of rpc_uri_address, release_ref when change value or call destructor
-    _group_host_port->add_ref();
+    _group_host_port = std::make_shared<rpc_group_host_port>(name);
 }
 
 error_s host_port::resolve_addresses(std::vector<rpc_address> &addresses) const
--- a/src/runtime/rpc/rpc_host_port.h
+++ b/src/runtime/rpc/rpc_host_port.h
@@ -24,6 +24,7 @@
 // IWYU pragma: no_include <experimental/string_view>
 #include <functional>
 #include <iosfwd>
+#include <memory>
 #include <string>
 #include <string_view>
 #include <vector>
@@ -72,7 +73,7 @@ public:
         return os << hp.to_string();
     }
 
-    rpc_group_host_port *group_host_port() const
+    std::shared_ptr<rpc_group_host_port> group_host_port() const
     {
         CHECK_NOTNULL(_group_host_port, "group_host_port cannot be null!");
         return _group_host_port;
@@ -93,7 +94,7 @@ private:
     std::string _host;
     uint16_t _port = 0;
     dsn_host_type_t _type = HOST_TYPE_INVALID;
-    rpc_group_host_port *_group_host_port = nullptr;
+    std::shared_ptr<rpc_group_host_port> _group_host_port;
 };
 
 inline bool operator==(const host_port &hp1, const host_port &hp2)
@@ -118,6 +119,21 @@ inline bool operator==(const host_port &hp1, const host_port &hp2)
 
 inline bool operator!=(const host_port &hp1, const host_port &hp2) { return !(hp1 == hp2); }
 
+inline bool operator<(const host_port &hp1, const host_port &hp2)
+{
+    if (hp1.type() != hp2.type()) {
+        return hp1.type() < hp2.type();
+    }
+
+    switch (hp1.type()) {
+    case HOST_TYPE_IPV4:
+        return hp1.host() < hp2.host() || (hp1.host() == hp2.host() && hp1.port() < hp2.port());
+    case HOST_TYPE_GROUP:
+        return hp1.group_host_port().get() < hp2.group_host_port().get();
+    default:
+        return true;
+    }
+}
 } // namespace dsn
 
 USER_DEFINED_STRUCTURE_FORMATTER(::dsn::host_port);
@@ -132,7 +148,7 @@ struct hash<::dsn::host_port>
         case HOST_TYPE_IPV4:
             return std::hash<std::string>()(hp.host()) ^ std::hash<uint16_t>()(hp.port());
         case HOST_TYPE_GROUP:
-            return std::hash<void *>()(hp.group_host_port());
+            return std::hash<void *>()(hp.group_host_port().get());
         default:
             return 0;
         }
--- a/src/runtime/test/host_port_test.cpp
+++ b/src/runtime/test/host_port_test.cpp
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-#include <string.h>
+#include <memory>
 #include <string>
 #include <utility>
 #include <vector>
@@ -45,12 +45,12 @@ TEST(host_port_test, host_port_to_string)
 {
     {
         host_port hp = host_port("localhost", 8080);
-        ASSERT_EQ(std::string("localhost:8080"), hp.to_string());
+        ASSERT_EQ("localhost:8080", hp.to_string());
     }
 
     {
         host_port hp;
-        ASSERT_EQ(std::string("invalid address"), hp.to_string());
+        ASSERT_EQ("invalid host_port", hp.to_string());
     }
 }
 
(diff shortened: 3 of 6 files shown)


src/runtime/rpc/rpc_host_port.cpp
```diff
@@ -157,12 +159,20 @@ std::string host_port::to_string() const
...
     case HOST_TYPE_GROUP:
-        return fmt::format("address group {}", group_host_port()->name());
+        return fmt::format("host_port group {}", group_host_port()->name());
     default:
-        return "invalid address";
+        return "invalid host_port";
     }
 }
 
+static __thread fixed_size_buffer_pool<8, 256> bf;
```
acelyc111, line 168: The entire hostname, including the delimiting dots, has a maximum of 253 ASCII characters, so 256 is not enough for host and port (5 bytes at maximum)
> https://en.wikipedia.org/wiki/Hostname

src/runtime/rpc/rpc_host_port.h
```diff
@@ -85,6 +85,8 @@ class host_port
     // This function is used for validating the format of string like "localhost:8888".
     bool from_string(const std::string &s);
 
+    const char *get_char_str() const;
```
acelyc111, line 88: How will this function be used? Is to_string() meet the requirement？

src/runtime/rpc/rpc_host_port.h
```diff
@@ -118,6 +120,22 @@ inline bool operator==(const host_port &hp1, const host_port &hp2)
 
 inline bool operator!=(const host_port &hp1, const host_port &hp2) { return !(hp1 == hp2); }
 
+inline bool operator<(const host_port &hp1, const host_port &hp2)
```
acelyc111, line 123: Add some tests.

src/runtime/rpc/rpc_host_port.cpp
```diff
@@ -80,6 +81,7 @@ host_port::host_port(rpc_address addr)
     } break;
     case HOST_TYPE_GROUP: {
         _group_host_port = new rpc_group_host_port(addr.group_address());
+        _group_host_port->add_ref();
```
acelyc111, line 84: What if not adding this?

src/runtime/test/host_port_test.cpp
```diff
@@ -114,6 +118,15 @@ TEST(host_port_test, rpc_group_host_port)
     rpc_group_host_port *g = hp_grp.group_host_port();
     ASSERT_EQ(std::string("test_group"), g->name());
 
+    host_port hp_grp1;
+    hp_grp1.assign_group("test_group");
+    if (reinterpret_cast<uint64_t>(hp_grp.group_host_port()) <
+        reinterpret_cast<uint64_t>(hp_grp1.group_host_port())) {
+        ASSERT_TRUE(hp_grp < hp_grp1);
+    } else {
+        ASSERT_FALSE(hp_grp < hp_grp1);
```
empiredan:
```suggestion
        ASSERT_GE(hp_grp, hp_grp1);
```

src/runtime/rpc/rpc_host_port.h
empiredan, line 123: Where would `operator<` be used ?

src/runtime/rpc/rpc_host_port.cpp
```diff
@@ -80,6 +81,7 @@ host_port::host_port(rpc_address addr)
     } break;
     case HOST_TYPE_GROUP: {
         _group_host_port = new rpc_group_host_port(addr.group_address());
```
empiredan, line 83: Are there some special reasons why a raw pointer should be used here ? Consider using the object instead ?

If a pointer is needed, consider `std::unique_ptr`, or use `dsn::ref_ptr` to manage a `ref_counter` object at least ?

src/runtime/rpc/rpc_host_port.cpp
GehaFearless: memory leak

src/runtime/rpc/rpc_host_port.h
GehaFearless: thrift struct `optional map<host_port, xxx>`

src/runtime/rpc/rpc_host_port.cpp
GehaFearless: `ref_counter` is super class of `rpc_group_host_port`.

src/runtime/rpc/rpc_host_port.cpp
```diff
@@ -157,19 +157,28 @@ std::string host_port::to_string() const
...
     default:
-        return "invalid address";
+        return "invalid host_port";
     }
 }
 
+// hostname has a maximum of 253 ASCII characters
+// uint_16 port maximum is 65535
+// 253 + 1(colon) + 5 = 259
+static __thread fixed_size_buffer_pool<8, 259> bf;
```
acelyc111, line 169: It would be better to align the memory address, say 64*N, i.e. 320.

src/runtime/test/host_port_test.cpp
```diff
@@ -166,6 +172,41 @@ TEST(host_port_test, rpc_group_host_port)
     ASSERT_FALSE(g->contains(hp2));
     ASSERT_EQ(0u, g->members().size());
     ASSERT_EQ(invalid_hp, g->leader());
+
+    // operator <
+    host_port hp_grp1;
+    hp_grp1.assign_group("test_group");
+    if (reinterpret_cast<uint64_t>(hp_grp.group_host_port().get()) <
```
acelyc111, line 179: The relationship between hp_grp and hp_grp1 definite IMO, so the judgement is not needed.

src/runtime/rpc/rpc_host_port.cpp
```diff
@@ -157,19 +157,28 @@ std::string host_port::to_string() const
...
+const char *host_port::get_char_str() const
```
empiredan, line 170: When `get_char_str()` would be used ? Once it is called beyond 8 times in the same thread, the content of the buffer might be updated, is this expected ?

src/runtime/test/host_port_test.cpp
GehaFearless: The relationship between hp_grp and hp_grp1 not predictable for my test. I think maybe malloc not usd space ordered.

src/runtime/rpc/rpc_host_port.cpp
GehaFearless: e.g.
`sprintf("%s", get_char_str())`
`strcpy(dest, get_char_str())`

If we just have a funcation return temporary vaule not enough.

Initially, I thought that a buffer for storing one string would be sufficient.
Later I found `fixed_size_buffer_pool` from class `rpc_address` and thought it was more suitable.
Updated buffer not have influence on use.

acelyc111 (dismissed):

src/runtime/rpc/rpc_host_port.cpp
empiredan: OK. Please add some comments to illustrate our special scenarios like `log_prefix()`:
企业微信截图_f370bcab-9324-4855-9da9-3d2d9de2b155

Otherwise,  using `get_char_str` by mistake may lead to some bugs:
```
// The size of hosts are more than 8.
std::vector<host_port> hosts;
......
std::vector<const char *> strs;
for (auto &e: hosts) {
    strs.push_back(e.get_char_str());
}
```

src/utils/fixed_size_buffer_pool.h
```diff
@@ -28,8 +28,8 @@
 
 ///
 /// A simple buffer pool designed for efficiently formatting
-/// frequently used types (like gpid, rpc_address) into string,
-/// without dynamic memory allocation.
+/// frequently used types (like gpid, rpc_address, host_port)
+///  into string, without dynamic memory allocation.
```
acelyc111, line 32:
```suggestion
/// into string, without dynamic memory allocation.
```

src/runtime/test/host_port_test.cpp
```diff
@@ -100,6 +102,9 @@ TEST(host_port_test, operators)
     std::string hp_str2 = "pegasus:8080";
     ASSERT_FALSE(hp4.from_string(hp_str2));
     ASSERT_TRUE(hp4.is_invalid());
+
+    host_port hp5("localhost", 8081);
+    ASSERT_TRUE(hp < hp5);
```
acelyc111, line 107:
```suggestion
    ASSERT_LT(hp, hp5);
```

src/runtime/rpc/rpc_host_port.h
```diff
@@ -118,6 +130,22 @@ inline bool operator==(const host_port &hp1, const host_port &hp2)
...
+    if (hp1.type() != hp2.type()) {
+        return hp1.type() < hp2.type();
+    }
+
+    switch (hp1.type()) {
+    case HOST_TYPE_IPV4:
+        return hp1.host() < hp2.host() || (hp1.host() == hp2.host() && hp1.port() < hp2.port());
+    case HOST_TYPE_GROUP:
+        return reinterpret_cast<uint64_t>(hp1.group_host_port().get()) <
+               reinterpret_cast<uint64_t>(hp2.group_host_port().get());
```
acelyc111, line 144: Can be replaced by this?
```suggestion
        return hp1.group_host_port().get() < hp2.group_host_port().get();
```

src/runtime/rpc/rpc_host_port.cpp
acelyc111: I still doubt the necessarity of introducing it, what will happen if using:
```
fmt::printf("{}", to_std_string())
strcpy(dest, to_std_string().c_str())
```

src/runtime/test/host_port_test.cpp
```diff
@@ -166,6 +172,41 @@ TEST(host_port_test, rpc_group_host_port)
...
     ASSERT_EQ(invalid_hp, g->leader());
+
+    // operator <
+    host_port hp_grp1;
+    hp_grp1.assign_group("test_group");
+    if (reinterpret_cast<uint64_t>(hp_grp.group_host_port().get()) <
+        reinterpret_cast<uint64_t>(hp_grp1.group_host_port().get())) {
+        ASSERT_TRUE(hp_grp < hp_grp1);
+    } else {
+        ASSERT_FALSE(hp_grp < hp_grp1);
```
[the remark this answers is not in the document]
GehaFearless: `>=` not declare

src/runtime/rpc/rpc_host_port.h
GehaFearless: I think it's better to be specific.

src/runtime/rpc/rpc_host_port.cpp
```diff
@@ -151,25 +151,35 @@ host_port &host_port::operator=(const host_port &other)
...
     default:
-        return "invalid address";
+        return "invalid host_port";
     }
 }
 
+// hostname has a maximum of 253 ASCII characters
+// uint_16 port maximum is 65535
+// 253 + 1(colon) + 5 = 259
+// align the memory address, 256 < 259 < 320
```
empiredan, line 169:
```suggestion
// A host name has 253 ASCII characters at most. The max length of
// port(uint16_t, 65535) is 5. Thus the max length of "<host>:<port>"
// is 253 + 1(colon) + 5 = 259.
// 
// Set the size to 64 * 5 = 320 to be 64-bytes aligned.
```

empiredan (dismissed):

src/utils/fixed_size_buffer_pool.h
```diff
@@ -28,8 +28,8 @@
 
 ///
 /// A simple buffer pool designed for efficiently formatting
-/// frequently used types (like gpid, rpc_address) into string,
-/// without dynamic memory allocation.
+/// frequently used types (like gpid, rpc_address, host_port)
+/// into string, without dynamic memory allocation.
```
empiredan, line 32:
```suggestion
/// frequently used types (like gpid, rpc_address) into string,
/// without dynamic memory allocation.
```

acelyc111 (approved):

empiredan (approved):

PR #1826, opened 2024-01-03
feat: A new port is added for prometheus collection, and the indicator init…

limowang:
A new port is added for prometheus collection, and the indicator initialization process is extracted to form a separate function.


collector/go.mod
```diff
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-module github.com/pegasus-kv/collector
+module github.com/limowang/incubator-pegasus/collector
```
acelyc111, line 18:
```suggestion
module github.com/apache/incubator-pegasus/collector
```

collector/main
acelyc111, line 1: Don't add binaries to the codebase.

collector/main.go
```diff
@@ -26,9 +26,12 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/pegasus-kv/collector/avail"
-	"github.com/pegasus-kv/collector/metrics"
-	"github.com/pegasus-kv/collector/webui"
+	// "github.com/pegasus-kv/collector/avail"
+	// "github.com/pegasus-kv/collector/metrics"
+	// "github.com/pegasus-kv/collector/webui"
+	"github.com/limowang/incubator-pegasus/collector/avail"
```
acelyc111, line 32: Remove the comments and use the apache repository.

collector/main.go
```diff
@@ -81,7 +84,8 @@ func main() {
 		return
 	}
 
-	webui.StartWebServer()
+	metrics.InitMetrics()
+	//webui.StartWebServer()
```
acelyc111, line 88: Just remove it if it's not needed.

collector/webui/webserver.go
```diff
@@ -53,9 +56,22 @@ func StartWebServer() {
 	app.RegisterView(tmpl)
 
 	go func() {
-		err := app.Listen(":8080")
+		err := app.Listen(":8081")
```
acelyc111, line 59: It would be better to read from config file or command line rather than hard code.

The port should be user-defined to avoid port conflicts in user environments.

collector/metrics/metric_collector.go
```diff
@@ -167,17 +149,15 @@ func initMetrics() {
 						Help: desc,
 					}, []string{"endpoint", "role", "level", "title"})
 					GaugeMetricsMap[name] = *gaugeMetric
-				case "Percentile":
-					if _, ok := SummaryMetricsMap[name]; ok {
+				case "Percentile": //这个需要改动不能用这个表示,用gauge来表示分位数  --level(p50,p99),title(task_name)来替代区分
```
acelyc111, line 152: Use English because this is a globally open-source project.

acelyc111:
The PR title seems too long and incomplete.

collector/config.yml
```diff
@@ -25,7 +25,7 @@ meta_servers:
   - 127.0.0.1:34603
 
 # local server port
-port : 34101
+port : 8081
```
acelyc111, line 28: 8080, 8081 are widely used ports for web servers, it would be better to use 34101 to keep compatible (old version of collectors may have been used by communituy users, don't change the behavior).

collector/config.yml
```diff
@@ -34,7 +34,7 @@ metrics:
 
 prometheus:
   # the exposed port for prometheus exposer
-  exposer_port : 1111 
+  exposer_port : 8080
```
acelyc111, line 37: Same, and 34*01 ports are used by Pegasus by habit, how about use 34201 instead?

collector/webui/webserver.go
```diff
@@ -53,9 +57,25 @@ func StartWebServer() {
...
 	go func() {
-		err := app.Listen(":8080")
+		localPort := ":" + viper.GetString("port")
+		err := app.Listen(localPort)
 		if err != nil {
 			return
 		}
 	}()
+
+	//Provide metrics for prometheus
```
acelyc111, line 67:
```suggestion
	// Provide metrics for Prometheus.
```

Issue #1820, opened 2023-12-29
Feature(new_metrics): go-collector collect the metrics from meta server and replica


empiredan:
zion_profiler_RPC_RRDB_RRDB_GET_qps{cluster="skv_offline",host_name="hybrid01",pegasus_job="replica",port="8171",service="pegasus",skv_job="replica",app="",partition=""} 0.000000

replica_app_pegasus_get_qps{cluster="skv_offline",host_name="hybrid01",pegasus_job="replica",port="8171",service="pegasus",skv_job="replica",app="4",partition="2"} 0.000000

limowang:
![新图表 1](https://github.com/apache/incubator-pegasus/assets/64081036/987d0fd0-8430-4340-adf4-98d9ab35d865)

# Go-collector indicator collection process analysis    
## Process analysis
1. Metrics obtained from the server are stored in global variables in the metrics package. For example: GaugeMetricsMap map[string]prometheus.GaugeVec,CounterMetricsMap map[string]prometheus.CounterVec.In the main function, initialize the GaugeMetricsMap and CounterMetricsMap variables by calling InitMetrics() in the metrics package, and access the http indicator interface exposed by each server through getProMetricsByAddrs(addrs string) to initialize.   
2. Generate a Collector object that collects meta-server indicators by calling NewMetaServerMetricCollector(), and then call its function Start() to update the indicators regularly. ProcessAllServerMetrics(dsource) will be called in Start(). This function will periodically update the global variables GaugeMetricsMap and CounterMetricsMap that store prometheus format indicators from the http indicator interface exposed on the server side, so that prometheus monitoring can update data periodically.    
3. main() will finally call StartServer() in the webui package. The function of this function is: first generate a prometheus indicator registration center through NewRegister(), and then call MustRegister(prometheusMetricType) to add the GaugeVec variable in the global variable GaugeMetricsMap and the CounterMetricsMap CounterVec is registered in the prometheus indicator registration center, and finally the listening port is opened to allow the prometheus server to pull indicator data.

PR #1827, opened 2024-01-03
feat(new_metrics): add some basic fields to the response to metrics query

empiredan:
https://github.com/apache/incubator-pegasus/issues/1820

Labels are needed by metric models like Prometheus Data Model.
Therefore, cluster name, role name, host name and port should be
provided in the response to metrics query as the labels. All of these
fields would be added in the form of name/value pairs of json object.

Original entity array would be the value of a json object with name
"entities".

### DIFF
--- a/src/utils/metrics.cpp
+++ b/src/utils/metrics.cpp
@@ -22,14 +22,15 @@
 #include <boost/date_time/posix_time/posix_time_duration.hpp>
 #include <boost/system/error_code.hpp>
 #include <fmt/core.h>
-#include <new>
 #include <unistd.h>
+#include <new>
 
-#include "common/common.h"
 #include "http/http_method.h"
 #include "http/http_status_code.h"
 #include "runtime/api_layer1.h"
+#include "runtime/rpc/rpc_address.h"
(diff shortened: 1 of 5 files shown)


src/utils/metrics.cpp
```diff
@@ -449,6 +444,89 @@ metric_entity_ptr metric_registry::find_or_create_entity(const metric_entity_pro
...
+        }                                                                                          \
+    } while (0)
+
+void encode_cluster(dsn::metric_json_writer &writer)
+{
+    writer.Key(dsn::kMetricClusterField.c_str());
+
+    ENCODE_UNKNOWN_IF(utils::is_empty(dsn::FLAGS_cluster_name));
+
+    dsn::json::json_encode(writer, dsn::FLAGS_cluster_name);
```
acelyc111, line 465: How about encapsulate the 2 lines to a single line, the judgment can be put into the macro.

The other places are the same.

empiredan: OK.

acelyc111 (approved):

GehaFearless (approved):

Issue #1838, opened 2024-01-08
specify ambiguous meta_server_list on recovery_test 


GehaFearless:
## Bug Report

For easier configuration modifications, `recovery_test` just start one meta_server of onebox on `run.sh`.
`
if [ "${module}" == "recovery_test" ]; then master_count=1; fi
`
But the meta_server_list of class `recovery_test` is still three meta_servers.

Let fix it.

PR #1839, opened 2024-01-08
fix: specify ambiguous meta_server_list on recovery_test

GehaFearless:

Issue: https://github.com/apache/incubator-pegasus/issues/1838

match the behavior of A and B in meta_server_list

### DIFF
--- a/src/test/function_test/recovery/test_recovery.cpp
+++ b/src/test/function_test/recovery/test_recovery.cpp
@@ -65,6 +65,11 @@ protected:
     }
 
 public:
+    // For easier configuration modifications,
+    // select "single_master_cluster" cluster
+    // which onebox just start one master.
+    recovery_test() : test_util(std::map<std::string, std::string>(), "single_master_cluster") {}
+
     std::vector<dsn::rpc_address> get_rpc_address_list(const std::vector<int> ports)
     {
         std::vector<dsn::rpc_address> result;
--- a/src/test/function_test/recovery/test_recovery.cpp
+++ b/src/test/function_test/recovery/test_recovery.cpp
@@ -66,9 +66,8 @@ protected:
     }
 
 public:
-    // For easier configuration modifications,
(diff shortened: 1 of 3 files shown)


acelyc111 (dismissed):

acelyc111 (dismissed):

src/test/function_test/recovery/test_recovery.cpp
```diff
@@ -65,6 +66,11 @@ class recovery_test : public test_util
     }
 
 public:
+    // For easier configuration modifications,
+    // select "single_master_cluster" cluster
+    // which onebox just start one master.
```
empiredan, line 71:
```suggestion
    // The cluster name "single_master_cluster" (see src/test/function_test/config.ini) means the
    // cluster has only one meta server, while "onebox" means the cluster has 3 meta servers.
```

src/test/function_test/utils/test_util.cpp
```diff
@@ -62,12 +62,12 @@ using std::vector;
 
 namespace pegasus {
 
-test_util::test_util(map<string, string> create_envs)
+test_util::test_util(map<string, string> create_envs, std::string cluster_name)
```
empiredan, line 65:
```suggestion
test_util::test_util(map<string, string> create_envs, const std::string &cluster_name)
```

src/test/function_test/utils/test_util.h
```diff
@@ -49,7 +49,8 @@ class pegasus_client;
 class test_util : public ::testing::Test
 {
 public:
-    test_util(std::map<std::string, std::string> create_envs = {});
+    test_util(std::map<std::string, std::string> create_envs = {},
+              std::string cluster_name = "onebox");
```
empiredan, line 53:
```suggestion
              const std::string &cluster_name = "onebox");
```

acelyc111 (approved):

PR #1843, opened 2024-01-09
feat(http_client): add http_url class to help build URLs for http client

empiredan:
There are multiple components for a URL used to issue http requests,
such as scheme, host, port, path and query. However, to build each
URL by such components, we could assemble them into a string only
by fmt::format or concatenating them together with some separators.

This PR is to provide a class named http_url that help build URLs on
demand. Just specify each component we need to http_url, then
extract the URL string. We could set the extracted the URL string,
or just set http_url object itself to http client to tell what is the URL
for the http requests.

### DIFF
--- a/src/http/http_client.cpp
+++ b/src/http/http_client.cpp
@@ -58,6 +58,9 @@ dsn::error_s http_url::init()
         return dsn::error_s::make(dsn::ERR_CURL_FAILED, "fail to initialize curl url");
     }
 
+    // Set "http" as the default scheme.
+    set_scheme("http");
+
     return dsn::error_s::ok();
 }
 
--- a/src/http/http_client.cpp
+++ b/src/http/http_client.cpp
@@ -312,10 +312,13 @@ size_t http_client::on_response_data(const void *data, size_t length)
 
 dsn::error_s http_client::set_url(const std::string &new_url)
 {
-    // DO NOT call curl_easy_setopt() on CURLOPT_URL, since The CURLOPT_URL string is ignored
+    // DO NOT call curl_easy_setopt() on CURLOPT_URL, since the CURLOPT_URL string is ignored
     // if CURLOPT_CURLU is set. See following docs for details:
     // * https://curl.se/libcurl/c/CURLOPT_CURLU.html
     // * https://curl.se/libcurl/c/CURLOPT_URL.html
+    //
+    // Use a temporary object for the reason that once the error occurred, `_url` would not
+    // be dirty.
     http_url tmp;
     RETURN_NOT_OK(tmp.init());
     RETURN_NOT_OK(tmp.set_url(new_url.c_str()));
@@ -325,6 +328,7 @@ dsn::error_s http_client::set_url(const std::string &new_url)
 
 dsn::error_s http_client::set_url(const http_url &new_url)
 {
+    // Extract the URL string from new_url and attach it to the client.
     std::string str;
     RETURN_NOT_OK(new_url.to_string(str));
 
--- a/src/http/http_client.cpp
+++ b/src/http/http_client.cpp
@@ -50,11 +50,14 @@ DSN_DEFINE_uint32(http,
             #expr == #ok, code == (ok), "{}: {}", fmt::format(__VA_ARGS__), to_error_msg(code));   \
     } while (0)
 
-#define CHECK_IF_CURL_URL_SET_OK(url, part, content, ...)                                          \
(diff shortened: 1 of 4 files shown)


src/http/http_client.cpp
```diff
@@ -34,6 +34,135 @@ DSN_DEFINE_uint32(http,
                   "The maximum time in milliseconds that you allow the libcurl transfer operation "
                   "to complete");
 
+#define RETURN_IF_CURL_NOT(expr, expected, ...)                                                    \
+    do {                                                                                           \
+        const auto code = (expr);                                                                  \
+        if (dsn_unlikely(code != expected)) {                                                      \
+            std::string msg(fmt::format("{}: {}", fmt::format(__VA_ARGS__), to_error_msg(code)));  \
```
acelyc111, line 41: How about swap the places of error code and message, which is the behavior of other macros like LOG_AND_RETURN_NOT_TRUE, LOG_AND_RETURN_NOT_RDB_OK, etc.

src/http/http_client.cpp
```diff
@@ -34,6 +34,135 @@ DSN_DEFINE_uint32(http,
...
+
+http_url::http_url() : _url(nullptr) {}
+
+http_url::~http_url() { free_curlu_object(); }
+
+http_url::http_url(http_url &&rhs) : _url(rhs._url) { rhs._url = nullptr; }
+
+http_url &http_url::operator=(http_url &&rhs)
+{
+    _url = rhs._url;
```
acelyc111, line 54: Prevent self assignment.
if (this != &rhs) {...}

acelyc111, line 54: Should free `_url` at first?

Then add a test to check there is no memory leak.

src/http/http_client.cpp
```diff
@@ -34,6 +34,135 @@ DSN_DEFINE_uint32(http,
...
+dsn::error_s http_url::init()
+{
+    if (_url != nullptr) {
+        free_curlu_object();
+    }
+
+    _url = curl_url();
+    if (_url == nullptr) {
+        return dsn::error_s::make(dsn::ERR_CURL_FAILED, "fail to initialize curl url");
+    }
```
acelyc111, line 68: Looks like the macro in https://github.com/apache/incubator-pegasus/pull/1706/files#diff-4cc4ac5293103b26f79ab90be5de1bf7bbe24a37bbdcda43b2f85bbdd8be360eR39, how about submit the macro at first? @Samunroyu

src/http/test/http_client_test.cpp
```diff
@@ -32,10 +33,108 @@
 #include "utils/error_code.h"
 #include "utils/errors.h"
 #include "utils/fmt_logging.h"
+#include "utils/strings.h"
 #include "utils/test_macros.h"
 
 namespace dsn {
 
+using http_url_case =
+    std::tuple<const char *, const char *, uint16_t, const char *, const char *, const char *>;
```
acelyc111, line 42: It's too long, how about adding a new struct which contains all the fields, the struct member names are self-explanatory, and it would be helpful to simplify the following code.

src/http/http_client.cpp
Samunroyu: > Looks like the macro in https://github.com/apache/incubator-pegasus/pull/1706/files#diff-4cc4ac5293103b26f79ab90be5de1bf7bbe24a37bbdcda43b2f85bbdd8be360eR39, how about submit the macro at first? @Samunroyu

Ok.It's already merged. PR:: https://github.com/apache/incubator-pegasus/pull/1849/files

src/http/http_client.cpp
empiredan: `to_error_msg` will show underlying messages, thus we could keep it behind.

src/http/http_client.cpp
empiredan: OK, I've added for both copy and move assignment operators.

empiredan: I've wrapped it with `std::unique_ptr` to avoid the problem, and add test cases for copy/move constructors, copy/move assignment operators.

src/http/http_client.cpp
empiredan: OK.

src/http/test/http_client_test.cpp
empiredan: Good idea, I've used `struct` for all of the cases that need many arguments.

acelyc111 (dismissed):

acelyc111 (approved):

GehaFearless (approved):

PR #1857, opened 2024-01-17
fix(collector): fix the write overwrite caused by multiple collectors sharing a global variable DataSource

limowang:


empiredan (dismissed):

collector/metrics/metric_collector.go
```diff
@@ -53,8 +53,7 @@ var GaugeMetricsMap map[string]prometheus.GaugeVec
 var CounterMetricsMap map[string]prometheus.CounterVec
 var SummaryMetricsMap map[string]prometheus.Summary
 
-// DataSource 0 meta server, 1 replica server.
-var DataSource int
+// RoleByDataSource 0 meta server, 1 replica server.
```
acelyc111, line 56: It‘s corresponding to the variables in line 36, it would be better to mention the variable names instead of the magic numbers (0,1) again.

collector/metrics/metric_collector.go
```diff
@@ -124,10 +123,10 @@ func getReplicaAddrs() ([]string, error) {
 }
 
 // Register all metrics.
-func initMetrics() {
+func initMetrics(dataSource int) {
```
acelyc111, line 126: How about making both the initMetrics function (some other functions are the same) and dataSource as members of MetricCollector, so the dataSource would not appear everywhere.


The global variables are the same.
var GaugeMetricsMap map[string]prometheus.GaugeVec
var CounterMetricsMap map[string]prometheus.CounterVec
var SummaryMetricsMap map[string]prometheus.Summary
var RoleByDataSource map[int]string
var TableNameByID map[string]string

acelyc111: You can do the refactor in next patches, this PR looks good to me.

limowang: OK，i will try.

collector/metrics/metric_collector.go
```diff
@@ -67,22 +63,26 @@ func NewMetricCollector(
...
 	SummaryMetricsMap = make(map[string]prometheus.Summary, 128)
-	RoleByDataSource = make(map[int]string, 128)
 	TableNameByID = make(map[string]string, 128)
-	RoleByDataSource[0] = "meta"
-	RoleByDataSource[1] = "replica"
-	initMetrics()
 
-	return &Collector{detectInterval: detectInterval, detectTimeout: detectTimeout}
+	var role string
+	if dataSource == 0 {
```
empiredan, line 72: We could just change `dataSource` to `role`, and certainly change the type of `dataSource` to `string` instead of `int` which means, changing the type of following constants to `string`:
```go
const (
        MetaServer    int = 0
        ReplicaServer int = 1
)
```

collector/metrics/metric_collector.go
```diff
@@ -283,13 +277,13 @@ func updateClusterLevelTableMetrics(metricsByTableID map[string]Metrics) {
...
-			updateMetric(metric, "cluster", "table", tableName)
+			collector.updateMetric(metric, "cluster", "table", tableName)
 		}
 	}
 }
 
-func updateMetric(metric Metric, endpoint string, level string, title string) {
-	role := RoleByDataSource[DataSource]
+func (collector *Collector) updateMetric(metric Metric, endpoint string, level string, title string) {
+	role := collector.role
```
empiredan, line 286: Could the variable `role` be removed ?

empiredan (approved):

acelyc111 (approved):

PR #1860, opened 2024-01-18
fix(collector): Turn the variable registry in package webui into glob…

limowang:


collector/webui/webserver.go
```diff
@@ -27,14 +27,15 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+var Registry = prometheus.NewRegistry()
```
empiredan, line 30: Could the registry be created and returned by `StartWebServer()` to `main()`, thus registry would become a local variable ?

PR #1861, opened 2024-01-19
fix(collector): Change the location of the prometheus.Registerer

limowang:

- Change the location of the prometheus.Registerer variable definition.    
- Later, the prometheus.Registerer variable will be passed as a parameter to functions such as NewReplicaServerMetricCollector and NewMetaServerMetricCollector to reduce the coupling of registering metrics into prometheus.Registerer.

### DIFF
--- a/collector/main.go
+++ b/collector/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/apache/incubator-pegasus/collector/avail"
 	"github.com/apache/incubator-pegasus/collector/metrics"
 	"github.com/apache/incubator-pegasus/collector/webui"
+	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"gopkg.in/natefinch/lumberjack.v2"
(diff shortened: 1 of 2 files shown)


collector/main.go
```diff
@@ -98,4 +99,6 @@ func main() {
 		return metrics.NewReplicaServerMetricCollector().Start(tom)
 	})
 	<-tom.Dead() // gracefully wait until all goroutines dead
+
+	webui.StartWebServer(registry)
```
empiredan, line 103: Should this be put before `tom`, otherwise this would never be called ?

collector/main.go
```diff
@@ -81,8 +84,6 @@ func main() {
 		return
 	}
 
-	webui.StartWebServer()
-
 	tom := &tomb.Tomb{}
```
empiredan, line 87:
```suggestion
        registry := prometheus.NewRegistry()
        webui.StartWebServer(registry)
        
	tom := &tomb.Tomb{}
```

empiredan (approved):

acelyc111 (approved):

Issue #1863, opened 2024-01-19
Meta_server unable to connect kerberos zookeeper where KDC configuration "rdns = false"


GehaFearless:
## Feature Request

/etc/krb5.conf
![image](https://github.com/apache/incubator-pegasus/assets/48315319/cef98f5c-7691-475e-87ad-9d767e92de06)

I need to deploy Pegasus on machine configured with KDC as above.
At the same time, ZooKeeper enables SASL restrictions on some znodes.

I got error as follow.

zookeeper client log
`2024-01-19 21:37:26,792:13938(0x7fe9d9852700):ZOO_INFO@check_events@2988: initiated connection to server x.x.x.x:2181
2024-01-19 21:37:26,795:13938(0x7fe9d9852700):ZOO_ERROR@zoo_sasl_client_start@287: Starting SASL negotiation: generic failure SASL(-1): generic failure: GSSAPI Error: Unspecified GSS failure.  Minor code may provide more information (Server zookeeper/x.x.x.x@REALMS.COM not found in Kerberos database)
2024-01-19 21:37:26,795:13938(0x7fe9d9852700):ZOO_ERROR@_zsasl_fail@148: SASL authentication failed. rc=-1`

kdc log
`Jan 19 21:37:26 cdh01.newcluster.com krb5kdc[18354](info): TGS_REQ (8 etypes {18 17 20 19 16 23 25 26}) x.x.x.x: LOOKING_UP_SERVER: authtime 0,  pegasus/FQDN1@REALMS.COM  for zookeeper/x.x.x.x@REALMS.COM, Server not found in Kerberos database
`

FQDN `FQDN1`correspond to IP `x.x.x.x`

zookeeper service principal are `zookeeper/x.x.x.x@REALMS.COM`

empiredan:
Have we tried to set `host` in `zoo_sasl_params_t` with `NULL` ? 

```
typedef struct zoo_sasl_params {
  const char *service;          /*!< The service name, usually "zookeeper" */
  const char *host;             /*!< The server name, e.g. "zk-sasl-md5" */
  const char *mechlist;         /*!< Mechanisms to try, e.g. "DIGEST-MD5" */
  const sasl_callback_t *callbacks;  /*!< List of callbacks */
} zoo_sasl_params_t;

ZOOAPI zhandle_t *zookeeper_init_sasl(const char *host, watcher_fn fn,
  int recv_timeout, const clientid_t *clientid, void *context, int flags,
[...]

PR #1864, opened 2024-01-19
feat: Support meta_server connect to Kerberos sasl authenticated Zookeeper where KDC configured 'rdns = false'

GehaFearless:
issue: https://github.com/apache/incubator-pegasus/issues/1863

Resolve the issue of Pegasus connection failure to zookeeper due to KDC service prohibiting DNS translation of IP to FQDN

### DIFF
--- a/src/zookeeper/zookeeper_session.cpp
+++ b/src/zookeeper/zookeeper_session.cpp
@@ -30,6 +30,7 @@
 #include <utility>
 
 #include "runtime/app_model.h"
+#include "runtime/rpc/rpc_address.h"
 #include "utils/flags.h"
 #include "utils/fmt_logging.h"
 #include "zookeeper/proto.h"
@@ -43,6 +44,10 @@ DSN_DEFINE_string(security,
                   zookeeper_kerberos_service_name,
                   "zookeeper",
                   "zookeeper kerberos service name");
+DSN_DEFINE_string(security,
+                  zookeeper_sasl_service_fqdn,
+                  "",
+                  "The FQDN of a Zookeeper server, used in Kerberos Principal");
 } // namespace security
 } // namespace dsn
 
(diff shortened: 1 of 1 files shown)


src/zookeeper/zookeeper_session.cpp
```diff
@@ -161,6 +166,11 @@ int zookeeper_session::attach(void *callback_owner, const state_callback &cb)
             zoo_sasl_params_t sasl_params = {0};
             sasl_params.service = dsn::security::FLAGS_zookeeper_kerberos_service_name;
             sasl_params.mechlist = "GSSAPI";
+            rpc_address addr;
+            CHECK(addr.from_string_ipv4(dsn::security::FLAGS_zookeeper_sasl_service_fqdn),
```
empiredan, line 170: Use `GetAddrInfo` to check if fqdn is valid ?

src/zookeeper/zookeeper_session.cpp
```diff
@@ -43,6 +44,10 @@ DSN_DEFINE_string(security,
                   zookeeper_kerberos_service_name,
                   "zookeeper",
                   "zookeeper kerberos service name");
+DSN_DEFINE_string(security,
+                  zookeeper_sasl_service_fqdn,
+                  "",
+                  "The fqdn of sasl server name which one connect to zookeeper service");
```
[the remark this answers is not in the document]
GehaFearless: `The FQDN of a Zookeeper server, used in Kerberos Principal.
`
is better. Not the first host in [zookeeper] hosts_lis

src/zookeeper/zookeeper_session.cpp
GehaFearless: It's better to improve `rpc_address::from_string_ipv4` which use `GetAddrInfo` replace to `gethostbyname`. I will push one new pr later.

src/zookeeper/zookeeper_session.cpp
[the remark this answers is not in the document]
GehaFearless: image

src/zookeeper/zookeeper_session.cpp
GehaFearless: image

empiredan: Good idea to declare this function as static ! However, since it would return int, how would we decide if it is successful ?

acelyc111 (approved):

acelyc111:
Remember to cherry pick it to branch v2.5 after it been merged.

empiredan (approved):

PR #1870, opened 2024-01-24
feat(new_metrics): build the http query string based on metric filters

empiredan:
Pegasus shell needs to fetch metrics by http client. However, building query
string is not simple. You have to write each field name yourself, and assemble
all field names/values into a valid query string.

To simplify this operation, we could allow metric filters to support exporting
the query string. Users could just set conditions they need to metric filters,
then get the query string just by a function conveniently.

### DIFF
--- a/src/http/http_server.cpp
+++ b/src/http/http_server.cpp
@@ -83,11 +83,6 @@ std::string build_rel_path(const std::string &root_path, const std::string &sub_
     http_call_registry::instance().remove(full_path);
 }
 
-/* static */ std::string get_full_path(const std::string &root_path, const std::string &sub_path)
-{
-    return '/' + build_rel_path(root_path, sub_path);
-}
-
 std::string http_service::get_rel_path(const std::string &sub_path) const
 {
     return build_rel_path(path(), sub_path);
--- a/src/http/http_server.cpp
+++ b/src/http/http_server.cpp
@@ -57,6 +57,16 @@ error_s update_config(const http_request &req)
     return update_flag(iter->first, iter->second);
 }
 
+// If sub_path is 'app/duplication', the built path would be '<root_path>/app/duplication'.
+std::string build_rel_path(const std::string &root_path, const std::string &sub_path)
+{
+    std::string rel_path(root_path);
+    if (!rel_path.empty()) {
+        rel_path += '/';
+    }
+    return rel_path += sub_path;
+}
+
 } // anonymous namespace
 
 /*extern*/ http_call &register_http_call(std::string full_path)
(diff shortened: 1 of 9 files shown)


src/http/http_server.cpp
```diff
@@ -57,6 +57,16 @@ error_s update_config(const http_request &req)
     return update_flag(iter->first, iter->second);
 }
 
+// If sub_path is 'app/duplication', the built path would be '<root_path>/app/duplication'.
+std::string build_rel_path(const std::string &root_path, const std::string &sub_path)
+{
+    std::string rel_path(root_path);
+    if (!rel_path.empty()) {
```
acelyc111, line 64: In which case the root_path is empty?

src/utils/metrics.cpp
```diff
@@ -245,14 +272,19 @@ metric_entity_prototype::metric_entity_prototype(const char *name) : _name(name)
 
 metric_entity_prototype::~metric_entity_prototype() {}
 
+const std::string metrics_http_service::kMetricsRootPath("");
```
acelyc111, line 275: Where will it be used?

src/utils/strings.cpp
```diff
@@ -424,5 +425,11 @@ std::string find_string_prefix(const std::string &input, char separator)
     }
     return input.substr(0, current);
 }
+
+bool has_space(const std::string &str)
+{
+    return std::any_of(str.begin(), str.end(), [](unsigned char c) { return std::isspace(c); });
```
acelyc111, line 431: Have check the absl implementation is better than Cpp standard or not?

```
// Determines whether the given character is a whitespace character (space,
// tab, vertical tab, formfeed, linefeed, or carriage return).
inline bool ascii_isspace(unsigned char c) {
  return (ascii_internal::kPropertyBits[c] & 0x08) != 0;
}
```

src/http/http_server.cpp
empiredan: Root path of the metrics http service  is empty.

src/utils/metrics.cpp
empiredan: Used to return the root path of http service:
```c++
std::string path() const override { return kMetricsRootPath; }
```

src/utils/strings.cpp
empiredan: OK, once this function could be used, no need to use lambda expression to wrap `std::isspace`. I'll try it.

acelyc111 (approved):

acelyc111 (approved):

GehaFearless (approved):

PR #1873, opened 2024-01-25
feat(dns_resolver): Add metrics for dns_resolver

acelyc111:
The DNS resolve function maybe slow, and the memory size of dns_resolver is
not limited currently, this patch adds 3 useful metrics to improve observability:
- dns_resolver_cache_size
- dns_resolver_resolve_duration_ns
- dns_resolver_resolve_by_dns_duration_ns

### DIFF
--- a/src/runtime/rpc/dns_resolver.cpp
+++ b/src/runtime/rpc/dns_resolver.cpp
@@ -19,12 +19,16 @@
 
 #include <algorithm>
 #include <memory>
+#include <set>
 #include <utility>
 
+#include "absl/strings/string_view.h"
+#include "fmt/core.h"
 #include "fmt/format.h"
 #include "runtime/rpc/dns_resolver.h"
 #include "runtime/rpc/group_address.h"
 #include "runtime/rpc/group_host_port.h"
+#include "utils/autoref_ptr.h"
 #include "utils/fmt_logging.h"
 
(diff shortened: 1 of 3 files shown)


GehaFearless (dismissed):

GehaFearless (approved):

src/runtime/rpc/dns_resolver.cpp
```diff
@@ -19,27 +19,52 @@
...
-void dns_resolver::add_item(const host_port &hp, const rpc_address &addr)
+dns_resolver::dns_resolver()
+    : METRIC_VAR_INIT_server(dns_resolver_cache_size),
+      METRIC_VAR_INIT_server(dns_resolver_resolve_duration_ns),
+      METRIC_VAR_INIT_server(dns_resolver_resolve_by_dns_duration_ns)
 {
-    utils::auto_write_lock l(_lock);
-    _dsn_cache.insert(std::make_pair(hp, addr));
+#ifndef MOCK_TEST
+    static int only_one_instance = 0;
```
empiredan, line 57: Is it necessary to use singleton ?

acelyc111: I don't think singleton is good design, it's lifetime and ownership is confused, I'd prefer to use common object rather than singleton unless it's imposiible.

empiredan (approved):

PR #1882, opened 2024-01-30
feat(new_metrics): show sst statistics by shell app_disk based on new metrics

empiredan:


src/shell/commands/table_management.cpp
```diff
@@ -173,6 +173,63 @@ bool query_app(command_executor *e, shell_context *sc, arguments args)
...
+{
+    dsn::error_s err;
+
+    rapidjson::Document doc;
+    rapidjson::ParseResult result = doc.Parse(json_string.c_str());
+    if (dsn_unlikely(result.IsError())) {
+        return dsn::error_s::make(dsn::ERR_INVALID_DATA, "invalid json string");
+    }
+
+    const auto &entities = doc.FindMember(dsn::kMetricEntitiesField.c_str())->value;
```
acelyc111, line 200: Is it safe to access the `value` after `FindMember()`, is it possible the member could not be found?

Other places are the same.

Theer are some macros to simplify the code of jugdments in ranger_resource_policy_manager.cpp, you can take a look.

src/shell/commands/table_management.cpp
```diff
@@ -262,46 +319,32 @@ bool app_disk(command_executor *e, shell_context *sc, arguments args)
...
-         fmt::format("replica*app.pegasus*disk.storage.sst.count@{}.", app_id)});
+    const auto &results = get_metrics(nodes, sst_stat_filters(app_id).to_query_string());
 
     std::map<dsn::rpc_address, std::map<int32_t, double>> disk_map;
     std::map<dsn::rpc_address, std::map<int32_t, double>> count_map;
     for (int i = 0; i < nodes.size(); ++i) {
-        if (!results[i].first) {
-            std::cout << "ERROR: query perf counter from node " << nodes[i].address.to_string()
-                      << " failed" << std::endl;
+        if (!results[i].error().is_ok()) {
```
acelyc111, line 327:
```suggestion
        if (!results[i].error()) {
```

src/shell/commands/table_management.cpp
empiredan: Yeah, it's safe to access by `FindMember()`, however this is not graceful; and while response structure is changed, the code has to be updated.

I've turned to define a special struct for this PR, and would add other structs if necessary for later PRs.

acelyc111 (approved):

GehaFearless (approved):

PR #1892, opened 2024-02-04
feat(fqdn): Limit the cache size on dns_resolve

GehaFearless:


acelyc111:
Good job!
Consider to use a LRU cache, thers is an implementation exported from RocksDB, see https://github.com/facebook/rocksdb/blob/main/include/rocksdb/cache.h#L270

PR #1894, opened 2024-02-05
refactor(host_port): Make some functions as static

acelyc111:
This patch mainly has cchanges:
- Change the constructor `host_port(rpc_address addr)` to a static function
  `host_port from_address(rpc_address addr)`. It's a good practice to keep
  constructors light and fast, this function contains a reverse DNS lookup,
  it may be slow.

- Change `bool from_string(const std::string &s)` to a static function
  `host_port from_string(const std::string &host_port_str)` in host_port.
  It seems there isn't a scenario to modify an exist host_port object by
  calling obj.from_string(s), so make it as static to clarify this.

This patch also adds SCOPED_LOG_SLOW_EXECUTION to log slow operations if it
exceed 100 milliseconds.

### DIFF
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -437,7 +437,7 @@ inline bool json_decode(const dsn::json::JsonObject &in, dsn::host_port &hp)
         return true;
     }
     hp = host_port::from_string(host_port_string);
-    return hp.is_invalid();
+    return !hp.is_invalid();
 }
 
 inline void json_encode(JsonWriter &out, const dsn::partition_configuration &config);
--- a/src/common/json_helper.h
+++ b/src/common/json_helper.h
@@ -436,7 +436,8 @@ inline bool json_decode(const dsn::json::JsonObject &in, dsn::host_port &hp)
     if (host_port_string == "invalid host_port") {
         return true;
     }
-    return hp.from_string(host_port_string);
+    hp = host_port::from_string(host_port_string);
+    return hp.is_invalid();
 }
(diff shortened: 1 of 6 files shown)


src/runtime/rpc/rpc_host_port.cpp
```diff
@@ -40,46 +42,55 @@ const host_port host_port::s_invalid_host_port;
...
 
-    if (dsn_unlikely(rpc_address::ipv4_from_host(_host.c_str()) == 0)) {
-        return false;
+    if (dsn_unlikely(rpc_address::ipv4_from_host(hp._host.c_str()) == 0)) {
+        return hp;
     }
 
-    _type = HOST_TYPE_IPV4;
-    return true;
+    // Now is_invalid() return true.
```
empiredan, line 91:
```suggestion
    // Now is_invalid() return false.
```

src/common/json_helper.h
```diff
@@ -436,7 +436,8 @@ inline bool json_decode(const dsn::json::JsonObject &in, dsn::host_port &hp)
     if (host_port_string == "invalid host_port") {
         return true;
     }
-    return hp.from_string(host_port_string);
+    hp = host_port::from_string(host_port_string);
+    return hp.is_invalid();
```
empiredan, line 440:
```suggestion
    return !hp.is_invalid();
```

GehaFearless (approved):

empiredan (approved):

Issue #1907, opened 2024-02-18
Feature: support force no idempotent wirte when doing duplication


ninsmiracle:
## Feature Request

  In our online situations , there are some Pegasus user need to copy some mutation from master cluster to backup cluster via Pegasus duplication function . However , user have used some No-idempotent write to master cluster. In current Pegasus version , master cluster will block those No-idempotent mutations.
 So we need a force dup option for those mutations.

**Describe the feature you'd like:**
An force dup option , which can config by admin-cli. And then , master cluster can receive and dup those No-idempotent  to backup cluster via duplication.

PR #1908, opened 2024-02-18
Support no idempotent dupFeature: support force no idempotent wirte when doing duplication 

ninsmiracle:
#1907

1. Change the judgment logic in the writing process, and configure it to allow non-idempotent operations in the master cluster.
2. Modify the logic of the specified mutation committer in prepare_list
3. Add configuration items and add them to the replica side to modify them through the server-config of admin-cli.
4. Add the parsing logic after the backup cluster receives the packaged RPC of duplication.
5. Add the logic for the master cluster to obtain relevant non-idempotent RPC from the plog
6. Add the metric index of dup non-idempotent write QPS

##### Tests 
- Unit test

when user want to force dup non-idempotent mutations. Type following command in admin-cli:
```
server-config replica force_send_no_idempotent_when_duplication set true
```

### DIFF
--- a/src/server/pegasus_mutation_duplicator.cpp
+++ b/src/server/pegasus_mutation_duplicator.cpp
@@ -101,6 +101,22 @@ using namespace dsn::literals::chrono_literals;
         dsn::from_blob_to_thrift(data, thrift_request);
         return pegasus_hash_key_hash(thrift_request.hash_key);
     }
+    if (tc == dsn::apps::RPC_RRDB_RRDB_INCR) {
+        dsn::apps::incr_request thrift_request;
+        dsn::from_blob_to_thrift(data, thrift_request);
+        return pegasus_hash_key_hash(thrift_request.key);
+    }
+    if (tc == dsn::apps::RPC_RRDB_RRDB_CHECK_AND_SET) {
+        dsn::apps::check_and_set_request thrift_request;
+        dsn::from_blob_to_thrift(data, thrift_request);
+        return pegasus_hash_key_hash(thrift_request.hash_key);
+    }
+    if (tc == dsn::apps::RPC_RRDB_RRDB_CHECK_AND_MUTATE) {
+        dsn::apps::check_and_mutate_request thrift_request;
+        dsn::from_blob_to_thrift(data, thrift_request);
+        return pegasus_hash_key_hash(thrift_request.hash_key);
+    }
+
     LOG_FATAL("unexpected task code: {}", tc);
     __builtin_unreachable();
 }
--- a/src/server/pegasus_write_service.cpp
+++ b/src/server/pegasus_write_service.cpp
@@ -456,7 +456,6 @@ int pegasus_write_service::duplicate(int64_t decree,
             }
         }
 
-
         resp.__set_error(rocksdb::Status::kInvalidArgument);
         resp.__set_error_hint(fmt::format("unrecognized task code {}", request.task_code));
         return empty_put(ctx.decree);
--- a/src/server/pegasus_write_service.cpp
+++ b/src/server/pegasus_write_service.cpp
@@ -126,6 +126,11 @@ METRIC_DEFINE_counter(replica,
                       dsn::metric_unit::kRequests,
                       "The number of DUPLICATE requests");
 
+METRIC_DEFINE_counter(replica,
+                      no_idempotent_duplicate,
+                      dsn::metric_unit::kRequests,
+                      "The number of forced idempotent requests when doing duplication");
+
 METRIC_DEFINE_percentile_int64(replica,
                                dup_time_lag_ms,
                                dsn::metric_unit::kMilliSeconds,
@@ -169,6 +174,7 @@ pegasus_write_service::pegasus_write_service(pegasus_server_impl *server)
       METRIC_VAR_INIT_replica(check_and_set_latency_ns),
       METRIC_VAR_INIT_replica(check_and_mutate_latency_ns),
       METRIC_VAR_INIT_replica(dup_requests),
+      METRIC_VAR_INIT_replica(no_idempotent_duplicate),
       METRIC_VAR_INIT_replica(dup_time_lag_ms),
       METRIC_VAR_INIT_replica(dup_lagging_writes),
       _put_batch_size(0),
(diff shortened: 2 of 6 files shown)


ninsmiracle:
This PR's base function is in line with expectations , however I was requested by business department to add more matrix for surveillance system about this feature. I will update this PR in recently days.

ninsmiracle:
Now , when user config `force_send_no_idempotent_when_duplication` to true . We can dup non-idempotent write to backup cluster , and we can verify the data consistency by check the number rows of pegasus app.
app in master-cluster
image

app in backup-cluster
image


And because the retries for non-idempotent writes will make influence of the  consistency between two clusters. So we should show the metric like following picture:
How many non-idempotent write have been retried when doing duplication

show in backup cluster:
How many non-idempotent write have been received by backup cluster

src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -205,6 +230,75 @@ void pegasus_mutation_duplicator::on_duplicate_reply(uint64_t hash,
...
+void pegasus_mutation_duplicator::type_force_send_no_idempotent_if_need(duplicate_rpc &rpc)
+{
+
+    // there maybe more than one mutation in one dup rpc
+    if (FLAGS_force_send_no_idempotent_when_duplication) {
+        for (auto entry : rpc.request().entries) {
+            if (entry.task_code == dsn::apps::RPC_RRDB_RRDB_INCR ||
+                entry.task_code == dsn::apps::RPC_RRDB_RRDB_CHECK_AND_SET ||
+                entry.task_code == dsn::apps::RPC_RRDB_RRDB_CHECK_AND_MUTATE) {
+
```
acelyc111, line 242: Add these codes in a set, then:

```
if (entry.task_code not in set) {
    continue;
}

METRIC_VAR_INCREMENT(dup_retry_no_idempotent_duplicate_qps);

switch (entry.task_code) {
case x:
...
case y:
...
}
```

src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -205,6 +230,75 @@ void pegasus_mutation_duplicator::on_duplicate_reply(uint64_t hash,
...
+                METRIC_VAR_INCREMENT(dup_retry_no_idempotent_duplicate_qps);
+
+                dsn::message_ex *write =
+                    dsn::from_blob_to_received_msg(entry.task_code, entry.raw_message);
+
+                if (entry.task_code == dsn::apps::RPC_RRDB_RRDB_INCR) {
+                    incr_rpc raw_rpc(write);
+                    absl::string_view unmarshall_key(raw_rpc.request().key.data(),
```
acelyc111, line 250: Is it possible to use `raw_rpc.request().key` directly in LOG_DEBUG?

src/server/pegasus_write_service.cpp
```diff
@@ -415,6 +421,40 @@ int pegasus_write_service::duplicate(int64_t decree,
...
+
+        if (request.task_code == dsn::apps::RPC_RRDB_RRDB_INCR ||
+            request.task_code == dsn::apps::RPC_RRDB_RRDB_CHECK_AND_SET ||
+            request.task_code == dsn::apps::RPC_RRDB_RRDB_CHECK_AND_MUTATE) {
+            // receive no idempotent request from master cluster via duplication
+            METRIC_VAR_INCREMENT(force_receive_no_idempotent_duplicate_qps);
+
+            if (request.task_code == dsn::apps::RPC_RRDB_RRDB_INCR) {
+                incr_rpc rpc(write);
+                resp.__set_error(_impl->incr(ctx.decree, rpc.request(), rpc.response()));
```
acelyc111, line 433: Add some comments to describe what's you aim here.

src/common/duplication_common.h
```diff
@@ -31,6 +31,7 @@
 #include "utils/fmt_utils.h"
 
 DSN_DECLARE_uint32(duplicate_log_batch_bytes);
+DSN_DECLARE_bool(force_send_no_idempotent_when_duplication);
```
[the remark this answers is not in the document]
ninsmiracle: I think it's necessary . Cause once I remove this line , compilation can not be passed.

src/server/pegasus_mutation_duplicator.cpp
ninsmiracle: `entry.task_code` is not a const variable. And I think using task_code.to_string is not more efficient than `if` structure.

src/server/pegasus_mutation_duplicator.cpp
ninsmiracle: In my opinion , `raw_rpc.request().key` is a blob. And `unmarshall_key` is more friendly for Pegasus operation and maintenance person.

src/common/duplication_common.h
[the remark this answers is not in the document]
acelyc111: The DSN_DECLARE_xxx macros are recommended to move to cpp files where actually use it, don't make the header files to be too large.

The same to `DSN_DECLARE_uint32(duplicate_log_batch_bytes); `

src/common/duplication_common.cpp
```diff
@@ -37,6 +37,13 @@ DSN_DEFINE_uint32(replication,
                   "send mutation log batch bytes size per rpc");
 DSN_TAG_VARIABLE(duplicate_log_batch_bytes, FT_MUTABLE);
 
+DSN_DEFINE_bool("replication",
+                force_send_no_idempotent_when_duplication,
+                false,
+                "receive client idempotent write requests and send them to backup cluster when "
```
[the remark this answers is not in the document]
acelyc111: Didn't update this?

src/server/pegasus_write_service.cpp
```diff
@@ -121,6 +121,11 @@ METRIC_DEFINE_counter(replica,
                       dsn::metric_unit::kRequests,
                       "The number of DUPLICATE requests");
 
+METRIC_DEFINE_counter(replica,
+                      force_receive_no_idempotent_duplicate_qps,
```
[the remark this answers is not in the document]
acelyc111: Didn't update this?

Similar to `dup_requests` above.

src/common/duplication_common.cpp
```diff
@@ -37,6 +37,13 @@ DSN_DEFINE_uint32(replication,
                   "send mutation log batch bytes size per rpc");
 DSN_TAG_VARIABLE(duplicate_log_batch_bytes, FT_MUTABLE);
 
+DSN_DEFINE_bool(replication,
+                force_send_non_idempotent_when_duplication,
```
acelyc111, line 41: To keep the coding style and convenient distinguish which component the config belongs to, it would be better to rename config names in duplication as `duplication_xxx`, metric names as `dup_xxx`.

For example, force_send_non_idempotent_when_duplication -> duplication_unsafe_allow_non_idempotent, the "unsafe" stands by issue may caused the retry mechanism.

src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -59,6 +60,11 @@ METRIC_DEFINE_counter(replica,
                       dsn::metric_unit::kRequests,
                       "The number of failed DUPLICATE requests sent from client");
 
+METRIC_DEFINE_counter(replica,
+                      dup_retry_non_idempotent_duplicate_request,
+                      dsn::metric_unit::kRequests,
+                      "The qps of Non-idempotent write when doing DUPLICATE which is Retried");
```
acelyc111, line 66: Maybe ?

```suggestion
                      "The number of retried non-idempotent DUPLICATE requests sent from client");
```

src/server/pegasus_mutation_duplicator.cpp
acelyc111: Is there any difference? the operator<</to_string()/to_string_view() member function of blob are using data() and length() to create a string/string_view object too.

src/common/duplication_common.cpp
ninsmiracle: OK

src/common/duplication_common.cpp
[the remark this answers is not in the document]
ninsmiracle: sorry. I corrected it

src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -189,6 +211,9 @@ void pegasus_mutation_duplicator::on_duplicate_reply(uint64_t hash,
             // retry this rpc
             _inflights[hash].push_front(rpc);
             _env.schedule([hash, cb, this]() { send(hash, cb); }, 1_s);
```
acelyc111, line 213: Will the non-idempotent request be retried twice here?

ninsmiracle: Yes,it will. So I record specific `raw_key` into log.

ruojieranyishen:
new pr is https://github.com/apache/incubator-pegasus/pull/2380

Issue #1880, opened 2024-01-30
Bug(go client):The cluster added a new meta node, but the meta server configuration of the go client was not updated. As a result, the client cannot find the new meta address and can only access the meta listed in the meta list


lengyuexuexuan:
Assuming the Pegasus client is configured with a meta server list of "127.0.0.1:34602" and "127.0.0.1:34603," but the actual primary meta server for the Pegasus server is "127.0.0.1:34601," the Pegasus client will not be able to connect to the Pegasus server until a timeout occurs.

   The reason is that when the go client searches for the primary, it iterates through the meta server list, sending an RPC RPC_CM_QUERY_PARTITION_CONFIG_BY_INDEX to each meta server and making a determination based on the response.

   Unlike the Java client, the go client cannot directly use indirection to add meta servers not specified in the configuration to the client.
   Below is the logic code for this part of the go client.
```
// go-client/session/meta_call.go  
func (c *metaCall) issueBackupMetas(ctx context.Context) {
    for i := range c.metas {
       if i == c.lead {
          continue
       }
       // concurrently issue RPC to the rest of meta servers.
       go func(idx int) {
          c.issueSingleMeta(ctx, idx)
       }(i)
    }
}

// issueSingleMeta returns false if we should try another meta
func (c *metaCall) issueSingleMeta(ctx context.Context, i int) bool {
    meta := c.metas[i]
    resp, err := c.callFunc(ctx, meta)
    if err != nil || resp.GetErr().Errno == base.ERR_FORWARD_TO_OTHERS.String() {
       return false
    }
    // the RPC succeeds, this meta becomes the new leader now.
    atomic.StoreUint32(&c.newLead, uint32(i))
    select {
    case <-ctx.Done():
    case c.respCh <- resp:
       // notify the caller
    }
    return true
}
```
Here is the relevant part of the Java client code for this:
```
// com/xiaomi/infra/pegasus/rpc/async/MetaSession.java  onFinishQueryMeta()

synchronized (this) {
  if (needSwitchLeader) {
    if (forwardAddress != null && !forwardAddress.isInvalid()) {
      boolean found = false;
      for (int i = 0; i < metaList.size(); i++) {
        if (metaList.get(i).getAddress().equals(forwardAddress)) {
          curLeader = i;
          found = true;
          break;
        }
      }
      if (!found) {
        logger.info("add forward address {} as meta server", forwardAddress);
        metaList.add(clusterManager.getReplicaSession(forwardAddress));
[...]

Issue #1856, opened 2024-01-16
BUG(go client):when go client is writing to one partition and the replica node core dump, go client will finish after timeout without updating the configuration.


lengyuexuexuan:
In the code, when replcia core dump, the function loopForResponse() will return "nil".
![1](https://github.com/apache/incubator-pegasus/assets/46274877/0f140e05-5572-4c55-9425-fcc909228aac)
Then, the process will be blocked in function CallWithGpid() until the time exceeds the timeout.
![2](https://github.com/apache/incubator-pegasus/assets/46274877/e34dcff9-538c-4d94-9c12-f72a2997442f)
![3](https://github.com/apache/incubator-pegasus/assets/46274877/d2c90da3-c116-4772-b8e1-b8f7bd9c807f)

why not update the configuration of table and retry previous operation when the above situation occurs?

lengyuexuexuan:
> @lengyuexuexuan Thanks for the feedback, could you please submit a patch to fix it?

OK. No problem.

PR #1909, opened 2024-02-19
fix(go-client): update config once replica server failed and forward to primary meta server if it was changed

lengyuexuexuan:
#1880
#1856

As for #1856. 
when go client is writing to one partition and the replica node core dump, go client will finish after timeout without updating the configuration. In this case, the go client only restart to solve the problem. 

In this pr, the client would update conconfiguration of table automatically when someone replica core dump.
After testing, we found that the the replicaerror is "context.DeadlineExceeded" when the replica core dump.

https://github.com/apache/incubator-pegasus/blob/41141c11c36930a19da727fd25a4876bd56f76a6/go-client/pegasus/table_connector.go#L705-L706

Therefore, when client meets the errror, the go client will update configuration automatically.
Besides, this request will not retry. Because only in the case of timeout, the configuration will be automatically updated. If you try again before then, it will still fail. There is also the risk of infinite retries. 
Therefore, it is better to directly return the request error to the user and let the user try again.

As for #1880
When the client sends an RPC message "RPC_CM_QUERY_PARTITION_CONFIG_BY_INDEX" to the meta server, if the meta server isn't primary, the response that forward to the primary meta server will return. 

According to the above description, assuming that the client does not have a primary meta server configured, we can connect to the primary meta server in this way.

In this PR, we implement this function through the following steps.

1. First parse the response, determine whether its errno is ERR_FORWARD_TO_OTHERS, and then parse it to get the primary meta server address.  
https://github.com/apache/incubator-pegasus/blob/41141c11c36930a19da727fd25a4876bd56f76a6/go-client/session/meta_call.go#L166-L177
2. Secondly, determine whether the address is already in the client configuration. If it is already there, skip it directly. Otherwise, establish a connection and pull the configuration directly from the primary meta server.
https://github.com/apache/incubator-pegasus/blob/41141c11c36930a19da727fd25a4876bd56f76a6/go-client/session/meta_call.go#L118-L138

It should be noted that the IP address and session do not have a one-to-one correspondence, because there may be situations where the IP address is unavailable. 
This is why there is a priamry meta server configuration in the client, but the curllead cannot be used as the index of the metaIPAddrs array.
https://github.com/apache/incubator-pegasus/blob/41141c11c36930a19da727fd25a4876bd56f76a6/go-client/session/meta_call.go#L123-L128

##### Tests 

- Unit test
1. Start onebox, and the primary meta server is not added to the go client configuration.
2. The go client writes data to a certain partition and then kills the replica process.

### DIFF
--- a/src/runtime/rpc/dns_resolver.h
+++ b/src/runtime/rpc/dns_resolver.h
@@ -26,6 +26,7 @@
 #include "runtime/rpc/rpc_host_port.h"
 #include "utils/errors.h"
 #include "utils/metrics.h"
+#include "utils/singleton.h"
 #include "utils/synchronize.h"
 
 namespace dsn {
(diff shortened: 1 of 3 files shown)


acelyc111:
@lengyuexuexuan It would be better to separate the bugfix and the refactor of generation thrift files to 2 pull requests.

lengyuexuexuan:
> @lengyuexuexuan It would be better to separate the bugfix and the refactor of generation thrift files to 2 pull requests.

ok.

PR #1929, opened 2024-03-06
refactor(go-client): refactor admin APIs for meta by reflection

empiredan:


go-client/admin/client.go
```diff
@@ -33,51 +34,101 @@ import (
...
-	CreateTable(ctx context.Context, tableName string, partitionCount int, successIfExist_optional ...bool) error
+	Close() error
 
-	DropTable(ctx context.Context, tableName string) error
+	GetTimeout() time.Duration
+	SetTimeout(timeout time.Duration)
 
-	ListTables(ctx context.Context) ([]*TableInfo, error)
-}
+	CreateTable(tableName string, partitionCount int32, replicaCount int32, envs map[string]string, maxWaitSeconds int32, successIfExistOptional ...bool) (int32, error)
```
acelyc111, line 42: It would be better to describe what does `maxWaitSeconds`, I doubt if there is any difference with `timeout`.

empiredan: It's necessary to add comments to explain this parameter. I'll add some comments.

acelyc111 (approved):

GehaFearless (approved):

PR #1930, opened 2024-03-06
feat(local_partition_split): Add local_partition_split CLI tool

acelyc111:
This patch introduces a new CLI tool `local_partition_split`, which is used
to split the local partitions offline.

Similar to the online partition split feature, it's helpful to split the
table which has large amount of data but with a few partitions into more
partitions to improve the throughput and lower latency.

Besides, this tool is possible to be used on older Pegasus versions, if you
want to do that, it's necessary to cherr-pick this patch to the corresponding
older branch and rebuild the tool, then this tool will use corresponding
version of RocksDB library to generate new replicas.

### DIFF
--- a/src/shell/commands/local_partition_split.cpp
+++ b/src/shell/commands/local_partition_split.cpp
@@ -15,24 +15,52 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include "shell/commands.h"
-
-#include <cmath>
+#include <ext/alloc_traits.h>
+#include <fmt/core.h>
+#include <fmt/format.h>
+#include <rocksdb/db.h>
+#include <rocksdb/env.h>
+#include <rocksdb/iterator.h>
+#include <rocksdb/metadata.h>
+#include <rocksdb/options.h>
+#include <rocksdb/slice.h>
 #include <rocksdb/sst_file_reader.h>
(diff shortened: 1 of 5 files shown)


empiredan (approved):

src/shell/commands/local_partition_split.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+    // 2. Check parameters.
+    if (!validate_parameters(lpsc)) {
+        return false;
+    }
+
+    // 3. Split each data directory.
+    auto data_dirs_thread_pool = std::unique_ptr<rocksdb::ThreadPool>(
+        rocksdb::NewThreadPool(static_cast<int>(lpsc.src_data_dirs.size())));
+    CHECK_EQ(lpsc.src_data_dirs.size(), lpsc.dst_data_dirs.size());
```
GehaFearless, line 713: `validate_parameters` already check it

acelyc111: Yes, we will use an index to iterate the 2 containers so we use CHECK_XX here to ensure the validate_parameters() works, to prevent one day the logic in validate_parameters() is removed or incorrectly modified.

GehaFearless (approved):

PR #1933, opened 2024-03-06
refactor(replica): Unify some unique variables to reduce redundances

acelyc111:
Unify some string variables:
- kRepsDir: "reps"
- kReplicaAppType: "replica"
- kDataDirPostfix: "data"
- kRdbPostfix: "rdb"

### DIFF
--- a/src/replica/bulk_load/test/replica_bulk_loader_test.cpp
+++ b/src/replica/bulk_load/test/replica_bulk_loader_test.cpp
@@ -23,6 +23,7 @@
 
 #include "common/bulk_load_common.h"
 #include "common/gpid.h"
+#include "common/replication_common.h"
 #include "dsn.layer2_types.h"
 #include "gtest/gtest.h"
 #include "replica/test/mock_utils.h"
--- a/src/replica/bulk_load/test/replica_bulk_loader_test.cpp
+++ b/src/replica/bulk_load/test/replica_bulk_loader_test.cpp
@@ -23,7 +23,6 @@
 
 #include "common/bulk_load_common.h"
 #include "common/gpid.h"
-#include "common/replication_common.h"
 #include "dsn.layer2_types.h"
 #include "gtest/gtest.h"
 #include "replica/test/mock_utils.h"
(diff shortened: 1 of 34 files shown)


src/replica/bulk_load/test/replica_bulk_loader_test.cpp
```diff
@@ -404,7 +405,6 @@ class replica_bulk_loader_test : public replica_test_base
     file_meta _file_meta;
     bulk_load_metadata _metadata;
 
-    std::string APP_NAME = "replica";
```
empiredan, line 407: I think this `APP_NAME` is the name of a table rather than the service app (meta/replica, i.e. `replication_options::kReplicaAppType`).  To clear this, we could use some name like `"replica_bulk_loader_test"` as the table name instead of `"replica"`.

src/replica/replication_app_base.cpp
```diff
@@ -119,6 +120,9 @@ error_code replica_app_info::store(const std::string &fname)
         fname, writer.get_buffer(), dsn::utils::FileDataType::kSensitive);
 }
 
+const std::string replication_app_base::kDataDirPostfix = "data";
+const std::string replication_app_base::kRdbPostfix = "rdb";
```
empiredan, line 124: Better to name them as `kDataDir` and `kRdbDir` ?

src/replica/replication_app_base.h
```diff
@@ -123,6 +126,8 @@ class replication_app_base : public replica_base
     typedef replication_app_base *factory(replica *r);
     static void register_storage_engine(const std::string &name, factory f);
     static replication_app_base *new_storage_instance(const std::string &name, replica *r);
+    static const std::string kDataDirPostfix;
+    static const std::string kRdbPostfix;
```
empiredan, line 130: Need to also replace `"data"` and `"rdb"` in other files such as `src/server/pegasus_server_impl.cpp` `src/server/test/pegasus_server_test_base.h` ?

empiredan (approved):

levy5307 (approved):

Issue #1937, opened 2024-03-08
Bug(message_ex::copy): In some case, the message_ex::copy function will trigger an assertion


Sunflower876:
## Bug Report

There are two formats for message_ex, one is that both the header and body are saved in the buffers field, and the header and buffers [0] are the same. the other format is that the header is not saved in the buffers field and exists separately. In the message_ex::copy function, total_length represents the sum of the lengths of the header and body. However, in the second case, i only counts the length of the body, so checking that i and total_length are equal will fail. 

![20240308-151800](https://github.com/apache/incubator-pegasus/assets/162662221/89ff3782-2ccb-46ac-84d2-96e74e7e220e)

Sunflower876:
@acelyc111

acelyc111:
@Sunflower876 Thanks!

Have you encountered such a check failure? Could you please leave some logs here, a backtrace would be better.

Sunflower876:
> @Sunflower876 Thanks!
> 
> Have you encountered such a check failure? Could you please leave some logs here, a backtrace would be better.

#0  0x00007fe05b4a3694 in vfprintf () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install glibc-2.17-157.el7_3.1.x86_64 keyutils-libs-1.5.8-3.el7.x86_64 libcom_err-1.42.9-9.el7.x86_64 libgcc-4.8.5-28.el7_5.1.x86_64 zlib-1.2.7-17.el7.x86_64
(gdb) #0  0x00007fe05b4a3694 in vfprintf () from /lib64/libc.so.6
#1  0x00007fe05f16ae00 in dsn::tools::simple_logger::dsn_logv (
    this=0x2b25030, file=<optimized out>, function=<optimized out>,
    line=<optimized out>, log_level=LOG_LEVEL_FATAL,
    fmt=0x7fe06058ab7d "%d VS %d, rpc_name = %s", args=<optimized out>)
    at /home/work/pegasus/rdsn/src/utils/simple_logger.cpp:206
#2  0x00007fe05f15e04c in dsn_logv (file=0x7fe0604f25b7 "compiler_depend.ts",
    function=0x7fe06058ac56 <dsn::message_ex::copy(bool, bool)::__FUNCTION__> "copy", line=248, log_level=LOG_LEVEL_FATAL,
    fmt=0x7fe06058ab7d "%d VS %d, rpc_name = %s",
    args=args@entry=0x7fe01b3ce598)
[...]

PR #1938, opened 2024-03-08
fix(message_ex): Fix the crash caused by message_ex::copy() in server internally

Sunflower876:
#1937

##### Tests 

- Unit test

### DIFF
--- a/src/runtime/rpc/rpc_message.cpp
+++ b/src/runtime/rpc/rpc_message.cpp
@@ -252,6 +252,7 @@ message_ex *message_ex::copy(bool clone_content, bool copy_for_receive)
 
         if ((const char *)header != buffers[0].data()) {
             memcpy(ptr, (const void *)header, sizeof(message_header));
+            i += sizeof(message_header);
             ptr += sizeof(message_header);
         }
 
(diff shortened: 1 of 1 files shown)


src/runtime/rpc/rpc_message.cpp
```diff
@@ -252,6 +252,7 @@ message_ex *message_ex::copy(bool clone_content, bool copy_for_receive)
 
         if ((const char *)header != buffers[0].data()) {
             memcpy(ptr, (const void *)header, sizeof(message_header));
+            i += sizeof(message_header);
```
acelyc111, line 255: How about removing the `i` variable totally, you can compare the offset of `ptr` and `recv_buffer` instead in line 264.

Sunflower876: I have made the changes, could you please help me take a look again.

acelyc111 (approved):

src/runtime/rpc/rpc_message.cpp
acelyc111: Thanks for the contribution!

empiredan (approved):

PR #1942, opened 2024-03-11
fix: fix recall_app with default app_name

GehaFearless:
fix bug from https://github.com/apache/incubator-pegasus/commit/2c9b5ed5d95ad89e2cd50ac3e34f57149788b9ce

### DIFF
--- a/src/client/replication_ddl_client.cpp
+++ b/src/client/replication_ddl_client.cpp
@@ -72,9 +72,11 @@ namespace replication {
 
 using tp_output_format = ::dsn::utils::table_printer::output_format;
 
-error_s replication_ddl_client::validate_app_name(const std::string &app_name)
+error_s replication_ddl_client::validate_app_name(const std::string &app_name,
+                                                  bool allow_empty_name)
 {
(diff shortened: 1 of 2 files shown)


src/client/replication_ddl_client.cpp
```diff
@@ -226,8 +226,11 @@ dsn::error_code replication_ddl_client::drop_app(const std::string &app_name, in
 
 dsn::error_code replication_ddl_client::recall_app(int32_t app_id, const std::string &new_app_name)
 {
-    RETURN_EC_NOT_OK_MSG(
-        validate_app_name(new_app_name), "invalid new_app_name: '{}'", new_app_name);
+    if (!std::all_of(new_app_name.cbegin(),
+                     new_app_name.cend(),
+                     (bool (*)(int))replication_ddl_client::valid_app_char)) {
```
acelyc111, line 231: valid_app_char() has been removed.

It would be better to add a new boolean parameter for validate_app_name() to indicate whether empty is allowed or not.

acelyc111 (approved):

empiredan (approved):

PR #1150, opened 2022-09-07
feat(backup): 9. replica handle backup clear context and invalid status

hycdong:
https://github.com/apache/incubator-pegasus/issues/1081

This pr is about replica handle clear after backup succeed or failed, and add invalid status check, including:
- add backup_status check in function `on_backup`
- implement function `clear_context` to clear backup context
- implement function `background_clear_backup_checkpoint` and `clear_backup_checkpoint` to delete replica checkpoint directory created by backup
- when replica configuration changed, call clear_context to stop possible backup, in `replica_config.cpp`
- add related unit tests

PR #1944, opened 2024-03-13
feat(backup): 1. remove pervious implementation base on master

ninsmiracle:
#1944

This pr removes some previous backup implementation, because the enhancement code is strongly different with the current one. Besides, this pr also disable related unit tests and function tests
##### Tests 
- Unit test


acelyc111:
@ninsmiracle Is the once-backup and retore functions still work after this patch?

ninsmiracle:
> @ninsmiracle Is the once-backup and retore functions still work after this patch?

It can't work util #1150 been merge , in other words, there are still eight PRs still need to be merged. Maybe , should I pull this request to a copy of master?

PR #1949, opened 2024-03-15
refactor(FQDN): Add some convenient macros and update idl/bulk_load.thrift related code

acelyc111:
- Add a new macros SET_IP_AND_HOST_PORT, SET_VALUE_FROM_IP_AND_HOST_PORT,
  SET_VALUE_FROM_HOST_PORT and FMT_HOST_PORT_AND_IP
- Add fill_partition_bulk_load_states() function to simplify code in
  src/meta/test/meta_bulk_load_service_test.cpp

### DIFF
--- a/idl/bulk_load.thrift
+++ b/idl/bulk_load.thrift
@@ -91,7 +91,7 @@ struct bulk_load_request
 {
     1:dsn.gpid                pid;
     2:string                  app_name;
-    3:dsn.rpc_address         primary_addr;
+    3:dsn.rpc_address         primary;
     4:string                  remote_provider_name;
     5:string                  cluster_name;
     6:i64                     ballot;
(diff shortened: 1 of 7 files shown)


idl/bulk_load.thrift
```diff
@@ -91,7 +91,7 @@ struct bulk_load_request
 {
     1:dsn.gpid                pid;
     2:string                  app_name;
-    3:dsn.rpc_address         primary_addr;
+    3:dsn.rpc_address         primary;
```
Smityz, line 94: why change this name？could you provide some context message for it？

acelyc111: In some newly added macros, they suppose the rpc_address and host_port fields are naming similarly with name `xxx` and `hp_xxx`. The macros are helpful to ensure the operations are taken both on the 2 fields. So I rename it.

Rename the thrift structure fileds dosen't break compatibility if not change the sequence number.

Smityz (approved):

foreverneverer (approved):

PR #1951, opened 2024-03-15
fix(balance): Distinguish cluster balanced or collecting balance info interruptted

Samunroyu:
After scaling out, Pegasus determines whether the shards are balanced by checking if the balance operation counts are equal to 0. However, since the meta server returns counts = 0 when it cannot collect all replicas information, this leads to the premature termination of balancing. Therefore, in such cases, the counts should be equal to -1

A new bool variable  introduce into class greedy_load_balancer to control balance operation count whether be -1.
```
class greedy_load_balancer{
private:
    bool _all_replca_infos_collected;
}
```

### DIFF
--- a/src/meta/greedy_load_balancer.cpp
+++ b/src/meta/greedy_load_balancer.cpp
@@ -63,7 +63,7 @@ greedy_load_balancer::greedy_load_balancer(meta_service *_svc) : server_load_bal
 {
     _app_balance_policy = std::make_unique<app_balance_policy>(_svc);
     _cluster_balance_policy = std::make_unique<cluster_balance_policy>(_svc);
-    _all_replca_info = false;
+    _all_replca_infos_collected = false;
 
     ::memset(t_operation_counters, 0, sizeof(t_operation_counters));
 }
@@ -164,10 +164,10 @@ void greedy_load_balancer::greedy_balancer(const bool balance_checker)
     for (auto &kv : *(t_global_view->nodes)) {
         node_state &ns = kv.second;
         if (!all_replica_infos_collected(ns)) {
-            _all_replca_info = false;
+            _all_replca_infos_collected = false;
             return;
         } else {
-            _all_replca_info = true;
+            _all_replca_infos_collected = true;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/meta/greedy_load_balancer.h
```diff
@@ -74,6 +74,7 @@ class greedy_load_balancer : public server_load_balancer
     migration_list *t_migration_result;
     int t_alive_nodes;
     int t_operation_counters[MAX_COUNT];
+    bool _all_replca_info;
```
acelyc111, line 77: How about renaming to `_all_replica_infos_collected`?

src/meta/greedy_load_balancer.cpp
```diff
@@ -235,7 +239,14 @@ void greedy_load_balancer::report(const dsn::replication::migration_list &list,
         }
     }
 
+    if (!_all_replca_info) {
+        counters[ALL_COUNT] = -1;
+    }
+
     ::memcpy(t_operation_counters, counters, sizeof(counters));
+    LOG_DEBUG(
+        "balance checker operation count = %d, due to meta server hasn't collected all replica",
```
acelyc111, line 248:
```suggestion
        "balance checker operation count = {}, due to meta server hasn't collected all replica",
```

acelyc111 (approved):

src/meta/greedy_load_balancer.cpp
```diff
@@ -163,7 +164,10 @@ void greedy_load_balancer::greedy_balancer(const bool balance_checker)
     for (auto &kv : *(t_global_view->nodes)) {
         node_state &ns = kv.second;
         if (!all_replica_infos_collected(ns)) {
+            _all_replca_infos_collected = false;
             return;
+        } else {
+            _all_replca_infos_collected = true;
         }
```
empiredan, line 171:
```suggestion
        _all_replca_infos_collected = all_replica_infos_collected(ns);
        if (!_all_replca_infos_collected) {
            return;
        }
```

empiredan (approved):

acelyc111 (approved):

PR #1960, opened 2024-03-21
feat(duplication): allow remote app name to be specified for duplication

empiredan:
Previously a different remote app name could not be specified while adding
a new duplication. However, there are some scenarios that the app should
have a different name from the source cluster. For example, when an app is
migrated from a small cluster to a large cluster, it needs to be renamed since
many apps on the target cluster where there might be some app with same
name.

Thus it is necessary to specify a custom name for the app in remote cluster.
This PR is to support this feature. To support this feature, an optional remote
app name is added for some structures of duplication RPCs such as adding a
new duplication and syncing duplications. Remote app name is also added for
meta data on consistent storage(ZK) for duplication. Compatibility is also
considered once remote app name is missing in RPCs and meta data. Pegasus
shell provides an optional parameter for specifying custom remote name that
is allowed to be different from source cluster while adding a new duplication.


src/shell/commands/duplication.cpp
```diff
@@ -145,44 +166,49 @@ bool query_dup(command_executor *e, shell_context *sc, arguments args)
...
+
+    dsn::utils::table_printer printer;
+    printer.add_title("dup_id");
+    printer.add_column("status");
+    printer.add_column("remote cluster");
+    printer.add_column("create time");
+
+    char create_time[25];
+    for (auto info : resp.entry_list) {
+        dsn::utils::time_ms_to_date_time(info.create_ts, create_time, sizeof(create_time));
```
acelyc111, line 201: How about using `void time_ms_to_string(uint64_t ts_ms, std::string &str)` version to avoid array overflow unexpectly?

src/common/duplication_common.cpp
```diff
@@ -143,13 +145,20 @@ static nlohmann::json duplication_entry_to_json(const duplication_entry &ent)
...
     if (ent.__isset.progress) {
         nlohmann::json sub_json;
         for (const auto &p : ent.progress) {
             sub_json[std::to_string(p.first)] = p.second;
         }
         json["progress"] = sub_json;
     }
+
+    if (ent.__isset.remote_app_name) {
+        // remote_app_name is shown only if it was specified explicitly for new versions.
```
acelyc111, line 158: Better to mention since which version (2.6).

idl/duplication.thrift
```diff
@@ -66,6 +66,9 @@ struct duplication_add_request
     // - if false, duplication start state=DS_LOG,
     // server will replay and send plog mutation to follower cluster derectly
     3:optional bool is_duplicating_checkpoint = true;
+
+    // Specify the app name of remote cluster.
+    4:optional string remote_app_name;
```
acelyc111, line 71: For the newly added fields, describe since which version (i.e., 2.6) they are added, as how this do in `struct app_info`.

src/meta/duplication/duplication_info.h
```diff
@@ -53,19 +53,21 @@ class duplication_info
     /// \see meta_duplication_service::new_dup_from_init
     /// \see duplication_info::decode_from_blob
     duplication_info(dupid_t dupid,
```
acelyc111, line 55: Unify to rename it to `dup_id`?

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -69,12 +70,13 @@ void meta_duplication_service::query_duplication_info(const duplication_query_re
...
-            for (auto &dup_id_to_info : app->duplications) {
-                const duplication_info_s_ptr &dup = dup_id_to_info.second;
-                dup->append_if_valid_for_query(*app, response.entry_list);
-            }
+            return;
+        }
+
+        response.appid = app->app_id;
+        for (auto &dup_id_to_info : app->duplications) {
+            const duplication_info_s_ptr &dup = dup_id_to_info.second;
```
acelyc111, line 78:
```suggestion
        for (const auto &[_, dup] : app->duplications) {
```

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -150,6 +152,19 @@ void meta_duplication_service::do_modify_duplication(std::shared_ptr<app_state>
         });
 }
 
+#define LOG_DUP_HINT_AND_RETURN(resp, level, ...)                                                  \
+    do {                                                                                           \
+        std::string _msg(fmt::format(__VA_ARGS__));                                                \
```
acelyc111, line 157:
```suggestion
        const std::string _msg(fmt::format(__VA_ARGS__));                                          \
```

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -158,24 +173,34 @@ void meta_duplication_service::add_duplication(duplication_add_rpc rpc)
...
+             request.remote_cluster_name,
+             remote_app_name);
 
     if (request.remote_cluster_name == get_current_cluster_name()) {
         response.err = ERR_INVALID_PARAMETERS;
-        response.__set_hint("illegal operation: adding duplication to itself");
-        return;
+        LOG_WARNING_DUP_HINT_AND_RETURN(response,
+                                        "illegal operation: adding duplication to itself");
     }
```
acelyc111, line 195: How about define the macro as:
LOG_WARNING_DUP_HINT_AND_RETURN_IF_NOT(exp, resp, ...)

The error code is always `ERR_INVALID_PARAMETERS` so I think it can be ommited.

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -184,50 +209,97 @@ void meta_duplication_service::add_duplication(duplication_add_rpc rpc)
...
+            // The duplication for the same app to the same remote cluster has existed.
+            remote_app_name = dup->remote_app_name;
+            LOG_INFO("no need to add duplication, since it has existed: app_name={}, "
+                     "remote_cluster_name={}, remote_app_name={}",
+                     request.app_name,
+                     request.remote_cluster_name,
+                     remote_app_name);
+        } else {
+            // Check if other apps of this cluster are duplicated to the same remote app.
+            for (const auto &kv : _state->_exist_apps) {
```
acelyc111, line 247:
```suggestion
            for (const auto &[app_name, cur_app_state] : _state->_exist_apps) {
```

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -184,50 +209,97 @@ void meta_duplication_service::add_duplication(duplication_add_rpc rpc)
...
+                if (kv.first == request.app_name) {
+                    // Skip this app since we want ot check other apps.
```
acelyc111, line 249:
```suggestion
                    // Skip this app since we want to check other apps.
```

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -184,50 +209,97 @@ void meta_duplication_service::add_duplication(duplication_add_rpc rpc)
...
-                                                  duplication_add_rpc &rpc)
+                                                  duplication_add_rpc &rpc,
+                                                  const std::string &remote_app_name)
 {
-    const auto err = dup->start(rpc.request().is_duplicating_checkpoint);
+    const auto &err = dup->start(rpc.request().is_duplicating_checkpoint);
     if (dsn_unlikely(err != ERR_OK)) {
-        LOG_ERROR("start dup[{}({})] failed: err = {}", app->app_name, dup->id, err);
-        return;
+        LOG_ERROR_DUP_HINT_AND_RETURN(
```
acelyc111, line 291: Ditto.

The `return` word is missing seems a little confused although it's in the macro name.

idl/duplication.thrift
acelyc111: Is it safe to use app **name**, what happen if the app renaming in remote cluster? Or even swap app names in remote cluster?

src/shell/commands/duplication.cpp
```diff
@@ -80,35 +81,55 @@ bool add_dup(command_executor *e, shell_context *sc, arguments args)
...
-                   "{}], checkpoint: {}\n",
-                   app_name,
-                   remote_cluster_name,
-                   resp.appid,
-                   resp.dupid,
-                   is_duplicating_checkpoint);
+        fmt::print("\nWARNING: meta server does NOT support specifying remote_app_name, "
+                   "remote_app_name might has been specified with {}\n",
+                   remote_app_name,
+                   app_name);
```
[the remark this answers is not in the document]
empiredan: This is a bug, where `remote_app_name` should be removed.

idl/duplication.thrift
empiredan: Good idea, I think it's reasonable to rename app of remote cluster while duplicating. We could support this on another pull request.

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -158,76 +184,125 @@ void meta_duplication_service::add_duplication(duplication_add_rpc rpc)
...
-    for (const auto &ent : app->duplications) {
-        auto it = ent.second;
-        if (it->follower_cluster_name == request.remote_cluster_name) {
-            dup = ent.second;
-            break;
+    {
+        zauto_read_lock l(app_lock());
+
+        app = _state->get_app(request.app_name);
+        if (!app || app->status != app_status::AS_AVAILABLE) {
```
acelyc111, line 230: Can reuse LOG_WARNING_DUP_HINT_AND_RETURN_IF_NOT?

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -158,76 +184,125 @@ void meta_duplication_service::add_duplication(duplication_add_rpc rpc)
...
+            response.err = ERR_APP_NOT_EXIST;
+            LOG_WARNING("app {} was not found", request.app_name);
```
acelyc111, line 232: Because it also judge the `app->status`, so it's better to indicate this.

src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -158,76 +184,125 @@ void meta_duplication_service::add_duplication(duplication_add_rpc rpc)
...
+            // Check if other apps of this cluster are duplicated to the same remote app.
+            for (const auto & [ app_name, cur_app_state ] : _state->_exist_apps) {
+                if (app_name == request.app_name) {
+                    // Skip this app since we want to check other apps.
+                    continue;
+                }
+
+                for (const auto &ent : cur_app_state->duplications) {
+                    LOG_WARNING_DUP_HINT_AND_RETURN_IF_NOT(
+                        ent.second->remote_cluster_name != request.remote_cluster_name ||
```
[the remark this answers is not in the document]
empiredan: Actually the invalid case is that `dup_info->remote_cluster_name == request.remote_cluster_name && dup_info->remote_app_name == remote_app_name`.

acelyc111 (approved):

acelyc111 (approved):

GehaFearless (approved):

foreverneverer (approved):

PR #1970, opened 2024-04-09
fix(java-client): fix client renew TGT thread leak when request table doesnt exist

Samunroyu:
In the Kerberos environment, java client will check TGT and relogin
kerberos if needed every 10 seconds by a thread. But when client
requests table  which doesnt exist will cause threads leak.

The root cause this problem is when Pegasus client close the client
but don't close TGT thread pool.

This patch adds KerberosProtocol close function.


shalk:
Handle Exception  is a good choice.  Exception will make the schedule job not schedule again 

But i am not quite understand the main cause about `deamon`  
I don't kown the relatetionship between daemon thread with exception handler @Samunroyu 

if you set daemon false, you should shutdown the threadpool when client jvm shutdown.

Apache9:
> Handle Exception is a good choice. Exception will make the schedule job not schedule again
> 
> But i am not quite understand the main cause about `daemon` I don't kown the relatetionship between daemon thread with exception handler @Samunroyu
> 
> if you set daemon false, you should shutdown the threadpool when client jvm shutdown.

I‘ve talked with @Samunroyu offline, the root cause here is we do not have a close method for `KerberosProtocol` class, so if we create the client many time due to kerberos authentication error, although we have already closed all the created client, the `KerberosProtocol` will not be closed, so the ExecutorService will be leaked and cause too many threads.

shalk:
> > Handle Exception is a good choice. Exception will make the schedule job not schedule again
> > But i am not quite understand the main cause about `daemon` I don't kown the relatetionship between daemon thread with exception handler @Samunroyu
> > if you set daemon false, you should shutdown the threadpool when client jvm shutdown.
> 
> I‘ve talked with @Samunroyu offline, the root cause here is we do not have a close method for `KerberosProtocol` class, so if we create the client many time due to kerberos authentication error, although we have already closed all the created client, the `KerberosProtocol` will not be closed, so the ExecutorService will be leaked and cause too many threads.

It make sense. I get the case about thread leak.

if the pr finish and ci pass,  click request , i will review again @Samunroyu

java-client/src/main/java/org/apache/pegasus/rpc/interceptor/ReplicaSessionInterceptor.java
```diff
@@ -27,4 +27,6 @@ public interface ReplicaSessionInterceptor {
   // The behavior when rpc session is sending a message.
   // @returns false if this message shouldn't be sent.
   boolean onSendMessage(ReplicaSession session, final ReplicaSession.RequestEntry entry);
+
+  void close();
```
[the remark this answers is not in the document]
Apache9: You missed this interface.

java-client/src/main/java/org/apache/pegasus/rpc/interceptor/ReplicaSessionInterceptorManager.java
```diff
@@ -49,4 +49,10 @@ public boolean onSendMessage(ReplicaSession session, final ReplicaSession.Reques
     }
     return true;
   }
+
+  public void close() {
```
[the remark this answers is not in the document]
Apache9: And please add the `@Override` annotation for this method as it is inherited from the Closeable interface.

java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
```diff
@@ -50,19 +50,20 @@ class KerberosProtocol implements AuthProtocol {
...
+  private final String keyTab;
+  private final String principal;
   final int CHECK_TGT_INTEVAL_SECONDS = 10;
   final ScheduledExecutorService service =
       Executors.newSingleThreadScheduledExecutor(
           new ThreadFactory() {
             @Override
             public Thread newThread(Runnable r) {
-              Thread t = new Thread(r);
+              Thread t = new Thread(r, "TGT renew for pegasus");
```
Apache9, line 63: Better introduce a unique suffix or prefix, otherwise if there are multiple client you will see multiple threads with the same name.

java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
```diff
@@ -96,11 +97,23 @@ public boolean isAuthRequest(final ReplicaSession.RequestEntry entry) {
   }
 
   private void scheduleCheckTGTAndRelogin() {
+    Runnable checkTGTAndReLogin =
```
Apache9, line 100: Could still use lambda here.

```
    service.scheduleAtFixedRate(
        () -> {
            try {
              checkTGTAndRelogin();
            } catch (Exception e) {
              logger.warn(
                  "check TGT and ReLogin kerberos failed , will retry after {} seconds.",
                  CHECK_TGT_INTEVAL_SECONDS,
                  e);
            }
        }
        CHECK_TGT_INTEVAL_SECONDS,
        CHECK_TGT_INTEVAL_SECONDS,
        TimeUnit.SECONDS);
```

shalk:
LGTM

java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
```diff
@@ -50,19 +52,21 @@ class KerberosProtocol implements AuthProtocol {
...
+  private final String keyTab;
+  private final String principal;
   final int CHECK_TGT_INTEVAL_SECONDS = 10;
   final ScheduledExecutorService service =
       Executors.newSingleThreadScheduledExecutor(
           new ThreadFactory() {
             @Override
             public Thread newThread(Runnable r) {
-              Thread t = new Thread(r);
+              String timestamp = new SimpleDateFormat("HHmmss").format(new Date());
```
shalk, line 65: from JDK8 , advise migration to the [java.time](http://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) classes.
String timestamp = LocalTime.now().format(DateTimeFormatter.ofPattern("HHmmss"));

shalk (approved):
thread leak has fixed

java-client/src/main/java/org/apache/pegasus/security/KerberosProtocol.java
Samunroyu: > from JDK8 , advise migration to the [java.time](http://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) classes. String timestamp = LocalTime.now().format(DateTimeFormatter.ofPattern("HHmmss"));

ok, i have changed to use java.time.

Apache9 (approved):

acelyc111 (approved):

PR #1993, opened 2024-04-28
feat(duplication): allow remote replica count to be specified for duplication

empiredan:
While duplicating from a cluster A only including one node where all tables have
just one replica to another cluster B including at least 3 nodes where all tables have
3 replicas, the target table on B would be also created with one replica. Thus we
should allow a duplication to be created with specified replica count.

### DIFF
--- a/src/shell/commands/duplication.cpp
+++ b/src/shell/commands/duplication.cpp
@@ -83,12 +83,11 @@ bool add_dup(command_executor *e, shell_context *sc, arguments args)
 
     // Read the replica count of the remote app, if any.
     const auto &remote_replica_count_ss = cmd({"-r", "--remote_replica_count"});
-    const auto missing_remote_replica_count = static_cast<bool>(remote_replica_count_ss);
 
     // 0 represents that remote_replica_count is missing, which means the remote_replica_count
     // would be set as the replica count of source app.
     int32_t remote_replica_count = 0;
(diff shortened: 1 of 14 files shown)


src/shell/commands/duplication.cpp
```diff
@@ -82,14 +75,32 @@ bool add_dup(command_executor *e, shell_context *sc, arguments args)
...
-    std::string remote_app_name(cmd({"-a", "--remote_app_name"}, app_name).str());
+    const std::string remote_app_name(cmd({"-a", "--remote_app_name"}, app_name).str());
+
+    // Read the replica count of the remote app, if any.
+    const auto &remote_replica_count_ss = cmd({"-r", "--remote_replica_count"});
+
+    // 0 represents that remote_replica_count is missing, which means the remote_replica_count
+    // would be set as the replica count of source app.
+    int32_t remote_replica_count = 0;
+    if (static_cast<bool>(remote_replica_count_ss)) {
```
acelyc111, line 90: Is it possible to use macros in src/shell/command_helper.h to simplify the code?

you can improve the macros if they don't meet requirements.

acelyc111: For example `PARSE_OPT_UINT`

acelyc111 (approved):

GehaFearless (approved):

PR #1996, opened 2024-04-29
feat(plog): garbage collection is allowed to be enabled/disabled  for plog

empiredan:
Sometimes we have to disable plog garbage collection. For example, to duplicate
data to the target cluster, we could firstly disable plog garbage collection, then do
copy_data. After copy_data is finished, a duplication with DS_LOG status could be
added to continue to duplicate data in plog to target cluster; at the same time, plog
garbage collection certainly should be enabled again.

Therefore a new configuration is added to enable/disable plog garbage collection:

```diff
[replication]
+ plog_gc_enabled = true
```

### DIFF
--- a/src/replica/replica_chkpt.cpp
+++ b/src/replica/replica_chkpt.cpp
@@ -117,8 +117,8 @@ void replica::on_checkpoint_timer()
     }
 
     if (!is_plog_gc_enabled()) {
-        LOG_INFO_PREFIX("gc_private {}: skip gc because plog gc is disabled",
-                        enum_to_string(status()));
+        LOG_WARNING_PREFIX("gc_private {}: skip gc because plog gc is disabled",
+                           enum_to_string(status()));
         return;
     }
 
@@ -134,6 +134,15 @@ void replica::on_checkpoint_timer()
     decree min_confirmed_decree = _duplication_mgr->min_confirmed_decree();
     decree cleanable_decree = last_durable_decree;
     int64_t valid_start_offset = _app->init_info().init_offset_in_private_log;
+
+    if (min_confirmed_decree < 0 && is_duplication_master()) {
+        // Not sure whether the plog files could be dropped, because min_confirmed_decree
+        // is currently unavailable.
+        LOG_INFO_PREFIX("gc_private {}: skip gc because confirmed duplication progress "
+                        "is unknown",
+                        enum_to_string(status()));
+        return;
+    }
 
     if (min_confirmed_decree >= 0) {
         // Do not rely on valid_start_offset for GC during duplication.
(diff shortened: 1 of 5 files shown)


src/replica/replica_stub.cpp
```diff
@@ -2299,6 +2299,41 @@ void replica_stub::register_ctrl_command()
                 });
             }));
 
+        _cmds.emplace_back(::dsn::command_manager::instance().register_single_command(
```
acelyc111, line 2302: How about simplify the commands in format of `replica.plog-gc [true|false]` by `register_bool_command()`, like some other remote-commands, e.g. `replica.deny-client`.

src/replica/replica_chkpt.cpp
```diff
@@ -112,66 +112,71 @@ void replica::on_checkpoint_timer()
...
+        return;
+    }
 
-        if (is_duplication_plog_checking()) {
-            LOG_DEBUG_PREFIX("gc_private {}: skip gc because duplication is checking plog files",
-                             enum_to_string(status()));
-            return;
-        }
+    if (!is_plog_gc_enabled()) {
+        LOG_INFO_PREFIX("gc_private {}: skip gc because plog gc is disabled",
```
acelyc111, line 120: How about using WARNING level logs?

src/replica/replica_chkpt.cpp
```diff
@@ -112,66 +112,71 @@ void replica::on_checkpoint_timer()
...
+                            min_confirmed_decree,
+                            last_durable_decree);
+            cleanable_decree = min_confirmed_decree;
+        } else {
+            LOG_INFO_PREFIX("gc_private {}: min_confirmed_decree({}) last_durable_decree({})",
+                            enum_to_string(status()),
+                            min_confirmed_decree,
+                            last_durable_decree);
+        }
+    } else if (is_duplication_master()) {
```
acelyc111, line 155: How about return earier like:
```
if (is_duplication_master() && min_confirmed_decree < 0) {
    ...
    return;
}

if (min_confirmed_decree >= 0) {
    ...
}
```

src/replica/replica_chkpt.cpp
```diff
@@ -112,66 +112,71 @@ void replica::on_checkpoint_timer()
...
+    }
+
+    tasking::enqueue(
+        LPC_GARBAGE_COLLECT_LOGS_AND_REPLICAS,
+        &_tracker,
+        [this, plog, cleanable_decree, valid_start_offset] {
+            // run in background thread to avoid file deletion operation blocking
+            // replication thread.
+            if (status() == partition_status::PS_ERROR || status() == partition_status::PS_INACTIVE)
+                return;
```
acelyc111, line 170:
```suggestion
            if (status() == partition_status::PS_ERROR || status() == partition_status::PS_INACTIVE) {
                return;
            }
```

acelyc111 (approved):

foreverneverer (approved):

PR #1997, opened 2024-04-30
feat(java-client): add maven-thrift-plugin

### LINKED ISSUE #1664: Use thrift-maven-plugin to generate thrift sources instead of script
We could use [thrift-maven-plugin](https://mvnrepository.com/artifact/org.apache.thrift/thrift-maven-plugin) to generate thrift sources instead of scripts, just like what parquet has done:

- https://issues.apache.org/jira/browse/PARQUET-1506
- https://github.com/apache/parquet-java/pull/600
- https://github.com/apache/parquet-java/blob/master/parquet-format-structures/pom.xml

shalk:
fix #1664 

ref: https://issues.apache.org/jira/browse/THRIFT-1536
ref: https://github.com/apache/parquet-mr/pull/600

##### Tests 

- Unit test
- Integration test
- No code

##### Code changes

- Has persistent data change

##### Side effects

- Increased code complexity

##### Related changes

### DIFF
--- a/java-client/pom.xml
+++ b/java-client/pom.xml
@@ -69,6 +69,7 @@
     <netty.version>4.1.85.Final</netty.version>
     <zstd-jni.version>1.3.7-1</zstd-jni.version>
     <libthrift.version>0.11.0</libthrift.version>
+    <thrift-maven-plugin.version>0.10.0</thrift-maven-plugin.version>
     <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
     <zookeeper.version>3.7.2</zookeeper.version>
     <doclint>none</doclint>
@@ -473,7 +474,7 @@
       <plugin>
         <groupId>org.apache.thrift</groupId>
         <artifactId>thrift-maven-plugin</artifactId>
-        <version>0.10.0</version>
+        <version>${thrift-maven-plugin.version}</version>
         <configuration>
           <thriftSourceRoot>${project.basedir}/../idl</thriftSourceRoot>
           <includes>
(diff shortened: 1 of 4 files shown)


java-client/pom.xml
```diff
@@ -470,6 +470,36 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.thrift</groupId>
+        <artifactId>thrift-maven-plugin</artifactId>
+        <version>0.10.0</version>
```
acelyc111, line 476: For compatibility reason, it's recommend to use 0.11.0 version.

java-client/scripts/download_thrift.sh
```diff
@@ -48,21 +48,4 @@ if ! $thrift -version | grep "0.11.0" ; then
     exit 1
 fi
 
-TMP_DIR=./gen-java
```
acelyc111, line 51: It seems `thrift` will not be used any where in java-client, can we remove this file entirely?

shalk: thrift will be used when `mvn compile` or `mvn test`. If i remove the file , if i remove the file  https://github.com/apache/incubator-pegasus/blob/master/.github/workflows/test_python-client.yml#L68 will be add in to github action

java-client/pom.xml
shalk: This is plugin version, i found this is the latest version. thrift java lib version will be 0.11.0 which defined in libthrift.version

java-client/scripts/download_thrift.sh
acelyc111: OK, let's keep it in current step. We can do that in next patches.

acelyc111 (approved):

empiredan (approved):

PR #2027, opened 2024-05-24
fix(duplication): parameters are not recognized while parsing add_dup shell command

empiredan:
https://github.com/apache/incubator-pegasus/issues/2028

Fix the problem that parameters failed to be extracted while parsing
add_dup command, and refactor the shell display of duplication
by text style and `println` both provided by `libfmt`.


src/shell/commands/duplication.cpp
```diff
@@ -87,6 +84,14 @@ bool add_dup(command_executor *e, shell_context *sc, arguments args)
     uint32_t remote_replica_count = 0;
     PARSE_OPT_UINT(remote_replica_count, 0, {"-r", "--remote_replica_count"});
 
+    fmt::print("trying to add duplication [app_name: {}, remote_cluster_name: {}, "
```
acelyc111, line 87: We can use fmt::println to avoid adding `\n` manually.

Other places are the same.

src/shell/commands/duplication.cpp
```diff
@@ -115,19 +124,34 @@ bool add_dup(command_executor *e, shell_context *sc, arguments args)
...
+    fmt::print("]\n");
+
+    if (!resp.__isset.remote_app_name) {
+        fmt::print("WARNING: meta server does NOT support specifying remote_app_name, "
+                   "remote_app_name might has been specified with '{}'\n",
+                   app_name);
+    }
+
+    if (!resp.__isset.remote_replica_count) {
+        fmt::print("WARNING: meta server does NOT support specifying remote_replica_count, "
```
acelyc111, line 153: For warning message, we can enable coloring by https://fmt.dev/latest/api.html#terminal-color-and-text-style.

acelyc111 (approved):

levy5307 (approved):

PR #2045, opened 2024-06-18
feat(remote_command): provide the query for the progress of decrees including both local writes and duplications

empiredan:
There are many kinds of decrees while writing locally and duplicating to remote
clusters, for example, the max decree in prepare list, the last decree that has ever
been committed, the last decree that has been applied into rocksdb memtable,
the last decree that has been flushed into rocksdb sst files, the max decree that
has been confirmed by remote cluster for duplication, etc..

These decrees are very useful while we want to watch the progress of all the local
writes and duplications. These decrees might also help us diagnose the problems.
Therefore, we provide a tool in the way of `remote_command` to show the decrees
for each replica.


src/replica/duplication/replica_duplicator_manager.cpp
```diff
@@ -47,23 +50,27 @@ replica_duplicator_manager::get_duplication_confirms_to_update() const
...
-                continue;
-            }
-            duplication_confirm_entry entry;
-            entry.dupid = duplicator->id();
-            entry.confirmed_decree = p.last_decree;
-            entry.__set_checkpoint_prepared(p.checkpoint_has_prepared);
-            updates.emplace_back(entry);
+    for (const auto & [ _, dup ] : _duplications) {
+        const auto &progress = dup->progress();
+        if (progress.last_decree == progress.confirmed_decree &&
```
acelyc111, line 55: This judgement is a bit of complex, could you please add some comments?

src/replica/duplication/replica_duplicator_manager.cpp
```diff
@@ -191,5 +198,30 @@ replica_duplicator_manager::get_dup_states() const
...
+    std::transform(_duplications.begin(),
+                   _duplications.end(),
+                   std::inserter(dup_messages, dup_messages.end()),
+                   [](const decltype(_duplications)::value_type &dup) {
+                       const auto &progress = dup.second->progress();
+                       return std::make_pair(
+                           dup.first,
+                           fmt::format("dupid={}, remote_cluster_name={}, remote_app_name={}, "
+                                       "confirmed_decree={}, persisted_decree={}",
+                                       dup.first,
```
acelyc111, line 216: dupid_t is treated as the map's key, is it necessary to add it here duplicately?

src/replica/duplication/replica_duplicator_manager.cpp
```diff
@@ -191,5 +198,30 @@ replica_duplicator_manager::get_dup_states() const
...
+                                       dup.second->remote_cluster_name(),
```
acelyc111, line 217: How about add a stringfy function for replica_duplicator, then we don't need to access its members one by one here?

src/replica/replica.cpp
```diff
@@ -595,6 +597,35 @@ decree replica::last_prepared_decree() const
...
+                               _app->last_flushed_decree(),
+                               last_durable_decree(),
+                               _private_log->max_gced_decree(get_gpid()));
+
+    const auto &dup_messages = _duplication_mgr->get_progress_messages();
+    if (dup_messages.empty()) {
+        return message;
+    }
+
+    message += "\n        duplications {";
```
acelyc111, line 619: Would it be more appropriate if we return a JSON format string, then the thirdparty tools are convenient to parse the result?

acelyc111 (approved):

foreverneverer (approved):

PR #2048, opened 2024-06-21
fix(duplication): reduce the delay before last mutation is duplicated to the remote cluster

empiredan:
https://github.com/apache/incubator-pegasus/issues/2050

As is described by the issue, the problem is that we have to waits 2 ~ 3 minutes
(until some empty write gets in) before the last mutation is duplicated to the remote
cluster.

The reason is that the last committed decree of the last mutation (i.e.
`mutation.data.header.last_committed_decree`), rather than the decree of the
last mutation (i.e. `mutation.data.header.decree`), is chosen as the max decree
that is duplicated to the remote cluster. Instead, the max committed decree should
be chosen as the max decree that is duplicated to the remote cluster.

After the optimization, the delay has been reduced from 2 ~ 3 minutes to about
0.1 seconds.

### DIFF
--- a/src/replica/duplication/mutation_batch.cpp
+++ b/src/replica/duplication/mutation_batch.cpp
@@ -27,6 +27,7 @@
 #include "metadata_types.h"
 #include "mutation_batch.h"
 #include "replica_duplicator.h"
+#include "replica/replica.h"
 #include "runtime/task/task_code.h"
 #include "runtime/task/task_spec.h"
 #include "utils/autoref_ptr.h"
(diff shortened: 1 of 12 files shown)


src/replica/mutation_log.h
```diff
@@ -76,67 +76,73 @@ class mutation_log : public ref_counter
...
+    // Flush the pending buffer at most once.
+    //
+    // Thread safe.
     virtual void flush_once() = 0;
 
 public:
     //
-    // ctors
+    // Ctors
     // when is_private = true, should specify "private_gpid"
```
acelyc111, line 121: Can we remove is_private now, it must be true after the slog has been removed ?

src/replica/duplication/mutation_batch.cpp
```diff
@@ -123,6 +126,15 @@ error_s mutation_batch::add(mutation_ptr mu)
             _start_decree);
     }
 
+    if (mu->get_decree() <= _replica->last_applied_decree()) {
+        // Once this mutation has been applied into rocksdb memtable, commit it for duplication;
+        // otherwise, this mutation would be delayed at least several minutes to be duplicated to
+        // the remote cluster. It would not be duplicated until some new mutations (such as empty
+        // writes) enter, since the last decree that is committed for this replica is NOT
+        // mu->data.header.decree but rather mu->data.header.last_committed_decree.
```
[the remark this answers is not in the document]
empiredan: OK.

src/replica/mutation_log.h
empiredan: Yes, actually `slog` has been removed gradually, I'll create another PRs to drop `_is_private`.

acelyc111 (approved):

foreverneverer (approved):

PR #2049, opened 2024-06-24
refactor(FQDN): further more refator on idl/dsn.layer2.thrift

acelyc111:
This is a further refactor on idl/dsn.layer2.thrift, the main effects are on
partition_configuration structure.

To keep compatible and support rolling update, it should be accepted that partial
servers or clients with the FQDN supporting feature and partial of them without
the feature. In this patch, macros like `GET_HOST_PORT` and `GET_HOST_PORTS` are
used in almost all places where partition_configuration objects appear. (Unit
tests don't require this because the tests must in the latest version which has
this feature).

When invoking a RPC, it's preferred to use host_port because we suppose that the
IP address is possible to be changed, but the hostname is permanent.

When received a proposol request, the optional `hp_primary` and `hp_secondaries`
will be filled by the related host_ports of the `primary` and `secondaries` and
fields. Then we can simplily deal with the logic.

### DIFF
--- a/src/replica/replica_config.cpp
+++ b/src/replica/replica_config.cpp
@@ -276,6 +276,7 @@ void replica::downgrade_to_secondary_on_primary(configuration_update_request &pr
     }
 
     CHECK_EQ(proposal.config.pid, _primary_states.pc.pid);
+    CHECK_EQ(proposal.config.primary, _primary_states.pc.primary);
     dsn::host_port primary;
     GET_HOST_PORT(proposal.config, primary, primary);
     // The local host_port type 'hp_primary' field must be set.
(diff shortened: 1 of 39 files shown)


acelyc111:
Please ignore the clang-tidy warnings, because there are so many warings, it would introduce too many clang-tidy related changes as well which may cause the main changes of this pull request hard to distinguish and review.

src/replica/replica_config.cpp
```diff
@@ -191,15 +192,24 @@ void replica::add_potential_secondary(const configuration_update_request &propos
 
     CHECK_EQ(proposal.config.ballot, get_ballot());
     CHECK_EQ(proposal.config.pid, _primary_states.pc.pid);
-    CHECK_EQ(proposal.config.hp_primary, _primary_states.pc.hp_primary);
-    CHECK(proposal.config.hp_secondaries == _primary_states.pc.hp_secondaries, "");
-
+    CHECK_EQ(proposal.config.primary, _primary_states.pc.primary);
```
Samunroyu, line 195: why funct add_potnetial_secondary need to check rpc_address of proposal.config.primary equal to _primary_states.pc.primary?
And func downgrade_to_secondary_on_primary is not check?

Samunroyu (dismissed):

PR #2051, opened 2024-06-25
refactor: Simplify code on partition_configuration

acelyc111:
There is no functional changes, but only refactoring, includes:
- Use range-based loop
- Unify the naming of partition_configuration variables to `pc` (or `pcs`)
- Use `fmt::join` for simplify container formating
- Reduce some if-else statements depth

This patch dosen't change the names in thrift IDL which leads more changes on
client drivers, we can do that in following patches.

### DIFF
--- a/src/meta/meta_split_service.cpp
+++ b/src/meta/meta_split_service.cpp
@@ -119,15 +119,15 @@ void meta_split_service::do_start_partition_split(std::shared_ptr<app_state> app
         app->helpers->split_states.splitting_count = app->partition_count;
         app->partition_count *= 2;
         app->helpers->contexts.resize(app->partition_count);
-        app->partitions.resize(app->partition_count);
+        app->pcs.resize(app->partition_count);
         _state->get_table_metric_entities().resize_partitions(app->app_id, app->partition_count);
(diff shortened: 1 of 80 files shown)


src/meta/meta_split_service.cpp
```diff
@@ -307,16 +307,17 @@ void meta_split_service::on_add_child_on_remote_storage_reply(error_code ec,
     update_child_request->type = config_type::CT_REGISTER_CHILD;
     SET_OBJ_IP_AND_HOST_PORT(*update_child_request, node, request, primary);
 
-    partition_configuration child_config = app->partitions[child_gpid.get_partition_index()];
-    child_config.secondaries = request.child_config.secondaries;
-    child_config.__set_hp_secondaries(request.child_config.hp_secondaries);
+    // TODO(yingchun): should use conference?
+    auto child_pc = app->pcs[child_gpid.get_partition_index()];
```
acelyc111, line 311: This may be a bug, I'll fix it in next patch.

empiredan (approved):

foreverneverer (dismissed):

empiredan (approved):

foreverneverer (dismissed):

empiredan (approved):

GehaFearless (approved):

PR #2063, opened 2024-07-05
refactor(format): Bump clang-format to 14

acelyc111:
After GitHub actions forced to run on node20 [1], and node20 depends
on GLIBC_2.28, we have to run actions on newer operation systems which
has higher builtin glibc.

Before this patch, we are using clang-format-3.9 to format C++ code,
but if we using newer OS (say Ubuntu 22.04), the clang-format-3.9 is
too old and it's difficult to install such an old version.

This patch bumps the clang-format to 14 as the code format tool, and
update relative chores, such as updating `cpp_clang_format_linter`
action job in .github/workflows/lint_and_test_cpp.yaml, removing
clang-format-3.9 docker images, adding more options in `.clang-format`
(almost all of the options are kept as old version and default values).

The main part of this patch is the C++ code updating according to the
newer clang-format, they are formated automatically.

1. https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/


acelyc111:
I'll fix the failed "Build Debug on CentOS 7" job in following patches.

empiredan (approved):

Issue #2015, opened 2024-05-22
Bug(duplication):some nodes never start GC plog after computer room failure


ninsmiracle:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
The computer room which service for our duplication master cluster meet an accidents. And most of the node in this room shutdown in a short time.
When all the nodes alive , we found some partition of the duplication table never GC private log (plog) again.

2. What did you expect to see?
All the partition can GC it's plog correctly.

4. What did you see instead?
stdout (error log):
```
// stdout
90146:E2024-05-14 15:59:52.512 (1715673592512665104 67086) replica.default8.040005fe0319646c: nfs_server_impl.cpp:221:on_get_file_size(): {nfs_service} get stat of file /home/work/ssd2/pegasus/alsgsrv-monetization-master/replica/reps/8.53.pegasus/plog/log.18129.608864535790 failed, err = No such file or directory
```
We can see this replica request a old plog. 
![image](https://github.com/apache/incubator-pegasus/assets/110282526/bc7ac83b-c6ea-4a3a-9fd2-d79a95d7b8f2)

Because the partition can not clear plog as nomarl,so the disk always fully. We need to clear the plog sometimes.

5. What version of Pegasus are you using?
Pegasus v2.4

acelyc111:
Is the file actually exists or not?

ninsmiracle:
> Is the file /home/work/ssd2/pegasus/alsgsrv-monetization-master/replica/reps/8.53.pegasus/plog/log.18129.608864535790 actually exists or not?

When coredump happened, the file actually not exists.

PR #2066, opened 2024-07-09
fix(duplication): fix plog do not gc after meta server restart

ninsmiracle:
#2015

When meta load duplication status from ZK, refresh `duplicating` string in memory ,base on the value in ZK.

##### Tests 
- Unit test

### DIFF
--- a/src/meta/duplication/meta_duplication_service.cpp
+++ b/src/meta/duplication/meta_duplication_service.cpp
@@ -54,6 +54,7 @@
 #include "utils/ports.h"
 #include "utils/string_conv.h"
 #include "utils/zlocks.h"
+#include "utils/defer.h"
 
 DSN_DECLARE_bool(dup_ignore_other_cluster_ids);
 
@@ -701,6 +702,10 @@ void meta_duplication_service::recover_from_meta_state()
         _meta_svc->get_meta_storage()->get_children(
             get_duplication_path(*app),
             [this, app](bool node_exists, const std::vector<std::string> &dup_id_list) {
+                dsn::defer([this, app]() {
+                    zauto_write_lock l(app_lock());
+                    refresh_duplicating_no_lock(app);
+                });
                 if (!node_exists) {
                     // if there's no duplication
                     return;


src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -701,6 +702,10 @@ void meta_duplication_service::recover_from_meta_state()
         _meta_svc->get_meta_storage()->get_children(
             get_duplication_path(*app),
             [this, app](bool node_exists, const std::vector<std::string> &dup_id_list) {
+                dsn::defer([this, app]() {
```
acelyc111, line 705: The memory state will be updated [1] after get data from ZK successfully in line 721, why update it again?

1. https://github.com/apache/incubator-pegasus/blob/master/src/meta/duplication/meta_duplication_service.cpp#L787

ninsmiracle: > The memory state will be updated [1] after get data from ZK successfully in line 721, why update it again?
> 
> 1. https://github.com/apache/incubator-pegasus/blob/master/src/meta/duplication/meta_duplication_service.cpp#L787

Because in function `recover_from_meta_state`, the vector `dup_id_list` may be empty.
Here is the detail of our online condition:
1. Delete the existing hot standby first
    A. Delete the corresponding dup from zk
    B. Set duplicating in meta_server memory to 0 (not saved in zk), duplicating in zk is still 1
2. meta_server restarts
  Read dup and app_info data from zk to memory. At this time, dup is empty, but duplicating in zk is still 1 (because dup is empty, it is impossible to delete the hot standby to make duplicating 0).
3. Meta passes the hot standby status to the replica through `on_config_sync`
4.  Replica GC checks that the dup map is empty (`min_confirmed_decree = invalid_decree`), but `_is_duplication_master = true`, so GC is not performed

acelyc111: Thanks for the information. So this patch is aim to complete step 2? Why not complete the 1.B step?

ninsmiracle: > Thanks for the information. So this patch is aim to complete step 2? Why not complete the 1.B step?

Both two ways can fix this problem. In my opinion, the ways like step 2 is easier to implement.

acelyc111: Correctness and consistent is important. Update the `duplicating` status to ZK is not very difficult if all duplications removed.
Or, if the `duplicating` memory status can be recovered from `<app>/duplication` ZK path, just competely remove the `duplicating` status from ZK.

Issue #2014, opened 2024-05-22
Bug(duplication):some nodes coredump after start duplication for a long time


ninsmiracle:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
-Deploy duplication matser and back-up cluster.
-Begin duplicate.
-Run about 2~3 days.
-Some nodes coredump

2. What did you expect to see?
Node run as normal.

3. What did you see instead?
memory monitoring table.
![image](https://github.com/apache/incubator-pegasus/assets/110282526/16c9406a-a96a-4d01-b980-a01a3ba0a166)

coredump detail:
```
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/home/work/app/pegasus/c3srv-browser/replica/package/bin/pegasus_server config.'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f01575401d7 in raise () from /lib64/libc.so.6
(gdb) bt
#0  0x00007f01575401d7 in raise () from /lib64/libc.so.6
#1  0x00007f01575418c8 in abort () from /lib64/libc.so.6
#2  0x00007f015c628f9e in dsn_coredump () at /home/work/temp/format_pegasus/pegasus/src/rdsn/src/runtime/service_api_c.cpp:93

#3  0x00007f015c422c83 in dsn::replication::log_file::log_file (this=0x73aa4a630, path=0x740561c98 "/home/work/ssd2/pegasus/c3srv-browser/replica/reps/72.173.pegasus/plog/log.92534.3105061495163", 
    handle=<optimized out>, index=<optimized out>, start_offset=3105061495163, is_read=<optimized out>) at /home/work/temp/format_pegasus/pegasus/src/rdsn/src/replica/log_file.cpp:166

#4  0x00007f015c4247ce in dsn::replication::log_file::open_read (path=0x740561c98 "/home/work/ssd2/pegasus/c3srv-browser/replica/reps/72.173.pegasus/plog/log.92534.3105061495163", err=...)
    at /home/work/temp/format_pegasus/pegasus/src/rdsn/src/replica/log_file.cpp:92
#5  0x00007f015c43ccfa in dsn::replication::log_utils::open_read (path=..., file=...) at /home/work/temp/format_pegasus/pegasus/src/rdsn/src/replica/mutation_log_utils.cpp:43
[...]

PR #2068, opened 2024-07-09
fix(duplication): deal with plog concurrent problem

ninsmiracle:
#2014

Old logic: If some file open with err  `ERR_FILE_OPERATION_FAILED`,  the others file could as a paramater to build a log_file object. However, plog GC thread can remove the file between those two actiion.(Just as the condition following...)
image

New logic: Add some judgement logic in `find_log_file_to_start` to prevent those plog (which has possible already be removed) from `open_read`.

##### Tests 

- Unit test

### DIFF
--- a/src/replica/duplication/load_from_private_log.cpp
+++ b/src/replica/duplication/load_from_private_log.cpp
@@ -158,9 +158,11 @@ void load_from_private_log::find_log_file_to_start()
     // is cleared once WAL replay finished. They are unable to read.
     mutation_log::log_file_map_by_index new_file_map;
     decree cleanable_decree = _private_log->get_cleanable_decree();
-    decree max_decree_gpid =  _private_log->max_decree(get_gpid());
-    if(max_decree_gpid <= cleanable_decree){
-        LOG_ERROR_PREFIX("plog_file all error: max_decree_gpid {} , cleanable_decree {}", max_decree_gpid, cleanable_decree);
+    decree max_decree_gpid = _private_log->max_decree(get_gpid());
+    if (max_decree_gpid <= cleanable_decree) {
+        LOG_ERROR_PREFIX("plog_file all error: max_decree_gpid {} , cleanable_decree {}",
+                         max_decree_gpid,
+                         cleanable_decree);
         return;
     }
 
@@ -176,9 +178,10 @@ void load_from_private_log::find_log_file_to_start()
         // next file map may can not open
         gpid pid = get_gpid();
         decree previous_log_max_decree = file->previous_log_max_decree(pid);
-        // these plog_file has possible be deleted do not open_read next plog_file , otherwise it may coredump
-        if(previous_log_max_decree <= cleanable_decree){
-            break ;
+        // these plog_file has possible be deleted do not open_read next plog_file , otherwise it
+        // may coredump
+        if (previous_log_max_decree <= cleanable_decree) {
+            break;
         }
     }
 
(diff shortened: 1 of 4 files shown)


ninsmiracle:
Here are some picture can make this problem clearly:

-  Firstly, we get `_current` plog file in `find_log_file_to_start`
image

This process is shown in the schematic diagram：
image

This time, `current` will be set to plog file 1.

- Then , in plog GC logic  `mutation_log::garbage_collection`
image

Traverse from new to old, find a file whose max_decree is less than cleanable, delete this plog file and all previous files.
image

In extreme cases, cleanable is `_start_decree - 1` (it can never be greater than start_decree).
So no matter what, when `_current` is specified as 1, file 2 will not be judged to need GC. (So,`switch_to_next_log_file` opens the potential file 2 in the figure above, and there will be no concurrency issues.)

src/replica/duplication/load_from_private_log.cpp
```diff
@@ -156,14 +157,32 @@ void load_from_private_log::find_log_file_to_start()
     // Reopen the files. Because the internal file handle of `file_map`
     // is cleared once WAL replay finished. They are unable to read.
     mutation_log::log_file_map_by_index new_file_map;
-    for (const auto &pr : file_map) {
+    decree cleanable_decree = _private_log->get_cleanable_decree();
+    decree max_decree_gpid = _private_log->max_decree(get_gpid());
+    if (max_decree_gpid <= cleanable_decree) {
```
acelyc111, line 162: Is it possible? `CHECK` it if it's impossible.

src/replica/duplication/load_from_private_log.cpp
```diff
@@ -156,14 +157,32 @@ void load_from_private_log::find_log_file_to_start()
...
+        LOG_ERROR_PREFIX("plog_file all error: max_decree_gpid {} , cleanable_decree {}",
+                         max_decree_gpid,
+                         cleanable_decree);
+        return;
+    }
+
+    for (auto it = file_map.rbegin(); it != file_map.rend(); ++it) {
```
acelyc111, line 169: Use const reference?

src/replica/duplication/load_from_private_log.cpp
```diff
@@ -156,14 +157,32 @@ void load_from_private_log::find_log_file_to_start()
...
-        error_s es = log_utils::open_read(pr.second->path(), file);
+        error_s es = log_utils::open_read(it->second->path(), file);
         if (!es.is_ok()) {
             LOG_ERROR_PREFIX("{}", es);
             return;
         }
-        new_file_map.emplace(pr.first, file);
+        new_file_map.emplace(it->first, file);
+
+        // next file map may can not open
```
acelyc111, line 178: Upper the first letter and add a dot at the end. Other places are the same.

And which code line does this comment for?

src/replica/duplication/load_from_private_log.cpp
```diff
@@ -156,14 +157,32 @@ void load_from_private_log::find_log_file_to_start()
...
+        gpid pid = get_gpid();
+        decree previous_log_max_decree = file->previous_log_max_decree(pid);
+        // these plog_file has possible be deleted do not open_read next plog_file , otherwise it
```
acelyc111, line 181: What is `plog_file`? Why you add a underline?

src/replica/duplication/load_from_private_log.cpp
ninsmiracle: Done. And this comment is the same meaning with line 181. So I deleted it.

PR #2078, opened 2024-07-18
perf: enhance the loading process of replicas particularly when a significant number of replicas are spread across multiple disks

empiredan:
Immediately after the replica server is started, all of the replicas under the data
directory would be loaded. Currently, a loading task is launched for each replica
directory. The tasks for loading replica directories are pushed into a partitioned
thread pool (namely `THREAD_POOL_REPLICATION`) disk by disk: only after all
of the replica directories on the current disk have been pushed into the thread
pool would the process move on to the next disk. Since the thread pool is partitioned
while the hash for each task is the total number of the tasks that have been in the
pool before this task is added, all of the replica directories on one disk would be
executed concurrently. This would lead to two problems once there are a great number
of replica directories on each disk:

- I/O usage for each disk might become saturated: its `%util` might become 100%;
- The entire loading process is blocked on each single disk: during a long period only
one disk is keeping busy while others are idle.

The replica server seems getting stuck in loading replicas after it is started. This is
unacceptable and should be changed.

The improved version allows the replica directories on different disks to be loaded
simultaneously: every disk would be busy loading replicas. Also, loading tasks would
be pushed into a non-partitioned thread pool (i.e. `THREAD_POOL_LOCAL_APP`)
instead of the partitioned, making tasks across multiple threads auto-balanced to
prevent some threads from being starved while others are stuffed.

And new parameter is added to restrict the max number of replicas allowed to be loaded
simultaneously for each disk, in case that I/O usage for each disk becomes saturated.
Another parameter is added to ensure that the main thread waiting all loading tasks to
finished would not be blocked on one task too long while the number of tasks for loading
replica directories simultaneously on a single disk has reached its limit.

Parameters are added as follows:

```diff
[replication]
+ max_replicas_on_load_for_each_disk = 256
+ load_replica_max_wait_time_ms = 10
```


src/replica/replica_stub.h
```diff
@@ -357,8 +365,36 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
+    // Get the dir name for a replica from a potentially longer path (both absolute and
+    // relative paths are possible).
+    static std::string get_replica_dir_name(const std::string &dir);
+
+    // Parse app id, partition id and app type from the dir name of a replica.
+    static bool
+    parse_replica_dir_name(const std::string &dir_name, gpid &pid, std::string &app_type);
+
+    // Load an existing replica which is located in `dn` with `dir`. Usually each different
+    // `dn` represents a unique disk. `dir` is the absolute path of the directory for a
```
acelyc111, line 383:
```suggestion
    // `dn` represents an unique disk. `dir` is the absolute path of the directory for a
```

src/replica/replica_stub.h
```diff
@@ -357,8 +365,36 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
+    // replica.
+    virtual replica_ptr load_replica(dir_node *dn, const char *dir);
```
acelyc111, line 385:
```suggestion
    virtual replica_ptr load_replica(dir_node *dn, const char *replica_dir);
```
Clarify it's the replica's dir, not the dir_node's.

src/replica/replica_stub.h
```diff
@@ -357,8 +365,36 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
+
+    // The same as the above `load_replica` function, except that this function is to load
+    // each replica to `reps` with protection from `reps_lock`.
+    void
+    load_replica(dir_node *dn, const std::string &dir, utils::ex_lock &reps_lock, replicas &reps);
+
+    // Load all replicas synchronously from all disks to `reps`. This function would ensure
+    // that data on each disk is loaded more evenly, rather than that a disk would begin to
+    // be loaded only after another has been finished, in case that there are too many replicas
+    // on a disk and other disks cannot start loading until this disk is finished.
```
acelyc111, line 395: Just simplified as this?
```suggestion
    // Load all replicas simultaneously from all disks to `reps`.
```

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+                                const std::string &dir,
+                                utils::ex_lock &reps_lock,
+                                replicas &reps)
+{
+    LOG_INFO("loading replica: tag={}, dir={}", dn->tag, dir);
+
+    const auto *const worker = task::get_current_worker2();
+    if (worker != nullptr) {
+        CHECK(!(worker->pool()->spec().partitioned),
+              "The thread pool for loading replicas must not be partitioned since load balancing "
```
acelyc111, line 464: It would be better to mention which the thread pool is, so that the administators could know how to adjust the config.

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+             dir,
+             rep->last_durable_decree(),
+             rep->last_committed_decree(),
+             rep->last_prepared_decree());
+
+    utils::auto_lock<utils::ex_lock> l(reps_lock);
+    CHECK(reps.find(rep->get_gpid()) == reps.end(),
+          "conflict replica dir: {} <--> {}",
+          rep->dir(),
+          reps[rep->get_gpid()]->dir());
```
acelyc111, line 487: Better to reuse the found iterator to avoid finding it again.

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+        for (size_t disk_index = 0; disk_index < disks.size(); ++disk_index) {
+            // Structured bindings can be captured by closures in g++, while not supported
+            // well by clang. Thus we do not use following statement to bind both variables
+            // until clang has been upgraded to version 16 which could support that well:
+            //
+            //     const auto &[dn, dirs] = disks[disk_index];
+            //
+            // For the docs of clang 16 please see:
+            //
+            // https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#c-20-feature-support:
```
acelyc111, line 514:
```suggestion
            // https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#c-20-feature-support.
```

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+
+    reps[rep->get_gpid()] = rep;
+}
+
+void replica_stub::load_replicas(replicas &reps)
+{
+    const auto &disks = get_all_disk_dirs();
+
+    std::vector<size_t> dir_indexes(disks.size(), 0);
+    std::vector<std::queue<std::pair<std::string, task_ptr>>> load_disk_queues(disks.size());
```
acelyc111, line 497: Better to add comments to describe what are they used for.

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+                } else {
+                    // There might be too many replicas that are being loaded which lead to
+                    // slow disk IO.
+                    LOG_WARNING("after {} ms, loading dir({}) is still not finished, there are "
+                                "{} replicas being loaded for disk(index={}, tag={}, path={}), "
+                                "skip dir(index={}, path={}), turn to next disk",
+                                FLAGS_load_replica_max_wait_time_ms,
+                                load_disk_queue.front().first,
+                                load_disk_queue.size(),
+                                disk_index,
```
acelyc111, line 541: The `disk_index` is just an internal variable, it may confused, is it necessary to be logged?

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+                // All of the replicas for the disk `disks[disk_index]` have begun to be loaded,
+                // thus just skip.
+                ++finished_disks;
+                continue;
+            }
+
+            const auto &dn = disks[disk_index].first;
+            auto &load_disk_queue = load_disk_queues[disk_index];
+            if (!load_disk_queue.empty() &&
+                load_disk_queue.size() >= FLAGS_max_replicas_on_load_for_each_disk) {
```
acelyc111, line 528:
```suggestion
            if (load_disk_queue.size() >= FLAGS_max_replicas_on_load_for_each_disk) {
```

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+                // Loading replicas should be throttled in case that disk IO is saturated.
+                if (load_disk_queue.front().second->wait(FLAGS_load_replica_max_wait_time_ms)) {
```
acelyc111, line 530: It seems this patch implemented a theadpool-with-max-threads, isn't it? The benifit compare to the former implementation is now we can limit the `max_replicas_on_load_for_each_disk`.

Could the rocksdb::ThreadPool work well here? https://github.com/apache/incubator-pegasus/blob/master/src/shell/commands/local_partition_split.cpp#L392

src/replica/replica_stub.cpp
```diff
@@ -2015,7 +2164,6 @@ replica *replica_stub::load_replica(dir_node *dn, const char *dir)
     const auto err = rep->initialize_on_load();
     if (err != ERR_OK) {
         LOG_ERROR("{}: load replica failed, err = {}", rep->name(), err);
-        rep->close();
```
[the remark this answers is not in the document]
empiredan: In that it would be called immediately by `delete rep;`.

src/replica/replica_stub.h
empiredan: I think `a` should be put before `unique` which begins with a consonant sound `j`.

src/replica/replica_stub.h
```diff
@@ -357,8 +365,36 @@ class replica_stub : public serverlet<replica_stub>, public ref_counter
...
-    replica *load_replica(dir_node *dn, const char *dir);
+
+    using disk_dirs = std::vector<std::pair<dir_node *, std::vector<std::string>>>;
+
+    // Get the absolute dirs of all replicas for all disks.
+    disk_dirs get_all_disk_dirs() const;
+
+    // Get the dir name for a replica from a potentially longer path (both absolute and
+    // relative paths are possible).
+    static std::string get_replica_dir_name(const std::string &dir);
```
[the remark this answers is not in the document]
empiredan: OK, added.

src/replica/replica_stub.h
empiredan: OK.

src/replica/replica_stub.h
empiredan: OK.

src/replica/replica_stub.cpp
empiredan: OK.

src/replica/replica_stub.cpp
empiredan: OK, use progress instead.

src/replica/replica_stub.cpp
```diff
@@ -411,6 +432,173 @@ void replica_stub::initialize(bool clear /* = false*/)
...
+                                load_disk_queue.size(),
+                                disk_index,
+                                dn->tag,
+                                dn->full_dir,
+                                dir_index,
+                                dirs[dir_index]);
+                    continue;
+                }
+
+                // Continue to load a replica since we are within the limit now.
```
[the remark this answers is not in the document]
empiredan: Temporarily, since disk IO might be saturated. Would try to load for the next round.

src/replica/replica_stub.cpp
```diff
@@ -495,75 +683,18 @@ void replica_stub::initialize(const replication_options &opts, bool clear /* = f
...
-                },
-                load_tasks.size()));
-            load_tasks.back()->enqueue();
-        }
-    }
-    for (auto &tsk : load_tasks) {
-        tsk->wait();
-    }
-    uint64_t finish_time = dsn_now_ms();
+    utils::chronograph chrono;
```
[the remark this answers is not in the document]
empiredan: OK, use `SCOPED_LOG_TIMING` instead.

src/replica/replica_stub.cpp
empiredan: Tried before, however it required RDSN-based threads; and while I tried to create with RDSN it still failed due to dead lock, thus I turn to current solution.

acelyc111 (approved):

foreverneverer (approved):

PR #2085, opened 2024-08-01
feat(new_metrics): support server_stat command showing some important server-level metrics (part 1)

empiredan:
As the 1st part that support server_stat command, both built-in metrics, the usage
of virtual and physical memory would be shown.


src/shell/command_helper.h
```diff
@@ -709,30 +709,68 @@ inline std::vector<dsn::http_result> get_metrics(const std::vector<node_desc> &n
     return results;
 }
 
+// Adapt the result returned by `get_metrics` into the structure that could be processed by
+// `remote_command`.
+template <typename... Args>
+inline std::pair<bool, std::string> process_get_metrics_result(const dsn::http_result &result,
```
acelyc111, line 715: It seems the return type can be replaced by `error_s` or `absl::Status`.

src/shell/main.cpp
```diff
@@ -101,7 +101,7 @@ static command_executor commands[] = {
         "get the node status for this cluster",
         "[-d|--detailed] [-j|--json] [-r|--resolve_ip] [-u|--resource_usage] "
         "[-o|--output file_name] [-s|--status all|alive|unalive] [-q|--qps] "
-        "[-t|--sample_interval_ms num]",
+        "[-i|--sample_interval_ms num]",
```
acelyc111, line 104: It should be mentioned in "Behavior changes" in the PR description for this PR.

src/shell/commands/node_management.cpp
```diff
@@ -232,6 +234,160 @@ dsn::metric_filters rw_requests_filters()
...
+    if (!command_result.first) {
+        // Metrics failed to be aggregated.
+        return command_result;
+    }
+
+    return std::make_pair(true,
+                          dsn::json::json_forwarder<meta_server_stats>::encode(stats).to_string());
+}
+
+struct replica_server_stats
```
acelyc111, line 278: It's the same to `meta_server_stats`, how about using a unify one?

src/shell/commands/node_management.cpp
```diff
@@ -232,6 +234,160 @@ dsn::metric_filters rw_requests_filters()
     return filters;
 }
 
+dsn::metric_filters server_stat_filters()
+{
+    dsn::metric_filters filters;
+    filters.with_metric_fields = {dsn::kMetricNameField, dsn::kMetricSingleValueField};
+    filters.entity_types = {"server"};
+    filters.entity_metrics = {"virtual_mem_usage_mb", "resident_mem_usage_mb"};
```
acelyc111, line 242: What about defining some static const strings for these metric names?

src/shell/command_helper.h
empiredan: OK, use `error_s` instead.

src/shell/commands/node_management.cpp
empiredan: Actually they are different, since more metrics would be introduced later.

src/shell/commands/node_management.cpp
empiredan: These names of metrics have been defined, initialized and set as following statements, thus I think we could just keep them in the format of strings.

```C++
METRIC_DEFINE_gauge_int64(server,
                          virtual_mem_usage_mb,
                          dsn::metric_unit::kMegaBytes,
                          "The total amount of virtual memory usage in MB");

METRIC_VAR_INIT_server(virtual_mem_usage_mb)

METRIC_VAR_SET(virtual_mem_usage_mb, virt_mb);
```

src/shell/main.cpp
empiredan: OK, the options `sample_interval_ms` was introduced recently and never released before.

acelyc111 (approved):

GehaFearless (approved):

PR #2088, opened 2024-08-03
fix(duplication): fix the dangling pointer after `blob` object is moved

empiredan:
https://github.com/apache/incubator-pegasus/issues/2089

Refactor `blob` object to avoid the dangling pointer leading to heap-use-after-free
error which happened in #2089 after `blob` object is moved. Also refactor and add
some tests for `blob` and `mutation_batch`.


acelyc111:
Good work! Could you please rebase the latest master branch to check if there are any clang-tidy warnings or errors.

You don't have to fix all the warnings/errors, some rules can be disabled after discussion.

acelyc111 (approved):

acelyc111 (approved):

foreverneverer (approved):

Issue #2092, opened 2024-08-08
Checking that a remote table with only one replica is ready for duplication after copying checkpoints failed for ERR_NOT_ENOUGH_MEMBER


empiredan:
After copying checkpoints for full duplication, checking a remote table whose replication factor is 1 failed due to `ERR_NOT_ENOUGH_MEMBER`. The error log of meta server is as follows:

```
E2024-08-07 18:52:42.537 (1723027962537681375 5005)   meta.meta_state0.0103000000000109: meta_duplication_service.cpp:577:operator()(): query follower app[target_cluster.dup_test_3] replica configuration completed, result: duplication_status = DS_APP, query_err = ERR_NOT_ENOUGH_MEMBER, update_err = ERR_NO_NEED_OPERATE
```

empiredan:
This issue has been fixed by https://github.com/apache/incubator-pegasus/pull/2093.

PR #2093, opened 2024-08-08
fix(duplication): fix unexpected ERR_NOT_ENOUGH_MEMBER while checking if the remote table with only one replica is ready for duplication

empiredan:
https://github.com/apache/incubator-pegasus/issues/2092

While the remote table created with only one replica for the full duplication
(copying checkpoints), the duplication would get stuck in `DS_APP` status
forever due to `ERR_NOT_ENOUGH_MEMBER`. The reason is that the check
would fail once there is no secondary replica. However, for single-replica tables,
there is certainly not any secondary replica.

### DIFF
--- a/src/meta/duplication/meta_duplication_service.cpp
+++ b/src/meta/duplication/meta_duplication_service.cpp
@@ -15,13 +15,16 @@
 // specific language governing permissions and limitations
 // under the License.
 
+// IWYU pragma: no_include <ext/alloc_traits.h>
 #include <fmt/core.h>
+#include <stddef.h>
 #include <algorithm>
 #include <cstdint>
+#include <functional>
 #include <queue>
+#include <string_view>
 #include <type_traits>
 
-#include <string_view>
 #include "common//duplication_common.h"
 #include "common/common.h"
 #include "common/gpid.h"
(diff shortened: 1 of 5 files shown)


src/meta/duplication/meta_duplication_service.cpp
```diff
@@ -535,79 +599,78 @@ void meta_duplication_service::check_follower_app_if_create_completed(
...
+                "create_app_ok",
+                std::bind(
+                    mock_create_app, std::placeholders::_1, dup, std::ref(resp), std::ref(err)));
+
+            // - ERR_INCONSISTENT_STATE: partition count of response isn't equal with local
+            // - ERR_INACTIVE_STATE: the follower table hasn't been healthy
+            error_code query_err = err == ERR_OK ? resp.err : err;
+            if (query_err == ERR_OK) {
+                if (resp.partitions.size() != dup->partition_count) {
+                    query_err = ERR_INCONSISTENT_STATE;
```
acelyc111, line 617: How about `continue` here to reduce the depth of the following code?

acelyc111 (approved):

foreverneverer (approved):

PR #2099, opened 2024-08-16
feat(duplication): add a new option to configure the duration of the delay until the next execution if there is no mutation to be loaded

empiredan:
A new option is added for duplication to configure the duration of the delay
until the next execution if there is no mutation to be loaded:

```diff
[replication]
+ dup_no_mutation_load_delay_ms = 100
```

### DIFF
--- a/src/replica/duplication/duplication_pipeline.cpp
+++ b/src/replica/duplication/duplication_pipeline.cpp
@@ -74,7 +74,7 @@ void load_mutation::run()
         std::min(_replica->private_log()->max_decree_on_disk(), _replica->last_applied_decree());
     if (_start_decree > max_plog_committed_decree) {
         // Wait for a while if no mutation was added.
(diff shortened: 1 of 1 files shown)


src/replica/duplication/duplication_pipeline.cpp
```diff
@@ -17,22 +17,29 @@
...
 #include "utils/autoref_ptr.h"
 #include "utils/errors.h"
+#include "utils/flags.h"
 #include "utils/fmt_logging.h"
 
+DSN_DEFINE_uint64(
+    replication,
+    dup_no_mutation_load_delay_ms,
+    100,
+    "The duration of the delay until the next execution if there is no mutation to be loaded.");
```
acelyc111, line 41: Mark it as FT_MUTABLE then we can modify it at runtime.

acelyc111 (approved):

foreverneverer (approved):

Issue #2100, opened 2024-08-19
Bulkload can not work during duplication


ninsmiracle:
## Bug Report

Please answer these questions before submitting your issue. Thanks!
There are a lot of Pegasus user need to do bulkload, however if those apps is doing duplication, clusters may coredump all.

1. What did you do?
- use this app
- add dup mession
- doing bulkload

4. What version of Pegasus are you using?
2.4

PR #2101, opened 2024-08-19
fix(duplication): deal with bulkload dup crash

ninsmiracle:
#2100 

Make replica server skip `RPC_RRDB_RRDB_BULK_LOAD` code when reply the plog.

##### Tests 
- Unit test
- Manual test

### DIFF
--- a/src/server/pegasus_mutation_duplicator.cpp
+++ b/src/server/pegasus_mutation_duplicator.cpp
@@ -239,6 +239,10 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
     auto batch_request = std::make_unique<dsn::apps::duplicate_request>();
     uint batch_count = 0;
     uint batch_bytes = 0;
+    // The rpc codes should be ignored:
+    // - RPC_RRDB_RRDB_DUPLICATE: Now not supports duplicating the deuplicate mutations to the
+    // remote cluster.
+    // - RPC_RRDB_RRDB_BULK_LOAD: Now not supports the control flow RPC.
     const static std::set<int> ingnored_rpc_code = {dsn::apps::RPC_RRDB_RRDB_DUPLICATE,
                                                     dsn::apps::RPC_RRDB_RRDB_BULK_LOAD};
 
@@ -249,9 +253,6 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
         dsn::blob raw_message = std::get<2>(mut);
         auto dreq = std::make_unique<dsn::apps::duplicate_request>();
 
-        // ignore if it is a DUPLICATE or BULKLOAD
-        // Because DUPLICATE comes from other clusters should not be forwarded to any other
-        // destinations. A DUPLICATE is meant to be targeting only one cluster.
         if (gutil::ContainsKey(ingnored_rpc_code, rpc_code)) {
             // It it do not recommend to use bulkload and normal writing in the same app,
             // it may also cause inconsistency between actual data and expected data
--- a/src/server/pegasus_mutation_duplicator.cpp
+++ b/src/server/pegasus_mutation_duplicator.cpp
@@ -35,6 +35,7 @@
 #include "common/common.h"
 #include "common/duplication_common.h"
 #include "duplication_internal_types.h"
+#include "gutil/map_util.h"
 #include "pegasus/client.h"
 #include "pegasus_key_schema.h"
 #include "rpc/rpc_message.h"
@@ -238,6 +239,9 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
     auto batch_request = std::make_unique<dsn::apps::duplicate_request>();
     uint batch_count = 0;
     uint batch_bytes = 0;
+    const static std::set<int> ingnored_rpc_code = {dsn::apps::RPC_RRDB_RRDB_DUPLICATE,
+                                                    dsn::apps::RPC_RRDB_RRDB_BULK_LOAD};
+
     for (auto mut : muts) {
         // mut: 0=timestamp, 1=rpc_code, 2=raw_message
         batch_count++;
@@ -248,7 +252,7 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
         // ignore if it is a DUPLICATE or BULKLOAD
         // Because DUPLICATE comes from other clusters should not be forwarded to any other
         // destinations. A DUPLICATE is meant to be targeting only one cluster.
-        if (ingnored_rpc_code.find(rpc_code) != ingnored_rpc_code.end()) {
+        if (gutil::ContainsKey(ingnored_rpc_code, rpc_code)) {
             // It it do not recommend to use bulkload and normal writing in the same app,
             // it may also cause inconsistency between actual data and expected data
             // And duplication will not dup the data of bulkload to backup clusters,
--- a/src/server/pegasus_mutation_duplicator.cpp
+++ b/src/server/pegasus_mutation_duplicator.cpp
@@ -264,15 +264,15 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
                 LOG_DEBUG_PREFIX("Ignore sending bulkload rpc when doing duplication");
             }
             continue;
-        } else {
-            dsn::apps::duplicate_entry entry;
-            entry.__set_raw_message(raw_message);
-            entry.__set_task_code(rpc_code);
-            entry.__set_timestamp(std::get<0>(mut));
-            entry.__set_cluster_id(dsn::replication::get_current_dup_cluster_id());
-            batch_request->entries.emplace_back(std::move(entry));
-            batch_bytes += raw_message.length();
         }
+
+        dsn::apps::duplicate_entry entry;
+        entry.__set_raw_message(raw_message);
+        entry.__set_task_code(rpc_code);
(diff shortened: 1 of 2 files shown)


src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -250,6 +250,9 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
             // Because DUPLICATE comes from other clusters should not be forwarded to any other
             // destinations. A DUPLICATE is meant to be targeting only one cluster.
             continue;
+        } else if (rpc_code == dsn::apps::RPC_RRDB_RRDB_BULK_LOAD) {
```
acelyc111, line 253: It's not a problem to add a judgement here like this, but how to deal with the ingested data? As a "master cluster", should the data been ingested to be duplicate to remote cluster?

ninsmiracle: Currently, there are many users that require both bulkload and duplication.
So the current strategy is that users have to ensure that a bulkload is also performed on the bakcup-cluster.
We will have an external component that can detect the bulkload usage of a cluster's table. If the master-cluster does bulkload but the backup-cluster does not, an email notification will be sent.

ninsmiracle: Last week our cluster crashed for this reason. 
A user did a bulkload on a table with dup enabled, which caused the whole cluster to coredump. We think this is very unsafe.

ninsmiracle:
Without this fix, cluster will core like this (if users doing bulkload on a table with dup enable)
```
(gdb) bt
#0  0x0000000000000001 in ?? ()
#1  0x00007f8c97f3698d in ?? () from /lib64/libgcc_s.so.1
#2  0x00007f8c97f36ded in _Unwind_Resume () from /lib64/libgcc_s.so.1
#3  0x00007f8c9cb5953e in ~_Function_base (this=<optimized out>, __in_chrg=<optimized out>) at /opt/rh/devtoolset-7/root/usr/include/c++/7/bits/std_function.h:276
#4  ~function (this=<optimized out>, __in_chrg=<optimized out>) at /opt/rh/devtoolset-7/root/usr/include/c++/7/bits/std_function.h:389
#5  dsn::replication::ship_mutation::ship(std::set<std::tuple<unsigned long, dsn::task_code, dsn::blob>, dsn::replication::mutation_tuple_cmp, std::allocator<std::tuple<unsigned long, dsn::task_code, dsn::blob> > >&&) (this=this@entry=0x2d110f950, in=...) at /home/work/test/pegasus/src/rdsn/src/replica/duplication/duplication_pipeline.cpp:71
#6  0x00007f8c9cb59b27 in dsn::replication::ship_mutation::run(long&&, std::set<std::tuple<unsigned long, dsn::task_code, dsn::blob>, dsn::replication::mutation_tuple_cmp, std::allocator<std::tuple<unsigned long, dsn::task_code, dsn::blob> > >&&) (this=0x2d110f950, last_decree=<optimized out>, in=...) at /home/work/test/pegasus/src/rdsn/src/replica/duplication/duplication_pipeline.cpp:88
#7  0x00007f8c9ccc1f11 in dsn::task::exec_internal (this=this@entry=0xb2df000f0) at /home/work/test/pegasus/src/rdsn/src/runtime/task/task.cpp:176
#8  0x00007f8c9ccd75c2 in dsn::task_worker::loop (this=0x2584d10) at /home/work/test/pegasus/src/rdsn/src/runtime/task/task_worker.cpp:224
#9  0x00007f8c9ccd7740 in dsn::task_worker::run_internal (this=0x2584d10) at /home/work/test/pegasus/src/rdsn/src/runtime/task/task_worker.cpp:204
#10 0x00007f8c9b953a1f in execute_native_thread_routine () from /home/work/app/pegasus/alsgsrv-monetization-master/replica/package/bin/libdsn_utils.so
#11 0x00007f8c9975edc5 in start_thread () from /lib64/libpthread.so.0
#12 0x00007f8c97c5d73d in clone () from /lib64/libc.so.6
(gdb) 
```

And stdout like this:
```
F2024-08-08 16:26:36.249 (1723105596249247134 67179) replica.replica20.0406000900422908: pegasus_mutation_duplicator.cpp:85:get_hash_from_request(): unexpected task code: RPC_RRDB_RRDB_BULK_LOAD
```

src/server/pegasus_mutation_duplicator.cpp
acelyc111: When bulkload on source and target clusters, will you stop client's normal writing? How to keep the consistent of the 2 clusters if not?

ninsmiracle: The current situation will not stop normal writing. And it may make some unconsistent data between 2 clusters.
In general, we will inform users that this is a situation that will lead to data inconsistency between master-cluster and backup-cluster.

ninsmiracle: In other words, if there is only one cluster, if there is both nomral writing and bulkload (without ingest_behind parameter), it may also cause inconsistency between actual data and expected data.

acelyc111: So it's better to deny bulkloading on a duplicating table (you can permit this risk in you own way).

This patch's change looks good to me.

acelyc111, line 253: How about defining a set which contains RPC_RRDB_RRDB_DUPLICATE and RPC_RRDB_RRDB_BULK_LOAD? Please add some more comments as well.

src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -250,6 +250,9 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
...
+            LOG_DEBUG_PREFIX("Ignore sending bulkload rpc when doing duplication");
+            continue;
         } else {
             dsn::apps::duplicate_entry entry;
```
acelyc111, line 257: Now that all the above branches have been continued, you can remove the `} else {` to reduce the indents.

src/server/pegasus_mutation_duplicator.h
```diff
@@ -97,5 +98,8 @@ class pegasus_mutation_duplicator : public dsn::replication::mutation_duplicator
 // calculates the hash value from the write's hash key.
 extern uint64_t get_hash_from_request(dsn::task_code rpc_code, const dsn::blob &request_data);
 
+const std::set<int> ingnored_rpc_code = {dsn::apps::RPC_RRDB_RRDB_DUPLICATE,
```
acelyc111, line 101: Nobody else use it, you can move it to the cpp file, define it as a local static variable in pegasus_mutation_duplicator::duplicate().

src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -245,10 +245,20 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
...
         auto dreq = std::make_unique<dsn::apps::duplicate_request>();
 
-        if (rpc_code == dsn::apps::RPC_RRDB_RRDB_DUPLICATE) {
-            // ignore if it is a DUPLICATE
-            // Because DUPLICATE comes from other clusters should not be forwarded to any other
-            // destinations. A DUPLICATE is meant to be targeting only one cluster.
+        // ignore if it is a DUPLICATE or BULKLOAD
+        // Because DUPLICATE comes from other clusters should not be forwarded to any other
+        // destinations. A DUPLICATE is meant to be targeting only one cluster.
+        if (ingnored_rpc_code.find(rpc_code) != ingnored_rpc_code.end()) {
```
acelyc111, line 251: Use Contains*() in src/gutil/map_util.h to simplify the code.

src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -238,28 +239,41 @@ void pegasus_mutation_duplicator::duplicate(mutation_tuple_set muts, callback cb
...
         dsn::blob raw_message = std::get<2>(mut);
         auto dreq = std::make_unique<dsn::apps::duplicate_request>();
 
-        if (rpc_code == dsn::apps::RPC_RRDB_RRDB_DUPLICATE) {
-            // ignore if it is a DUPLICATE
-            // Because DUPLICATE comes from other clusters should not be forwarded to any other
-            // destinations. A DUPLICATE is meant to be targeting only one cluster.
+        // ignore if it is a DUPLICATE or BULKLOAD
+        // Because DUPLICATE comes from other clusters should not be forwarded to any other
+        // destinations. A DUPLICATE is meant to be targeting only one cluster.
```
acelyc111, line 254:
```suggestion
        // The rpc codes should be ignored:
        // - RPC_RRDB_RRDB_DUPLICATE: Now not supports duplicating the deuplicate mutations to the remote cluster.
        // - RPC_RRDB_RRDB_BULK_LOAD: Now not supports the control flow RPC.
```
And move these before line 242.

ninsmiracle:
@acelyc111  Shoud I fix `.github/workflows/module_labeler_conf.yml` to deal with CI stage Module Labeler?

acelyc111:
> @acelyc111 Shoud I fix `.github/workflows/module_labeler_conf.yml` to deal with CI stage Module Labeler?

Will. Just ignore it now.

acelyc111 (approved):
Thanks! @ninsmiracle

acelyc111:
@ninsmiracle Please rebase the master branch and fix the clang-tidy errors.

acelyc111 (approved):

empiredan (approved):

Issue #2109, opened 2024-09-04
replica server with security enabled exited abnormally due to `blob::create_from_bytes` assertion


empiredan:
Replica server was started with security enabled as following configurations:

```ini
[security]
enable_auth = true
krb5_keytab = ...
krb5_config = /etc/krb5.conf
krb5_principal = ...
sasl_plugin_path = /usr/lib64/sasl2
service_fqdn = ...
service_name = ...
mandatory_auth = true
enable_acl = true
super_users =
meta_acl_rpc_allow_list =
enable_ranger_acl = true
```

However, it failed immediately after started with following error logs and generate core dump file:

```bash
E2024-09-04 12:06:36.870 (1725422796870535314 3e18) replica.default2.0201000500000008: compaction_filter_rule.cpp:109:create_from_bytes(): assertion expression: (s) != nullptr
F2024-09-04 12:06:36.870 (1725422796870576704 3e18) replica.default2.0201000500000008: compaction_filter_rule.cpp:109:create_from_bytes(): null source pointer would lead to undefined behaviour
```

The backtrace of core dump file is:

```bash
#0  0x00007f6375ec7387 in raise () from /lib64/libc.so.6
#1  0x00007f6375ec8a78 in abort () from /lib64/libc.so.6
#2  0x00007f637ae8b716 in dsn_coredump () at /data/code/pegasus-2.4/src/rdsn/src/runtime/service_api_c.cpp:95
#3  0x00007f6379af0e39 in dsn::tools::(anonymous namespace)::process_fatal_log (log_level=LOG_LEVEL_FATAL) at /data/code/pegasus-2.4/src/rdsn/src/utils/simple_logger.cpp:91
#4  0x00007f6379af2293 in dsn::tools::simple_logger::dsn_log (this=0x2ad8e70, file=0xcb99e5 "compaction_filter_rule.cpp",
    function=0xcbc730 <dsn::blob::create_from_bytes(char const*, unsigned long)::__FUNCTION__> "create_from_bytes", line=109, log_level=LOG_LEVEL_FATAL,
    str=0x6e667e8 "null source pointer would lead to undefined behaviour") at /data/code/pegasus-2.4/src/rdsn/src/utils/simple_logger.cpp:334
#5  0x00007f6379abf2df in dsn_log (file=0xcb99e5 "compaction_filter_rule.cpp",
[...]

PR #2110, opened 2024-09-04
fix(blob): allow `blob` object to be initialized with null-pointer char array whose length must be zero

empiredan:
Fix https://github.com/apache/incubator-pegasus/issues/2109.

Null-pointer char array with zero length is allowed to be used to initialize
`blob` object. Use `DCHECK` instead to check if the pointer is valid while
initializing `blob` object.

### DIFF
--- a/src/utils/blob.h
+++ b/src/utils/blob.h
@@ -109,7 +109,13 @@ public:
     /// NOTE: this operation is not efficient since it involves a memory copy.
     [[nodiscard]] static blob create_from_bytes(const char *s, size_t len)
     {
-        CHECK_NOTNULL(s, "null source pointer would lead to undefined behaviour");
+        if (s == nullptr) {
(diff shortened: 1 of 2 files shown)


src/utils/blob.h
```diff
@@ -109,7 +109,15 @@ class blob
     /// NOTE: this operation is not efficient since it involves a memory copy.
     [[nodiscard]] static blob create_from_bytes(const char *s, size_t len)
     {
-        CHECK_NOTNULL(s, "null source pointer would lead to undefined behaviour");
+        if (s == nullptr) {
```
acelyc111, line 112: If this happens in rare cases, wrap it by dsn_unlikely.

empiredan: OK, I dropped this and use `DCHECK` instead.

acelyc111 (approved):

foreverneverer (approved):

PR #2137, opened 2024-10-21
fix(github): remove centos7 on git action

### LINKED ISSUE #2135: Workflows failed due to EOL of CentOS 7
[BuildCompilationEnvDocker](https://github.com/apache/incubator-pegasus/actions/workflows/build-push-env-docker.yml) failed due to [EOL of CentOS 7](https://www.redhat.com/en/blog/centos-linux-has-reached-its-end-life-eol):

image

shalk:

fix #2135

##### Tests 

- Unit test
- Integration test
- No code

##### Code changes

- Has persistent data change

##### Side effects

- Increased code complexity

##### Related changes


acelyc111:
@shalk Thanks for the contribution!

As CentOS 7 is widely used and EOL for only 3 months, it's not good to terminate the supporting right now. If do it like this, some build and runtime issues may occurs, we may got complains
in the community.

shalk:
The CenOS 7 Dockerfile can not work anymore. docker pull will fail.

I think  almalinux or rockylinux is a alternative for centos7.  Should be anouther feature

Issue #1945, opened 2024-03-13
Feature: refactor cold backup and restore function


ninsmiracle:
## Feature Request
Need to continue merge new backup code to current master. And do not do priodic backup in pegasus server side. 

For more detail design:
#1081

PR #2142, opened 2024-10-31
feat(backup): 2. update and refactor meta backup engine class

ninsmiracle:
#1945 

As the issue shows, meta_backup_engine will intertact with replica server and contorl backup lifecycle and status switch. This pr is the first part about it, including:
1.update init_backup and start functions
2.update backup_status and backup_item structures
3.update varieties of meta_backup_engine.h
4.rename backup_engine into meta_backup_engine
5.refactor meta_backup_engine.h including public/private, function definition order etc.


idl/backup.thrift
```diff
@@ -184,16 +184,28 @@ struct start_backup_app_response
...
-    3:string        backup_provider_type;
+    2:i32           app_id;
+    3:string        app_name;
+    4:string        backup_provider_type;
     // user specified backup_path.
-    4:string        backup_path;
-    5:i64           start_time_ms;
-    6:i64           end_time_ms;
-    7:bool          is_backup_failed;
+    5:string        backup_path;
```
acelyc111, line 205: While you change the field sequence, it would cause compatiable issues. How do you resolve it?

It's just the backup policy is incompatiable, however, the backup data is compatiable, that is to say, the backup data of the old implementation can be read normally by the new implemantation, right?

src/meta/meta_backup_engine.cpp
```diff
@@ -48,72 +48,93 @@
...
-                         ->get_block_service_manager()
-                         .get_or_create_block_filesystem(provider);
-    if (_block_service == nullptr) {
-        return ERR_INVALID_PARAMETERS;
+    LOG_DEBUG("App[{}] start {} backup[{}] on {}, root_path = {}",
+              _cur_backup.app_name,
+              _is_periodic_backup ? "periodic" : "onetime",
+              _cur_backup.backup_id,
+              _cur_backup.backup_provider_type,
+              _cur_backup.backup_path);
```
acelyc111, line 98: What about mention the URI of the backup destination?

src/meta/meta_backup_engine.cpp
```diff
@@ -48,72 +48,93 @@
...
+// ThreadPool: THREAD_POOL_DEFAULT
+error_code meta_backup_engine::write_app_info()
 {
-    if (_block_service && _block_service->is_root_path_set()) {
-        return ERR_INVALID_PARAMETERS;
-    }
-    LOG_INFO("backup path is set to {}.", path);
-    _backup_path = path;
+    // TODO(heyuchen): TBD
+    // guoningshen: write app info on block service
```
acelyc111, line 123:
```suggestion
    // TODO(guoningshen): write app info on block service
```

src/meta/meta_backup_engine.h
```diff
@@ -63,45 +56,81 @@ struct app_backup_info
...
 
     int64_t get_current_backup_id() const { return _cur_backup.backup_id; }
     int32_t get_backup_app_id() const { return _cur_backup.app_id; }
-    bool is_in_progress() const;
 
-    backup_item get_backup_item() const;
+    backup_item get_backup_item() const
+    {
+        zauto_read_lock l(_lock);
+        backup_item item = _cur_backup;
```
acelyc111, line 85:
```suggestion
        {
            zauto_read_lock l(_lock);
            backup_item item = _cur_backup;
        }
```

PR #2148, opened 2024-11-19
fix(meta): meta server failed and could not be started normally while setting environment variables after dropping table

empiredan:
https://github.com/apache/incubator-pegasus/issues/2149.

There are two problems that should be solved:

1. why the primary meta server failed with `segfault` while dropping tables ?
2. why all meta servers were never be restarted normally after the primary
meta server failed ?

A pegasus cluster would flush security policies to remote meta storage
periodically (by `update_ranger_policy_interval_sec`) in the form of environment
variables. We do this by `server_state::set_app_envs()`. However, after updating
the meta data on the remote storage (namely ZooKeeper), the table is not checked
that if it still exists while updating environment variables of local memory:

```C++
void server_state::set_app_envs(const app_env_rpc &env_rpc)
{

...

    do_update_app_info(app_path, ainfo, [this, app_name, keys, values, env_rpc](error_code ec) {
        CHECK_EQ_MSG(ec, ERR_OK, "update app info to remote storage failed");

        zauto_write_lock l(_lock);
        std::shared_ptr<app_state> app = get_app(app_name);
        std::string old_envs = dsn::utils::kv_map_to_string(app->envs, ',', '=');
        for (int idx = 0; idx < keys.size(); idx++) {
            app->envs[keys[idx]] = values[idx];
        }
        std::string new_envs = dsn::utils::kv_map_to_string(app->envs, ',', '=');
        LOG_INFO("app envs changed: old_envs = {}, new_envs = {}", old_envs, new_envs);
    });
}
```

In `std::string old_envs = dsn::utils::kv_map_to_string(app->envs, ',', '=');`, since
`app` is `nullptr`, `app->envs` would point an invalid address, leading to `segfault`
in `libdsn_utils.so` where `dsn::utils::kv_map_to_string` is.

Therefore, the reason for the 1st problem is very clear: the callback for updating meta
data on remote storage is called immediately after the table is removed, and an invalid
address is accessed due to null pointer.

Then, the meta server would load meta data from remote storage after it is restart.
However, the intermediate status `AS_DROPPING` is also flushed to remote storage
with security policies since all meta data for a table is an unitary `json` object: the whole
`json` would be set to remote storage once any property is updated.

However `AS_DROPPING` is invalid, and cannot pass the assertion which would make
meta server fail again and again, which is the reason of the 2nd problem: 

```C++
server_state::sync_apps_from_remote_storage()
{

...

                    std::shared_ptr<app_state> app = app_state::create(info);
                    {
                        zauto_write_lock l(_lock);
                        _all_apps.emplace(app->app_id, app);
                        if (app->status == app_status::AS_AVAILABLE) {
                            app->status = app_status::AS_CREATING;
                            _exist_apps.emplace(app->app_name, app);
                            _table_metric_entities.create_entity(app->app_id, app->partition_count);
                        } else if (app->status == app_status::AS_DROPPED) {
                            app->status = app_status::AS_DROPPING;
                        } else {
                            CHECK(false,
                                  "invalid status({}) for app({}) in remote storage",
                                  enum_to_string(app->status),
                                  app->get_logname());
                        }
                    }

...

}
```

To fix the 1st problem, we just check if the table still exists after meta data is updated
on the remote storage. To fix the 2nd problem, we prevent meta data with intermediate
status `AS_DROPPING` from being flushed to remote storage.

### DIFF
--- a/src/meta/server_state.cpp
+++ b/src/meta/server_state.cpp
@@ -1006,7 +1006,7 @@ void server_state::query_configuration_by_index(const query_cfg_request &request
                                                 /*out*/ query_cfg_response &response)
 {
     zauto_read_lock l(_lock);
-    auto iter = _exist_apps.find(request.app_name.c_str());
+    auto iter = _exist_apps.find(request.app_name);
     if (iter == _exist_apps.end()) {
         response.err = ERR_OBJECT_NOT_FOUND;
         return;
@@ -3024,22 +3024,26 @@ void server_state::set_app_envs(const app_env_rpc &env_rpc)
     app_info ainfo;
     std::string app_path;
     {
-        zauto_read_lock l(_lock);
+        FAIL_POINT_INJECT_NOT_RETURN_F("set_app_envs_failed", [app_name, this](std::string_view s) {
+            zauto_write_lock l(_lock);
 
-        auto app = get_app(app_name);
-
-        FAIL_POINT_INJECT_NOT_RETURN_F("set_app_envs_failed", [&app](std::string_view s) {
             if (s == "not_found") {
-                app.reset();
+                CHECK_EQ(_exist_apps.erase(app_name), 1);
                 return;
             }
 
             if (s == "dropping") {
-                app->status = app_status::AS_DROPPING;
+                const auto &iter = _exist_apps.find(app_name);
+                CHECK_TRUE(iter != _exist_apps.end());
+
+                iter->second->status = app_status::AS_DROPPING;
                 return;
             }
         });
 
+        zauto_read_lock l(_lock);
+
+        const auto &app = get_app(app_name);
         if (!app) {
             LOG_WARNING("set app envs failed since app_name({}) cannot be found", app_name);
             env_rpc.response().err = ERR_APP_NOT_EXIST;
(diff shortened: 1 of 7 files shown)


src/meta/server_state.cpp
```diff
@@ -3020,31 +3024,87 @@ void server_state::set_app_envs(const app_env_rpc &env_rpc)
...
+            zauto_write_lock l(_lock);
+
+            if (s == "not_found") {
+                CHECK_EQ(_exist_apps.erase(app_name), 1);
+                return;
+            }
+
+            if (s == "dropping") {
+                const auto &iter = _exist_apps.find(app_name);
+                CHECK_TRUE(iter != _exist_apps.end());
```
acelyc111, line 3037: Use `gutil::FindOrDie()` would simplify the code.

src/meta/server_state.h
```diff
@@ -140,19 +140,26 @@ class server_state
...
-    const meta_view get_meta_view() { return {&_all_apps, &_nodes}; }
+
+    meta_view get_meta_view() { return {&_all_apps, &_nodes}; }
+
     std::shared_ptr<app_state> get_app(const std::string &name) const
     {
-        auto iter = _exist_apps.find(name);
-        if (iter == _exist_apps.end())
-            return nullptr;
+        const auto &iter = _exist_apps.find(name);
```
acelyc111, line 148: It would be simpler if using find* in map-utils.h

acelyc111 (approved):

Samunroyu (approved):

PR #2159, opened 2024-11-28
feat(duplication): collect last committed decree from replica to meta for duplication

empiredan:


src/meta/duplication/duplication_info.cpp
```diff
@@ -20,10 +20,20 @@
...
-namespace dsn {
-namespace replication {
+DSN_DEFINE_uint64(replication,
+                  dup_progress_min_update_period_ms,
+                  5000,
+                  "The minimum period in milliseconds that progress of duplication is updated");
+
+DSN_DEFINE_uint64(replication,
+                  dup_progress_min_report_period_ms,
+                  static_cast<uint64_t>(5) * 60 * 1000,
```
acelyc111, line 33: Why add static cast? It this OK?
```suggestion
                  5ULL * 60 * 1000,
```

empiredan: This is warned by `clang-tidy`, and I think both could eliminate this warning by explicitly converting to unsigned long int:
```
warning: performing an implicit widening conversion to type 'uint64_t' (aka 'unsigned long') of a multiplication performed in type 'int' [bugprone-implicit-widening-of-multiplication-result]
                  5 * 60 * 1000,
```

acelyc111 (approved):

Samunroyu (approved):

PR #2161, opened 2024-12-02
feat(new_metircs): collect the number of primary and secondary replicas

empiredan:
Currently we don't have metrics about how many primary replicas or secondary
replicas on a replica server, which means we have to find them using `nodes`
command by Pegasus shell rather than just monitoring graphs.

### DIFF
--- a/src/replica/replica_stub.cpp
+++ b/src/replica/replica_stub.cpp
@@ -116,6 +116,16 @@ METRIC_DEFINE_gauge_int64(server,
                           "The number of closing replicas");
 
 METRIC_DEFINE_gauge_int64(server,
+                          inactive_replicas,
+                          dsn::metric_unit::kReplicas,
+                          "The number of inactive replicas");
+
+METRIC_DEFINE_gauge_int64(server,
+                          error_replicas,
+                          dsn::metric_unit::kReplicas,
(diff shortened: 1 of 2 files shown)


src/replica/replica_stub.cpp
```diff
@@ -1654,26 +1688,34 @@ void replica_stub::on_replicas_stat()
...
                 std::max(splitting_max_async_learn_time_ms, rep->_split_states.async_learn_ms());
             splitting_max_copy_file_size =
                 std::max(splitting_max_copy_file_size, rep->_split_states.splitting_copy_file_size);
+
+            continue;
         }
     }
 
-    METRIC_VAR_SET(learning_replicas, learning_count);
+    METRIC_VAR_SET(inactive_replicas, status_counts[partition_status::PS_INACTIVE]);
```
acelyc111, line 1708: These METRIC_VAR_SET will be skipped if using `continue` above, is it in purpose?

empiredan: `continue` is in `for` loop, and `METRIC_VAR_SET` is out of `for` loop, right ?

acelyc111 (approved):

foreverneverer (approved):

Issue #2160, opened 2024-12-02
refactor: support close admin-cli client


ruojieranyishen:
## Feature Request
**Is your feature request related to a problem? Please describe:**
I found that the current admin CLI client does not provide an interface to close replica server TCP connections.  And the `QueryAllNodesDiskInfo`  function is difficult to use.  `sendQueryDiskInfoRequest` function is private.

**Describe the feature you'd like:**
1. Support close admin-cli client.
2. Refactor `QueryAllNodesDiskInfo` function.

PR #2162, opened 2024-12-03
refactor: support close admin-cli client

ruojieranyishen:
https://github.com/apache/incubator-pegasus/issues/2160

Current admin-cli client does not provide an interface to close replica server TCP connections.
And `QueryAllNodesDiskInfo` function is difficult to use. `SendQueryDiskInfoRequest` function is private.

support close client, disconnect tcp link with replica server.

##### Tests 
The TCP connection to the replica server cannot be closed.
image

After support close client，the persistent TCP connection between admincli and the replica server can be disconnected.
image

### DIFF
--- a/admin-cli/executor/client.go
+++ b/admin-cli/executor/client.go
@@ -73,11 +73,11 @@ func NewClient(writer io.Writer, metaAddrs []string, willExit bool) (*Client, er
 	}, nil
 }
 
-func CloseClient(writer io.Writer, client *Client) error {
+func CloseClient(client *Client) error {
 	var errorStrings []string
 	err := client.Meta.Close()
 	if err != nil {
-		fmt.Fprintf(writer, "fatal: failed to close meta session [%s]\n", err)
+		fmt.Printf("Error: failed to close meta session [%s]\n", err)
 		errorStrings = append(errorStrings, err.Error())
 	}
 
@@ -85,7 +85,7 @@ func CloseClient(writer io.Writer, client *Client) error {
 
 	err = client.Nodes.CloseAllNodes()
 	if err != nil {
-		fmt.Fprintf(writer, "fatal: failed to close nodes session [%s]\n", err)
+		fmt.Printf("Error: failed to close nodes session [%s]\n", err)
 		errorStrings = append(errorStrings, err.Error())
 	}
 
--- a/admin-cli/executor/client.go
+++ b/admin-cli/executor/client.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/apache/incubator-pegasus/admin-cli/client"
 	"github.com/apache/incubator-pegasus/admin-cli/util"
@@ -45,6 +46,7 @@ type Client struct {
 }
 
 // NewClient creates a client for accessing Pegasus cluster for use of admin-cli.
+// This function will call os.Exit.
 func NewClient(writer io.Writer, metaAddrs []string) *Client {
 	meta := client.NewRPCBasedMeta(metaAddrs)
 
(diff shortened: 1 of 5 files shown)


admin-cli/executor/client.go
```diff
@@ -67,3 +69,46 @@ func NewClient(writer io.Writer, metaAddrs []string) *Client {
 		Perf:   aggregate.NewPerfClient(metaAddrs),
 	}
 }
+
+// NewClientWithoutExit creates a client for accessing Pegasus cluster for use of admin-cli.
+// This function will not call os.Exit.
+func NewClientWithoutExit(writer io.Writer, metaAddrs []string) (*Client, error) {
```
acelyc111, line 75: This fucntion is similar to the function above, how about unify them?

admin-cli/executor/client.go
```diff
@@ -67,3 +69,46 @@ func NewClient(writer io.Writer, metaAddrs []string) *Client {
...
+	meta := client.NewRPCBasedMeta(metaAddrs)
+
+	nodes, err := meta.ListNodes()
+	if err != nil {
+		fmt.Fprintf(writer, "fatal: failed to list nodes [%s]\n", err)
```
acelyc111, line 80: The "fatal" level log is conventional considered as unacceptable condition, and will be  used before exit the process. It's better to use error or warning level here.

admin-cli/executor/disk_info.go
```diff
@@ -97,6 +99,31 @@ func QueryAllNodesDiskInfo(client *Client, tableName string) (map[string]*radmin
 	return respMap, nil
 }
 
+func QueryAliveNodesDiskInfo(client *Client, tableName string) (map[string]*radmin.QueryDiskInfoResponse, error) {
```
acelyc111, line 102: Is it related to the subject of this patch?

admin-cli/executor/client.go
```diff
@@ -67,3 +69,46 @@ func NewClient(writer io.Writer, metaAddrs []string) *Client {
...
+
+	return &Client{
+		Writer: writer,
+		Meta:   meta,
+		Nodes:  util.NewPegasusNodeManager(metaAddrs, replicaAddrs),
+		Perf:   aggregate.NewPerfClient(metaAddrs),
+	}, nil
+}
+
+func CloseClient(writer io.Writer, client *Client) error {
```
acelyc111, line 97: Where are the functions used?

admin-cli/executor/client.go
ruojieranyishen: Thinks, i will unify them.

admin-cli/executor/disk_info.go
ruojieranyishen: The original QueryNodesDiskInfo was not user-friendly and couldn't handle certain situations, such as missing table partition on nodes or some nodes being unalive.

admin-cli/executor/client.go
ruojieranyishen: The context is that I am estimating the required disk space for a bulk load on the Pegasus gateway using admincli, but currently, admincli does not support disconnect tcp from the replica server.

admin-cli/executor/disk_info.go
acelyc111: Yes, submit this in another PR would be better. Keep each commit independent.

ruojieranyishen: Thinks, I have removed the function.

acelyc111 (approved):

admin-cli/executor/client.go
```diff
@@ -45,14 +46,18 @@ type Client struct {
 }
 
 // NewClient creates a client for accessing Pegasus cluster for use of admin-cli.
-func NewClient(writer io.Writer, metaAddrs []string) *Client {
+// When listing nodes fails, willExit == true means call os.Exit().
+func NewClient(writer io.Writer, metaAddrs []string, willExit bool) (*Client, error) {
```
empiredan, line 50: When `willExit` would be set `false` ? Is it necessary to add the param ?

And returned error is also be ignored, is it needed ?

ruojieranyishen: The context is that I am estimating the required disk space for a bulk load on the Pegasus gateway using admincli, but currently, admincli does not support disconnect tcp from the replica server. 
When using admincli to connect to the Pegasus cluster, it should not call the os.Exit function; it should only return an error.

acelyc111 (approved):

admin-cli/executor/client.go
```diff
@@ -65,5 +70,24 @@ func NewClient(writer io.Writer, metaAddrs []string) *Client {
...
+
+	client.Perf.Close()
+
+	err = client.Nodes.CloseAllNodes()
+	if err != nil {
+		fmt.Printf("Error: failed to close nodes session [%s]\n", err)
+		errorStrings = append(errorStrings, err.Error())
 	}
+
+	return fmt.Errorf("%s", strings.Join(errorStrings, "\n"))
```
empiredan, line 92: Just return `nil` if there is not any error.

admin-cli/executor/client.go
```diff
@@ -65,5 +70,24 @@ func NewClient(writer io.Writer, metaAddrs []string) *Client {
...
 		Perf:   aggregate.NewPerfClient(metaAddrs),
+	}, nil
+}
+
+func CloseClient(client *Client) error {
+	var errorStrings []string
+	err := client.Meta.Close()
+	if err != nil {
+		fmt.Printf("Error: failed to close meta session [%s]\n", err)
+		errorStrings = append(errorStrings, err.Error())
```
empiredan, line 81: What are the possible errors occurring here ? I think they could just be ignored so we would continue to close other components as belows ?

admin-cli/executor/client.go
ruojieranyishen: Ok.

admin-cli/executor/client.go
ruojieranyishen: Thinks, I agree and modify code.

empiredan (approved):

acelyc111 (approved):

PR #2164, opened 2024-12-04
feat(duplication): add a new RPC to list duplications for one or multiple tables with specified pattern

empiredan:
Currently the only query API is used for just one table. However, sometimes
we need to query duplications for more than one tables by just a name pattern
(prefix, postfix, regex, etc.). Even we just want to get all of the duplications of
all tables of the cluster, without giving a specific table name. For example, while
migrating all tables to another cluster, we want to query the progress of all the
tables at any time; or, we just care about the tables of one database(with the
prefix of table name as the database name).

Therefore, a new RPC is added to allow listing duplication info of tables whose
names are given by some pattern, such as prefix, postfix, regex(would be
supported in the future), exactly match or even all tables of the cluster.


src/utils/strings.h
```diff
@@ -62,6 +73,10 @@ bool iequals(const char *lhs, const std::string &rhs, size_t n);
 // Decide whether the first n bytes of two memory areas are equal, even if one of them is NULL.
 bool mequals(const void *lhs, const void *rhs, size_t n);
 
+error_code pattern_match(const std::string &str,
```
acelyc111, line 76: There is a similar implementation in `bool pegasus_server_impl::validate_filter(...)`, is it possible to unify to use it?

empiredan: OK, seems similar. I've added a `TODO` comment for this. And corner cases would also be checked later in case some errors occurred.

acelyc111 (approved):

foreverneverer (approved):

PR #2165, opened 2024-12-10
feat(duplication): add a new command to shell to list duplications for one or multiple tables

empiredan:
Based on https://github.com/apache/incubator-pegasus/pull/2164, this PR is to
introduce a new command for Pegasus shell to list the duplications for one or
multiple tables of this cluster according to provided table name pattern.

This command could list table-level, duplication-level and partition-level info for each
duplication. It could also provide summary stats for listed duplications. It could check
the progress, which is useful while we want to know if the duplication-based migrations
have been finished, or how many decrees have not been duplicated. The duplications
that have not been finished could also be listed.


src/shell/command_helper.h
```diff
@@ -1034,13 +1034,27 @@ class aggregate_stats_calcs
...
     } while (false)
 
+// Parse enum value from the parameters of command line.
+#define PARSE_OPT_ENUM(enum_val, invalid_val, ...)                                                 \
+    do {                                                                                           \
+        const std::string __str(cmd(__VA_ARGS__, "").str());                                       \
+        if (!__str.empty()) {                                                                      \
+            const auto &__val = enum_from_string(__str.c_str(), invalid_val);                      \
+            if (__val == invalid_val) {                                                            \
+                SHELL_PRINTLN_ERROR("invalid enum {}", __str);                                     \
```
acelyc111, line 1051:
```suggestion
                SHELL_PRINTLN_ERROR("invalid enum: '{}'", __str);                                     \
```

acelyc111 (approved):

acelyc111 (approved):

Samunroyu (approved):

PR #2169, opened 2024-12-13
feat: introduce a new task code for replicas stat

empiredan:
Currently calculating stats for replicas is still controlled by a configuration of
`gc_disabled` and based on the  task code of  "garbage collect", which is already
the legacy for GC of shared logs. We could introduce individual controller and
task code for the replicas stats.

Configurations are changed as follows:

```diff
[replication]
+ replicas_stat_disabled = false
- gc_disabled = false
```


acelyc111 (approved):
Alright, but it's still some kind of breaking change, the config key has been changed, we should mention it and show it in release note.

empiredan:
> Alright, but it's still some kind of breaking change, the config key has been changed, we should mention it and show it in release note.

Exactly, I'll mention it both in `Behavior changes` and `Configuration Changes` for the new release.

Samunroyu (approved):

Issue #2149, opened 2024-11-21
Meta server with authentication enabled failed and could not be started normally after dropping table


empiredan:
There were a pegasus cluster with 3 meta servers and 5 replica servers. The authentication was enabled. And a script is written to drop a great number of tables.

While the script was being executed, the meta server failed with nothing but `got signal id: 11` and following `dmesg`:

```
[Tue Nov 12 15:32:39 2024]  meta.meta_stat[681978]: segfault at 40 ip 00007faa351ea839 sp 00007faa0d48abc0 error 4 in libdsn_utils.so[7faa35124000+115000]
[Tue Nov 12 15:32:39 2024] Code: 23 f9 ff 0f 1f 40 00 66 2e 0f 1f 84 00 00 00 00 00 55 48 89 e5 41 57 41 56 41 55 41 54 48 8d 45 cf 53 4c 8d 67 08 48 83 ec 28 <4c> 8b 7f 18 48 89 45 b8 48 8d 45 ce 4d 39 e7 48 89 45 b0 74 6c 48
```

In the logs of failed meta server (namely primary meta server), lots of errors are also found: 

```
E2024-11-12 15:32:45.711 (1731396765711870108 a67f5)   meta.meta_server4.01010000000009fa: ranger_resource_policy_manager.cpp:641:sync_policies_to_app_envs(): ERR_INVALID_PARAMETERS: set_app_envs failed.
E2024-11-12 15:32:45.713 (1731396765713890084 a67f5)   meta.meta_server4.01010000000009fa: ranger_resource_policy_manager.cpp:304:update_policies_from_ranger_service(): ERR_INVALID_PARAMETERS: Sync policies to app envs failed.
E2024-11-12 15:32:52.225 (1731396772225348205 a67f6)   meta.meta_server5.01010000000009fa: ranger_resource_policy_manager.cpp:641:sync_policies_to_app_envs(): ERR_INVALID_PARAMETERS: set_app_envs failed.
E2024-11-12 15:32:52.226 (1731396772226529887 a67f6)   meta.meta_server5.01010000000009fa: ranger_resource_policy_manager.cpp:304:update_policies_from_ranger_service(): ERR_INVALID_PARAMETERS: Sync policies to app envs failed.
[...]

PR #2170, opened 2024-12-13
fix(meta): fix null pointer while deleting environment variables after table was dropped

empiredan:
https://github.com/apache/incubator-pegasus/issues/2149.

Previously we've fixed the problem that meta server failed due to null pointer while
deleting environment variables locally immediately after a table was dropped in
https://github.com/apache/incubator-pegasus/pull/2148. There's the same problem
while deleting environment variables.

### DIFF
--- a/src/meta/server_state.cpp
+++ b/src/meta/server_state.cpp
@@ -3096,6 +3096,8 @@ void server_state::set_app_envs(const app_env_rpc &env_rpc)
             return;
         }
 
+        env_rpc.response().err = ERR_OK;
+
         const auto &old_envs = dsn::utils::kv_map_to_string(app->envs, ',', '=');
 
         // Update envs of local memory.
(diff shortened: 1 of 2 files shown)


src/meta/server_state.cpp
```diff
@@ -3109,70 +3112,119 @@ void server_state::set_app_envs(const app_env_rpc &env_rpc)
...
+    size_t deleted_count = 0;
     for (const auto &key : keys) {
-        if (ainfo.envs.erase(key) > 0) {
-            oss << std::endl << "    " << key;
-            deleted++;
+        if (ainfo.envs.erase(key) == 0) {
+            continue;
         }
+
+        deleted_keys_info += fmt::format("\n    {}", key);
```
acelyc111, line 3177: Will `fmt::format_to` be faster than `std::string` `+` operator?

empiredan: OK, I'll use `fmt::format_to` instead.

acelyc111 (approved):

Samunroyu (approved):

PR #2172, opened 2024-12-25
feat(meta): list tables with specified name pattern

empiredan:
Once there are a great number of tables on a cluster(say several thousands), the
whole result of executing `ls` by Pegasus shell cannot even be seen on the terminal.
If we want to see one or multiple tables with some specified pattern which could not
be shown on the terminal by `ls`, the only way is to use bash commands or even script
to list the target table.

We would allow specifying name pattern for `ls` command, to show just one or several
tables according to the pattern instead of all. This is useful while we want to find just
several tables with obvious characteristic(e.g. the same prefix) among hundreds or even
thousands of tables on the cluster.


src/client/replication_ddl_client.cpp
```diff
@@ -1477,17 +1396,22 @@ void replication_ddl_client::end_meta_request(const rpc_response_task_ptr &callb
 dsn::error_code replication_ddl_client::get_app_envs(const std::string &app_name,
                                                      std::map<std::string, std::string> &envs)
 {
+    // Just match the table with the provided name exactly since we want to get the envs from
+    // a specific table.
     std::vector<::dsn::app_info> apps;
-    auto r = list_apps(dsn::app_status::AS_AVAILABLE, apps);
-    if (r != dsn::ERR_OK) {
-        return r;
+    const auto &result = list_apps(
```
acelyc111, line 1402: There are many code snippet like this, we can wrap it by a macro in next patch to simplify the code.
```
    auto a = xxx;
    if (!a) {
        return a.code();
    }
```

src/client/replication_ddl_client.cpp
```diff
@@ -1477,17 +1396,22 @@ void replication_ddl_client::end_meta_request(const rpc_response_task_ptr &callb
...
+    if (!result) {
+        return result.code();
     }
 
-    for (auto &app : apps) {
-        if (app.app_name == app_name) {
-            envs = app.envs;
-            return dsn::ERR_OK;
+    for (const auto &app : apps) {
+        if (app.app_name != app_name) {
```
acelyc111, line 1409: While we match the app by name exactly by `PMT_MATCH_EXACT`, how can an app's name doesn't equal to `app_name`?

src/client/replication_ddl_client.h
```diff
@@ -95,22 +95,59 @@ class replication_ddl_client
...
+                      bool json,
+                      const std::string &output_file,
+                      dsn::app_status::type status);
+
+    // Create and send request to meta server to get the tables chosen to be listed according
+    // to the following parameters:
+    // `status`: the status of the tables chosen to be listed. `app_status::AS_INVALID` means
+    // no restriction.
+    // `app_name_pattern`: the name pattern of the tables chosen to be listed.
+    // `match_type`: the type in which the name pattern would be matched.
```
acelyc111, line 131: Do the 3 parameters have any difference with the ones of the above functions? If not, they can be omitted to reduce confusion.

src/client/replication_ddl_client.h
```diff
@@ -312,28 +354,52 @@ class replication_ddl_client
...
+    // The same as the above, except that the thread hash for the RPC response task is set to 0.
+    template <typename TRequest>
+    rpc_response_task_ptr request_meta(const dsn::task_code &code,
+                                       std::shared_ptr<TRequest> &req,
+                                       int timeout_milliseconds)
+    {
+        return request_meta(code, req, timeout_milliseconds, 0);
+    }
+
+    // The same as the above, except that the timeout for the RPC request is set to 0.
```
acelyc111, line 366: Add code annotation to explain what does `0` mean to the function with `timeout_milliseconds` parameter. (never timeout?)

src/meta/meta_http_service.cpp
```diff
@@ -391,15 +400,14 @@ void meta_http_service::list_node_handler(const http_request &req, http_response
     for (const auto &node : _service->_dead_set) {
         tmp_map.emplace(node, list_nodes_helper(node.to_string(), "UNALIVE"));
     }
-    int alive_node_count = (_service->_alive_set).size();
-    int unalive_node_count = (_service->_dead_set).size();
+
+    size_t alive_node_count = (_service->_alive_set).size();
+    size_t unalive_node_count = (_service->_dead_set).size();
```
acelyc111, line 405:
```suggestion
    size_t alive_node_count = _service->_alive_set.size();
    size_t unalive_node_count = _service->_dead_set.size();
```

src/client/replication_ddl_client.cpp
empiredan: OK, define `RETURN_EC_NOT_OK` to wrap it.

src/client/replication_ddl_client.cpp
empiredan: Once the meta server does not support `app_name` and `match_type` (still the old version) for `RPC_CM_LIST_APPS`, the response would include all available tables. Thus here we should still check if the table name in the response is the target. Also add this description as comments.

src/client/replication_ddl_client.h
empiredan: OK, added the comments to describe.

foreverneverer (approved):

PR #2179, opened 2025-01-12
chore(gha): introduce RockyLinux as an action runner

acelyc111:
Since the node16 has been removed in Github actions, the CentOS 7 is hard to run as
a runner, we have to make many tricky works if we want to do that.

As an alternative, this patch introduces RockyLinux 9 as a new github runner.

https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

### DIFF
--- a/.github/workflows/build-push-env-docker.yml
+++ b/.github/workflows/build-push-env-docker.yml
@@ -35,7 +35,7 @@ jobs:
   build_compilation_env_docker_images:
     runs-on: ubuntu-latest
     env:
-      # The glibc version on ubuntu1804 and centos7 is lower than the node20 required, so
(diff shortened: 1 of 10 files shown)


acelyc111:
The actions were running successfully:
https://github.com/apache/incubator-pegasus/actions/runs/12734826022/job/35492841863
https://github.com/apache/incubator-pegasus/actions/runs/12734859276/job/35492914173
https://github.com/apache/incubator-pegasus/actions/runs/12734859276/job/35493269893

empiredan (approved):

Samunroyu (approved):

Issue #2183, opened 2025-01-20
feat: update load_from_private_log of duplication task priority from LOW to COMMON


ninsmiracle:
## Feature Request

**Is your feature request related to a problem? Please describe:**
When I test the performance bottleneck of duplication, I feel strange about the task priority of  `load_from_private_log stage`.
`load_from_private_log` is the main time-consuming stage that generates IO, and is handled by the THREAD_POOL_REPLICATION_LONG thread. The priority of the dup-related tasks is also defined as LOW.
We can see all the RPC heandled by THREAD_POOL_REPLICATION_LONG:
```
// THREAD_POOL_REPLICATION_LONG
#define CURRENT_THREAD_POOL THREAD_POOL_REPLICATION_LONG
MAKE_EVENT_CODE(LPC_LEARN_REMOTE_DELTA_FILES, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE_AIO(LPC_REPLICATION_COPY_REMOTE_FILES, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_GARBAGE_COLLECT_LOGS_AND_REPLICAS, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_OPEN_REPLICA, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_CLOSE_REPLICA, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_CHECKPOINT_REPLICA, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_CATCHUP_WITH_PRIVATE_LOGS, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_DISK_STAT, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_BACKGROUND_COLD_BACKUP, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_PARTITION_SPLIT_ASYNC_LEARN, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_REPLICATION_LONG_LOW, TASK_PRIORITY_LOW)
MAKE_EVENT_CODE(LPC_REPLICATION_LONG_COMMON, TASK_PRIORITY_COMMON)
MAKE_EVENT_CODE(LPC_REPLICATION_LONG_HIGH, TASK_PRIORITY_HIGH)
[...]

PR #2184, opened 2025-01-20
feat(duplication): make the task code for incremental loading from private logs configurable

ninsmiracle:

https://github.com/apache/incubator-pegasus/issues/2183

We can make the task code configurable, allowing the thread priority incremental
loading from private logs of to be adjusted from **LOW** to **COMMON**, thereby
enabling support for low-latency real-time synchronization.

##### Performance Testing 
I do some test cases as following:
In these test cases, I first wrote 8k QPS of write traffic to the master cluster (this is the traffic that my test cluster will not generate dup log backlogs) to verify the effect of the priority modification. Then I wrote 20k QPS of write traffic to the master cluster (this is the traffic that my test cluster will generate a certain degree of dup log backlogs) to verify the effect of the priority modification.

| load/ship task priority | load_from_private_log task priority | qps  | duplicate_log_batch_bytes | plog Maximum backlog | Master cluster write delay p99 | master/slave dup delay     |
| -------------------- | -------------------------------- | ---- | ------------------------- | ----------- | ---------------- | ------------------------ |
| LOW                  | LOW                              | 8K   | 4096                      | 9k          | 1ms              | p95 127ms、p99 27473ms   |
| LOW                  | COMMON                           | 8K   | 4096                      | 200         | 1ms              | p95 101ms、p99 109ms     |
| HIGH                 | HIGH                             | 8K   | 4096                      | 150         | 1ms              | p95 107ms、p99 115ms     |
| LOW                  | LOW                              | 20K  | 4096                      | 61K         | 1.5ms            | p95 139ms、p99 20506ms   |
| LOW                  | COMMON                           | 20K  | 4096                      | 42K         | 1.5ms            | p95 126ms、p99 18127ms   |
| LOW                  | COMMON                           | 20K  | 0                         | Continue to increase over time    | 1.5ms            | 95 10618ms、p99 303519ms |

As you see , change the priority from LOW to COMMON of `load_from_private_log` will not t increase the online delay. And priority from LOW to HIGH is no benefit for further speeding up duplication.

So based on the above experimental conclusions, I think this issues' argument is valid.

### DIFF
--- a/src/replica/duplication/replica_duplicator.cpp
+++ b/src/replica/duplication/replica_duplicator.cpp
@@ -161,7 +161,7 @@ void replica_duplicator::start_dup_log()
     _load = std::make_unique<load_mutation>(this, _replica, _load_private.get());
 
     from(*_load).link(*_ship).link(*_load);
-    fork(*_load_private, LPC_REPLICATION_LONG_LOW, 0).link(*_ship);
+    fork(*_load_private, LPC_REPLICATION_LONG_COMMON, 0).link(*_ship);
 
     run_pipeline();
 }
--- a/src/replica/duplication/replica_duplicator.cpp
+++ b/src/replica/duplication/replica_duplicator.cpp
@@ -46,15 +46,14 @@ METRIC_DEFINE_counter(replica,
                       dsn::metric_unit::kMutations,
                       "The number of confirmed mutations for dup");
 
-DSN_DEFINE_bool("replication",
+DSN_DEFINE_bool(replication,
                 load_from_private_log_level_common,
                 false,
                 "The level of load_from_private_log when doing a duplication.Be false means the "
                 "task level of replaing plog is low, otherwise the task level is common");
 DSN_TAG_VARIABLE(load_from_private_log_level_common, FT_MUTABLE);
 
-namespace dsn {
-namespace replication {
+namespace dsn::replication {
 
 replica_duplicator::replica_duplicator(const duplication_entry &ent, replica *r)
     : replica_base(r),
@@ -329,5 +328,5 @@ void replica_duplicator::set_duplication_plog_checking(bool checking)
     _replica->set_duplication_plog_checking(checking);
 }
 
-} // namespace replication
-} // namespace dsn
+} // namespace dsn::replication
+
--- a/src/replica/duplication/replica_duplicator.cpp
+++ b/src/replica/duplication/replica_duplicator.cpp
@@ -46,11 +46,12 @@ METRIC_DEFINE_counter(replica,
                       dsn::metric_unit::kMutations,
                       "The number of confirmed mutations for dup");
 
-DSN_DEFINE_string(replication,
-                  dup_load_plog_task,
-                  "LPC_REPLICATION_LONG_LOW",
-                  "The task code for incremental loading from private logs while duplicating. Tasks with "
-                  "TASK_PRIORITY_HIGH are not recommended.");
+DSN_DEFINE_string(
+    replication,
+    dup_load_plog_task,
+    "LPC_REPLICATION_LONG_LOW",
+    "The task code for incremental loading from private logs while duplicating. Tasks with "
+    "TASK_PRIORITY_HIGH are not recommended.");
 DSN_TAG_VARIABLE(dup_load_plog_task, FT_MUTABLE);
 
 namespace dsn::replication {
@@ -169,12 +170,11 @@ void replica_duplicator::start_dup_log()
     _load = std::make_unique<load_mutation>(this, _replica, _load_private.get());
 
     from(*_load).link(*_ship).link(*_load);
-    auto dup_load_plog_task =
-        dsn::task_code::try_get(FLAGS_dup_load_plog_task, TASK_CODE_INVALID);
+    auto dup_load_plog_task = dsn::task_code::try_get(FLAGS_dup_load_plog_task, TASK_CODE_INVALID);
     if (dup_load_plog_task == TASK_CODE_INVALID) {
         dup_load_plog_task = LPC_REPLICATION_LONG_LOW;
         LOG_ERROR_PREFIX("invalid dup_load_plog_task ({}), set it to LPC_REPLICATION_LONG_LOW",
-                                              FLAGS_dup_load_plog_task);
+                         FLAGS_dup_load_plog_task);
     }
     fork(*_load_private, dup_load_plog_task, 0).link(*_ship);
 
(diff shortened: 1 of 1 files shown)


src/replica/duplication/replica_duplicator.cpp
```diff
@@ -161,7 +161,7 @@ void replica_duplicator::start_dup_log()
     _load = std::make_unique<load_mutation>(this, _replica, _load_private.get());
 
     from(*_load).link(*_ship).link(*_load);
-    fork(*_load_private, LPC_REPLICATION_LONG_LOW, 0).link(*_ship);
+    fork(*_load_private, LPC_REPLICATION_LONG_COMMON, 0).link(*_ship);
```
acelyc111, line 164: The reason of priority set to LPC_REPLICATION_LONG_LOW is to reduce the impact of duplication on normal read/write operations on the master cluster. There are some usage scenarios of the duplication, some user cares about the latency, some cares about the master cluster's thoughput.

How about add a configuration for it?

ninsmiracle: I also want to add this configuration, but I haven't decided whether to make it a table-level configuration or a cluster-level configuration. 
At present, the duplicate_log_batch_bytes parameter is cluster-level, so should I make this parameter cluster-level as well? This is more convenient to implement.

acelyc111: Cluster level is OK.

acelyc111 (approved):

src/replica/duplication/replica_duplicator.cpp
```diff
@@ -35,17 +35,25 @@
...
 
 METRIC_DEFINE_counter(replica,
                       dup_confirmed_mutations,
                       dsn::metric_unit::kMutations,
                       "The number of confirmed mutations for dup");
 
-namespace dsn {
-namespace replication {
+DSN_DEFINE_bool(replication,
+                load_from_private_log_level_common,
```
empiredan, line 50: False `load_from_private_log_level_common` does not necessarily mean `low`: it may be `high`. I think we could use `DSN_DEFINE_string` to define the task name directly, and initialize the code object by `task_code::try_get`.

src/replica/duplication/replica_duplicator.cpp
```diff
@@ -161,7 +170,12 @@ void replica_duplicator::start_dup_log()
     _load = std::make_unique<load_mutation>(this, _replica, _load_private.get());
 
     from(*_load).link(*_ship).link(*_load);
-    fork(*_load_private, LPC_REPLICATION_LONG_LOW, 0).link(*_ship);
+    const dsn::task_code load_from_private_log_level =
+        dsn::task_code::try_get(FLAGS_load_from_private_log_level, LPC_REPLICATION_LONG_LOW);
+    LOG_ERROR_PREFIX("unexpected load_from_private_log_level ({}), put it to default value "
+                     "LPC_REPLICATION_LONG_LOW",
+                     load_from_private_log_level);
+    fork(*_load_private, load_from_private_log_level, 0).link(*_ship);
```
empiredan, line 178:
```suggestion
    auto dup_load_plog_task =
        dsn::task_code::try_get(FLAGS_dup_load_plog_task, TASK_CODE_INVALID);
    if (dup_load_plog_task == TASK_CODE_INVALID) {
        dup_load_plog_task = LPC_REPLICATION_LONG_LOW;
        LOG_ERROR_PREFIX("invalid dup_load_plog_task ({}), set it to LPC_REPLICATION_LONG_LOW",
                                              FLAGS_dup_load_plog_task);
    }
    fork(*_load_private, dup_load_plog_task, 0).link(*_ship);
```

src/replica/duplication/replica_duplicator.cpp
```diff
@@ -35,17 +35,26 @@
...
                       "The number of confirmed mutations for dup");
 
-namespace dsn {
-namespace replication {
+DSN_DEFINE_string(replication,
+                  load_from_private_log_level,
+                  "LPC_REPLICATION_LONG_LOW",
+                  "The level of load_from_private_log when doing a duplication.Be false means the "
+                  "task level of replaing plog is low, otherwise the task level is common (We do "
+                  "not recommend high level)");
```
empiredan, line 54:
```suggestion
                  "The task code for incremental loading from private logs while duplicating. Tasks with "
                  "TASK_PRIORITY_HIGH are not recommended.");
```

acelyc111 (approved):

ninsmiracle:
### Add some information about dup sending delay
  We conducted multiple control experiments on the test cluster with `duplicate_log_batch_bytes` of 0, 4096, and 8192. It can be clearly seen that configuring a larger `duplicate_log_batch_bytes` can improve the consumption capacity of the cluster dup. For the table below, when `duplicate_log_batch_bytes` is configured to 8192, the cluster is still able to consume writes at 40k write QPS; but if `duplicate_log_batch_bytes` is configured to 0, the cluster loses its consumption capacity at 20k write QPS. However, if the cluster dup can consume existing writes, the larger the `duplicate_log_batch_bytes`, the longer the delay in dup a piece of data between the master and slave clusters.
​  Explain the 4th and 5th columns of the following table. When the delay between the master and standby clusters is too small, the delay data displayed by the monitoring is inaccurate. This is due to the counter reporting granularity. So we make a program to read and write the corresponding keys on both sides to determine the precise delay. However, when the delay between the master and slave clusters is too large, the delay of reading and writing each shard takes too long and is sometimes difficult to calculate. Therefore, we mainly use monitoring data to compare the experimental results in the scenario of large delay.

<!DOCTYPE html>
qps | plog Maximum backlog | duplicate_log_batch_bytes | master/slave dup delay p99(Monitoring delay avg) | master/slave dup delay program test
-- | -- | -- | -- | --
0 | 3 | 0 |   | p95 105ms/p99 108ms
0 | 3 | 4096 |   | p95 106ms/p99 108ms
0 | 3 | 8192 |   | p95 127ms/p99 150ms
8k | 13K | 0 | 120ms | p95 106ms/p99 137ms
8K | 17k | 4096 | 3.7s | p95 119ms/p99 1673ms
8K | 17.2k | 8192 | 6s | p95 138ms/p99 20s
20K | Continue to increase | 0 | Continue to increase | Difficult to observe
20K | 75k | 4096 | 25s | Difficult to observe
20K | 70k | 8192 | 25s | Difficult to observe
30K | 120k | 8192 | 26s | Difficult to observe
40K | 24k | 8192 | 28s | Difficult to observe
45K | Continue to increase | 8192 | Continue to increase | Difficult to observe

==================================================

And here is an effect of adjusting the parameters of one of our online clusters:

<!DOCTYPE html>
集群名 | duplicate_log_batch_bytes = 4096 | duplicate_log_batch_bytes = 0
-- | -- | --
c3srv-online | p95 1008msp99 1327ms | p95 100msp99 108ms

Issue #2194, opened 2025-02-17
feat: reduce partition ingest time


ruojieranyishen:
## Feature Request

**Is your feature request related to a problem? Please describe:**
During the ingest phase, partitions block writes. Through testing, this time is approximately 30 seconds, while the rocksDB ingest time is around 200 milliseconds. The main reason is that during the ingest phase, the time interval for the meta to send RPC_BULK_LOAD is too long (once every 10 seconds).

| Single - node ingest concurrency level | Table - level ingest time (ms) | Minimum partition - level ingest time (ms) | Maximum partition - level ingest time (ms) | Average partition - level ingest time (ms) | Highest Rocksdb ingest time (ms) | Average Rocksdb ingest time (ms) |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- |
| 8 | 470006 | 15007 | 30075 | 27244 | 618 | 459.06 |
| 16 | 240005 | 19978 | 30002 | 28652 | 580 | 482.43 |
| 100 | 40004 | 29967 | 29986 | 29978 | 620 | 509.65 |

**Describe the feature you'd like:**
reduce partition ingest time

AntiTopQuark:
Hi, I would like to work on this function. Could you please assign the issue to me?

ruojieranyishen:
> Hi, I would like to work on this function. Could you please assign the issue to me?
Sorry, I have already submitted the code for this issue. Would you like me to find some other issues for you?

ruojieranyishen:
@AntiTopQuark Sorry, I have already submitted the code for this issue. Would you like me to find some other issues for you?

AntiTopQuark:
> [@AntiTopQuark](https://github.com/AntiTopQuark) Sorry, I have already submitted the code for this issue. Would you like me to find some other issues for you?

No worries at all! Thanks for letting me know. : )

PR #2195, opened 2025-02-17
feat(bulk_load): reduce partition ingest time

ruojieranyishen:
https://github.com/apache/incubator-pegasus/issues/2194

The main reason is that during the ingest phase, the time between meta's RPC_BULK_LOAD transmissions is too long (once every 10 seconds). Shortening this time interval can cut down the partition write-blocking time.

##### Tests 
Before：partition ingest time 30s.
After refactor: partition ingest time 300~900ms.

### DIFF
--- a/src/meta/meta_bulk_load_service.cpp
+++ b/src/meta/meta_bulk_load_service.cpp
@@ -1266,11 +1266,12 @@ void bulk_load_service::partition_ingestion(const std::string &app_name, const g
     if (!try_partition_ingestion(pc, app->helpers->contexts[pid.get_partition_index()])) {
         LOG_WARNING(
             "app({}) partition({}) couldn't execute ingestion, wait and try later", app_name, pid);
-        tasking::enqueue(LPC_META_STATE_NORMAL,
-                         _meta_svc->tracker(),
-                         std::bind(&bulk_load_service::partition_ingestion, this, app_name, pid),
-                         pid.thread_hash(),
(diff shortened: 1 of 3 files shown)


src/meta/meta_bulk_load_service.cpp
```diff
@@ -587,11 +587,14 @@ void bulk_load_service::try_resend_bulk_load_request(const std::string &app_name
...
     if (is_app_bulk_loading_unlocked(pid.get_app_id())) {
+        int32_t interval = _partition_bulk_load_info[pid].status == bulk_load_status::BLS_INGESTING
+                               ? bulk_load_constant::BULK_LOAD_INGEST_REQUEST_INTERVAL
+                               : bulk_load_constant::BULK_LOAD_REQUEST_INTERVAL;
         tasking::enqueue(LPC_META_STATE_NORMAL,
                          _meta_svc->tracker(),
                          std::bind(&bulk_load_service::partition_bulk_load, this, app_name, pid),
                          0,
-                         std::chrono::seconds(bulk_load_constant::BULK_LOAD_REQUEST_INTERVAL));
+                         std::chrono::milliseconds(interval));
```
empiredan, line 597: Could this be made configurable, or 10 seconds and 150 ms carry special meaning ?

ruojieranyishen: The partition_bulk_load function prompts the replica to check if the RocksDB ingest is finished. Shortening its execution interval can reduce the write-blocking time of the partition, from an average of 30 seconds to 500 milliseconds.

empiredan (approved):

Issue #2215, opened 2025-03-24
bug: When loading the replica in the replica_server, the assertion !missing_meta_cf failed


limowang:
When we use rolling upgrades and duplication for cluster migration, we encountered a situation where the `replica_server` failed to start. Upon checking the corresponding logs, the following output was found:

E2025-03-20 17:01:49.749 (1742461309749987408 28d421) replica.replica22.020100000000014f: pegasus_server_impl.cpp:1603:start(): assertion expression: !missing_meta_cf
F2025-03-20 17:01:49.750 (1742461309750019459 28d421) replica.replica22.020100000000014f: pegasus_server_impl.cpp:1603:start(): [42.0@##] You must upgrade Pegasus server from 2.0

PR #2216, opened 2025-03-24
bug: When loading the replica in the replica_server, the assertion !missing_meta_cf failed

limowang:
[#2215](https://github.com/apache/incubator-pegasus/issues/2215)

During rolling upgrades and duplication data migration, there are cases where a replica first learns the checkpoint from the primary replica and then loads the partition. However, if the primary replica crashes during the execution of `::dsn::error_code pegasus_server_impl::copy_checkpoint_to_dir_unsafe(const char *checkpoint_dir, int64_t *checkpoint_decree, bool flush_memtable)` while generating the checkpoint, it may skip the final block of code:

```cpp
std::vector<rocksdb::ColumnFamilyDescriptor> column_families(
    {{DATA_COLUMN_FAMILY_NAME, rocksdb::ColumnFamilyOptions()},
     {META_COLUMN_FAMILY_NAME, rocksdb::ColumnFamilyOptions()}});
status = rocksdb::DB::OpenForReadOnly(
    rocksdb::DBOptions(), checkpoint_dir, column_families, &handles_opened, &snapshot_db);
if (!status.ok()) {
    derror_replica(
        "OpenForReadOnly from {} failed, error = {}", checkpoint_dir, status.ToString());
    snapshot_db = nullptr;
    cleanup(true);
}
dcheck_eq_replica(handles_opened.size(), 2);
dcheck_eq_replica(handles_opened[1]->GetName(), META_COLUMN_FAMILY_NAME);
```
As a result, the META_COLUMN_FAMILY_NAME column family might be missing from the checkpoint. When other potential follower replicas request checkpoint data, the primary replica does not validate whether the checkpoint data is complete, which could lead to triggering the !missing_meta_cf assertion failure.

Therefore, we can add checkpoint integrity validation in the primary replica when sending checkpoint data, specifically in the ::dsn::error_code pegasus_server_impl::get_checkpoint(int64_t learn_start, const dsn::blob &learn_request, dsn::replication::learn_state &state) function.


src/server/pegasus_server_impl.cpp
```diff
@@ -184,6 +184,48 @@ const std::unordered_map<std::string, cf_opts_getter> cf_opts_getters = {
...
+    std::vector<rocksdb::ColumnFamilyDescriptor> column_families(
+        {{meta_store::DATA_COLUMN_FAMILY_NAME, rocksdb::ColumnFamilyOptions()},
+         {meta_store::META_COLUMN_FAMILY_NAME, rocksdb::ColumnFamilyOptions()}});
+    auto status = rocksdb::DB::OpenForReadOnly(
+        _db_opts, checkpoint_dir, column_families, &handles_opened, &snapshot_db);
+    if (!status.ok()) {
+        LOG_ERROR_PREFIX(
+            "OpenForReadOnly from {} failed, error = {}", checkpoint_dir, status.ToString());
+        snapshot_db = nullptr;
+        cleanup();
```
acelyc111, line 215: use defer to execute `cleanup()` automatically

limowang: Okay, thank you for your advice ！

.clang-tidy
```diff
@@ -20,7 +20,7 @@
...
...,-concurrency-mt-unsafe,-cppcoreguidelines-avoid-c-arrays,-cppcoreguidelines-avoid-magic-numbers,-cppcoreguidelines-avoid-non-const-global-variables,-cppcoreguidelines-macro-usage,-cppcoreguidelines-non-private-member-variables-in-classes,-cppcoreguidelines-owning-memory,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-pro-bounds-pointer-arithmetic,-cppcoreguidelines-pro-type-const-cast,-cppcoreguidelines-pro-type-union-access,-fuchsia-default-arguments-calls,-fuchsia-multiple-inheritance,-fuchsia-overloaded-operator,-fuchsia-statically-constructed-objects,-google-readability-avoid-underscore-in-googletest-name,-hicpp-avoid-c-arrays,-hicpp-named-parameter,-hicpp-no-array-decay,-llvm-include-order,-misc-definitions-in-headers,-misc-non-private-member-variables-in-classes,-misc-unused-parameters,-modernize-avoid-bind,-modernize-avoid-c-arrays,-modernize-replace-disallow-copy-and-assign-macro,-modernize-use-trailing-return-type,-performance-unnecessary-value-param,-readability-function-cognitive-complexity,-readability-identifier-length,-readability-magic-numbers,-readability-named-parameter,-readability-suspicious-call-argument,-abseil-string-find-str-contains'
```
empiredan, line 23: Which rules are disabled ? Should also be consistent with [build_tools/clang_tidy.py](https://github.com/apache/incubator-pegasus/blob/master/build_tools/clang_tidy.py).

limowang: The parameter -abseil-string-find-str-contains was added, otherwise the line d1.find("checkpoint") != std::string::npos would not pass the format check (but I didn't modify it).

PR #2238, opened 2025-04-21
feat(duplication): support resumable download for checkpoint files during full duplication on the server side

empiredan:
https://github.com/apache/incubator-pegasus/issues/2242

When a dup follower requests the latest checkpoint information from the dup
master, the server supports returning the size and checksum of each file under
the checkpoint directory. This allows the client to decide which files to fetch:
- if it is the first time downloading the checkpoint, all files need to be downloaded;
- if it is the second time or later, only files that are missing or partially downloaded
locally need to be fetched.

Additionally, the client is allowed to specify the checksum algorithm to use —
currently, only MD5 is supported.

### DIFF
--- a/src/test_util/test_util.h
+++ b/src/test_util/test_util.h
@@ -93,6 +93,8 @@ private:
 
 void create_local_test_file(const std::string &full_name, dsn::replication::file_meta *fm);
 
+// Generate a file under `file_path` with size `file_size`. The file is filled with
+// `file_size` 'a's.
 void generate_test_file(const std::string &file_path, int64_t file_size);
 
 #define ASSERT_EVENTUALLY(expr)                                                                    \
--- a/src/test_util/test_util.h
+++ b/src/test_util/test_util.h
@@ -30,12 +30,13 @@
 
 #include "metadata_types.h"
 #include "runtime/api_layer1.h"
-#include "utils/env.h"
 // IWYU refused to include "utils/defer.h" everywhere, both in .h and .cpp files.
 // However, once "utils/defer.h" is not included, it is inevitable that compilation
 // will fail since dsn::defer is referenced. Thus force IWYU to keep it.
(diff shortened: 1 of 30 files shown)


src/test_util/test_util.h
```diff
@@ -92,6 +93,10 @@ class stop_watch
 
 void create_local_test_file(const std::string &full_name, dsn::replication::file_meta *fm);
 
+// Generate a file under `file_path` with size `file_size`. The file is filled with
+// `file_size` 'a's.
+void generate_test_file(const std::string &file_path, int64_t file_size);
```
acelyc111, line 98: Will the file to be deleted after the tests?

src/utils/test/checksum_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+    {4097, utils::checksum_type::CST_INVALID, ERR_NOT_IMPLEMENTED, ""},
+};
+
+// Calculate checksum for encrypted files.
+TEST_P(CalcChecksumTest, CalcEncrypted) { test_calc_checksum(true); }
+
+// Calculate checksum for unencrypted files.
+TEST_P(CalcChecksumTest, CalcUnencrypted) { test_calc_checksum(false); }
+
+INSTANTIATE_TEST_SUITE_P(ChecksumTest, CalcChecksumTest, testing::ValuesIn(calc_checksum_tests));
```
acelyc111, line 102: The encryption switch can also be parameterized in INSTANTIATE_TEST_SUITE_P by an additional parameter, e.g. `::testing::Combine(a_values, b_values)`

src/test_util/test_util.h
empiredan: OK, I'll implement a class managing the file's life cycle with RAII.

acelyc111 (approved):

foreverneverer (approved):

PR #2240, opened 2025-04-25
ci(github): fix Go workflows which failed due to Ubuntu 20.04 retirement

empiredan:
Resolve https://github.com/apache/incubator-pegasus/issues/2241.

Go workflows run on Ubuntu 20.04 because Go client uses thrift-compiler 0.13.0
to generate code and the thrift-compiler version on Ubuntu 20.04 is 0.13.0. However
Ubuntu 20.04 runner has been retired and removed on 2025-04-15. To solve this
problem, we choose to build thrift 0.13.0 in a image based on Ubuntu 22.04.

Also add Collector to the workflow for regular building.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,10 +19,10 @@ linters:
   enable:
     - gofmt
     - goimports
-    - golint
+    - revive
     - bodyclose
     - exhaustive
-    - exportloopref
+    - copyloopvar
 
 linters-settings:
   exhaustive:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,13 +16,21 @@
 # under the License.
 ---
 linters:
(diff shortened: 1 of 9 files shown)


empiredan:
The workflows of go-client and collector failed because both of them are used for pull requests, which means we must use `github.base_ref` rather than `github.ref_name`: Go image for `github.base_ref` (namely master branch) will only be built after this pull request is merged into master branch.

.golangci.yml
```diff
@@ -16,6 +16,14 @@
 # under the License.
 ---
 linters:
+  # TODO(wangdan): consider replacing 'golint' with 'revive' because 'golint' is
```
acelyc111, line 19: 👍

acelyc111 (approved):

foreverneverer (approved):

Issue #2197, opened 2025-02-21
Support making atomic write requests idempotent


empiredan:
# Motivation

Pegasus does not support duplicating atomic write requests including `incr`, `check_and_set` and `check_and_mutate` since they are not idempotent. In practice, various applications use atomic write interfaces in many scenarios. However, such applications cannot use duplication to synchronize data, and therefore cannot benefit from the high performance that duplication provides.

# Design

Due to the urgency of the requirements, the first version of the idempotent implementation for the atomic writes should be as simple as possible, without making fundamental changes to the write path.

Therefore, we decided to implement the idempotence of atomic write requests as follows: for each replica, ensure that only one atomic write request is being processed in the write pipeline at any given time. Once the replica server receives an atomic request, firstly it will be cached. It will not be pushed into the write pipeline until all requests before it have been applied.  The write pipeline consists of the following stages:
1. read the current value from RocksDB, calculate the final value according to specific semantics of requested atomic write and build the idempotent request based on it;
2. append the corresponding mutation to plog;
3. broadcast the prepare requests to the secondary replicas;
4. apply the final result back to RocksDB ultimately;

The primary replicas have all 1 ~ 4 stages, and at last reply to the client while the secondary replicas only have stages 2 and 4.

# Task List

Making incr requests idempotent:
- [ ] https://github.com/apache/incubator-pegasus/pull/2185
- [ ] https://github.com/apache/incubator-pegasus/pull/2192
- [ ] https://github.com/apache/incubator-pegasus/pull/2196

PR #2246, opened 2025-05-07
feat(make_idempotent): support making `check_and_mutate` request idempotent in `pegasus_write_service::impl`

empiredan:
https://github.com/apache/incubator-pegasus/issues/2197

Implement two APIs in `pegasus_write_service::impl` to make `check_and_mutate`
requests idempotent.

Unlike `incr` and `check_and_set`, a `check_and_mutate` request may be translated
into multiple `put` and `remove` requests. To conveniently store both in the same
`std::vector`, they are all converted into `update_request` objects, with the `type`
field used to distinguish between them:

- implement `make_idempotent()` function to translate a `check_and_mutate` request
into multiple single-put and single-remove requests which are naturally idempotent,
and process the possible errors during the translation, e.g. failed to read the check
value or did not pass the check;
- implement `put()` function to apply multiple single-put and single-remove requests
into the RocksDB instance (certainly the atomicity of the batch operations must be
guaranteed) and make response for `check_and_mutate`.

### DIFF
--- a/src/server/pegasus_write_service_impl.h
+++ b/src/server/pegasus_write_service_impl.h
@@ -532,8 +532,8 @@ public:
         return rocksdb::Status::kOk;
     }
 
-    // Used to call make_idempotent for incr and check_and_set to get the idempotent single
-    // put request which is stored as the only element of `updates`.
+    // Used to call make_idempotent() for incr and check_and_set to get the idempotent single-put
+    // request which is stored as the unique element of `updates`.
     //
     // This interface is provided to ensure consistency between the make_idempotent() interfaces
     // of incr/check_and_set operations and that of check_and_mutate (both using std::vector for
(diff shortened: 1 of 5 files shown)


src/server/pegasus_write_service_impl.h
```diff
@@ -540,6 +532,165 @@ class pegasus_write_service::impl : public dsn::replication::replica_base
...
+                _pegasus_data_version, std::move(get_ctx.raw_value), check_value);
+        }
+
+        bool invalid_argument = false;
+        const bool passed = validate_check(
+            req.check_type, req.check_operand, value_exist, check_value, invalid_argument);
+        if (!passed) {
+            make_check_value(req, value_exist, check_value, err_resp);
+            return make_error_response(invalid_argument ? rocksdb::Status::kInvalidArgument
+                                                        : rocksdb::Status::kTryAgain,
```
acelyc111, line 619: Why return the error code kTryAgain? Is it very likely to success if the caller try again?

acelyc111 (approved):

foreverneverer (approved):

src/server/pegasus_write_service_impl.h
empiredan: Great question! The reason for returning `kTryAgain` here is that the condition check in `check_and_mutate` failed, which signals the caller not to proceed with the subsequent steps (including 2PC and applying to the storage engine). At the same time, it responds to the client indicating that no other error occurred — it's simply that the condition check didn’t pass. If necessary, the client can still choose to resubmit the request to the server. `kTryAgain` is chosen as the return code also to stay consistent with the previous non-idempotent version of `check_and_mutate()`.

Issue #2201, opened 2025-03-05
How about introducing Rcoksdb BlobDB feature?


ninsmiracle:
## Feature Request

**Is your feature request related to a problem? Please describe:**
We have some business requirements to write relatively large kv into the database.
For example, a user wants to write hundreds of thousands of data per second, and every singal date is about 80KB.
We have made many attempts to optimize the write parameters, such as adjust the memtable number/size of rocksdb,but they are far away from our target.
I noticed a paper named **WiscKey** https://www.usenix.org/system/files/conference/fast16/fast16-papers-lu.pdf, this paper proposes a solution to separate the key and value of RocksDB, which helps to significantly reduce the write amplification factor.
I read the logic of Rocksdb **BlobDB** carefully (Inspiration comes from **WiscKey**):https://rocksdb.org/blog/2021/05/26/integrated-blob-db.html, I think it's the time to introduce the feature to pegasus now.

**Describe the feature you'd like:**
I'd like to implement this feature in Pegasus.

acelyc111:
Good proposal! Is there any breaking changes? I guess it's an optional to enable BlobDB or not, right?

ninsmiracle:
I read the source code of rocksdb. If it is a version after 8.0.0, the implementation of blobdb is relatively complete. Therefore, I think there is no need to make too many changes to introduce blobdb into the Pegasus current master branch. But specifically, I need to test this function on a test cluster.

ninsmiracle:
> Good proposal! Is there any breaking changes? I guess it's an optional to enable BlobDB or not, right?

I read the source code of rocksdb. If it is a version after 8.0.0, the implementation of blobdb is relatively complete. Therefore, I think there is no need to make too many changes to introduce blobdb into the Pegasus current master branch. But specifically, I need to test this function on a test cluster.

acelyc111:
> > Good proposal! Is there any breaking changes? I guess it's an optional to enable BlobDB or not, right?
> 
> I read the source code of rocksdb. If it is a version after 8.0.0, the implementation of blobdb is relatively complete. Therefore, I think there is no need to make too many changes to introduce blobdb into the Pegasus current master branch. But specifically, I need to test this function on a test cluster.


Good, look forward to your good news!

PR #2251, opened 2025-05-19
feat: introduce rocksdb blobdb

ninsmiracle:
#2201 

introduce rocksdb blobdb interface

##### Tests 

- Unit test

##### Code changes
- Has interface methods change in `src/server/pegasus_server_impl_init.cpp`

### DIFF
--- a/src/server/pegasus_server_impl_init.cpp
+++ b/src/server/pegasus_server_impl_init.cpp
@@ -588,6 +588,58 @@ DSN_DEFINE_uint64(pegasus.server,
                   600, // 600 is the default value in RocksDB.
                   "If not zero, dump rocksdb.stats to RocksDB every stats_persist_period_sec");
 
+/* Rocksdb blobdb for Key-value separation.
+ * For more infomation, see: https://github.com/facebook/rocksdb/wiki/BlobDB */
+DSN_DEFINE_bool(
+    pegasus.server,
+    rocksdb_enable_blob_files,
+    false,
+    "switch of the key-value separation function"
+);
+
+DSN_DEFINE_uint32(
+    pegasus.server,
+    rocksdb_min_blob_size,
+    4 * 1024,  // 4KB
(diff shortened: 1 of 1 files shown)


src/server/pegasus_server_impl_init.cpp
```diff
@@ -693,6 +745,15 @@ pegasus_server_impl::pegasus_server_impl(dsn::replication::replica *r)
     _data_cf_opts.max_bytes_for_level_base = FLAGS_rocksdb_max_bytes_for_level_base;
     _data_cf_opts.max_bytes_for_level_multiplier = FLAGS_rocksdb_max_bytes_for_level_multiplier;
 
+    // open db with key-value separation option (rocksdb blobdb)
+    _data_cf_opts.enable_blob_files = FLAGS_rocksdb_enable_blob_files;
```
acelyc111, line 749: What happen if enable this switch on an already running rocksdb instance? Will the data to be lost?

ninsmiracle: `enable_blob_files` will affects  flush/compaction task after your change, and will not  interrupting ongoing background tasks.   
If your turn it from 'true' to 'false', and at the same time  `enable_blob_garbage_collection=true`.:
The background Blob GC thread will continue to scan old files. When the garbage rate of the entire file exceeds the threshold (`blob_garbage_collection_force_threshold`), the surviving entries in the file will be moved to the new SST (not the new blob), and then the original file will be deleted.
On the other hand, if `enable_blob_garbage_collection=false`, or the keys of this blob file remain unchanged for a long time, the blob file  will stay in rocksdb for a long time.

acelyc111: @ninsmiracle If `enable_blob_files` is set as true at the beginning, I wrote `k1` -> `v1` to a table. Then set it to false and reboot the replica servers, can I read the former `k1` from the DB? How it act if the data in both SST and blob files?

Issue #2258, opened 2025-05-30
`ddd_diagnose` failed while trying to diagnosing all tables without specifying any table


empiredan:
Error occurred while executing `ddd_diagnose` command on shell:

```
>>> ddd_diagnose
ERROR: DDD diagnose failed with err = ERR_INVALID_PARAMETERS
```

Following warning message was found while checking logs for meta server:

```
W2025-05-30 14:56:49.556 (1748588209556064035 14634) meta.meta_server4.72120373587612541: meta_service.cpp:1088:ddd_diagnose(): reject request on app_id = -1
```

PR #2259, opened 2025-05-30
fix(security): fix DDD diagnose request failure with `app_id = -1` and improve ACL handling

empiredan:
Fix https://github.com/apache/incubator-pegasus/issues/2258.

When the client (via shell) decides to diagnose all DDD partitions, it sets
`app_id = -1` and sends the request to the meta server. Upon receiving the
request, the meta server attempts to look up the table name based on `app_id`
and then performs ACL checks to verify whether the user has permission to
access the table. However, since `app_id = -1` represents **all** tables rather
than a specific one, there is no corresponding table name to look up. As a result,
the lookup fails and the meta server returns an `ERR_INVALID_PARAMETERS`
error to the client.

To fix this issue, the meta server must first check whether `app_id == -1` when it
receives a DDD diagnose request:
- If `app_id == -1`, the ACL check is deferred: the server first retrieves all DDD
partitions and then performs ACL checks on each one individually, returning only
those partitions that the client user is authorized to access.
- If `app_id != -1`, the `app_id` must correspond to a real table, so the existing logic
remains unchanged: the server resolves the table name by `app_id` and performs a
single and standard ACL check.

In addition to this root cause, several related issues were discovered and fixed as part
of this change:
- Changed the access type for `RPC_CM_DDD_DIAGNOSE` from `access_type::kControl`
to `access_type::kList`, since the request only retrieves partition information and does
not modify any data or metadata.
- The meta server previously handled only Ranger ACLs for the `RPC_CM_LIST_APPS`
request, ignoring the legacy ACLs; this has now been fixed, where legacy ACLs are also
respected.
- Refactored some ACL-related code for clarity and maintainability.

### DIFF
--- a/src/ranger/ranger_resource_policy_manager.cpp
+++ b/src/ranger/ranger_resource_policy_manager.cpp
@@ -165,7 +165,7 @@ ranger_resource_policy_manager::ranger_resource_policy_manager(
                               "RPC_CLI_CLI_CALL_ACK"},
                              _ac_type_of_global_rpcs);
     // DATABASE - kList
-    register_rpc_access_type(access_type::kList, {"RPC_CM_LIST_APPS"}, _ac_type_of_database_rpcs);
+    register_rpc_access_type(access_type::kList, {"RPC_CM_LIST_APPS", "RPC_CM_DDD_DIAGNOSE"}, _ac_type_of_database_rpcs);
     // DATABASE - kCreate
     register_rpc_access_type(
         access_type::kCreate, {"RPC_CM_CREATE_APP"}, _ac_type_of_database_rpcs);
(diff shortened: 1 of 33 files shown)


src/ranger/ranger_resource_policy_manager.cpp
```diff
@@ -206,7 +208,12 @@ ranger_resource_policy_manager::ranger_resource_policy_manager(
 
 void ranger_resource_policy_manager::start()
 {
-    CHECK_NOTNULL(_meta_svc, "");
+    if (_meta_svc == nullptr) {
+        // `_meta_svc` being null implies that the policies will be updated manually,
```
acelyc111, line 212: Is the situation only possible in tests?

src/zookeeper/distributed_lock_service_zookeeper.cpp
```diff
@@ -31,45 +31,46 @@
...
-            _zookeeper_locks.clear();
+    if (_session == nullptr) {
+        return;
+    }
+
+    std::vector<lock_struct_ptr> handle_vec;
+    {
+        utils::auto_write_lock l(_service_lock);
+
+        for (const auto &zk_lock : _zookeeper_locks) {
```
acelyc111, line 61:
```suggestion
        for (const auto &[_, handle] : _zookeeper_locks) {
```

src/ranger/ranger_resource_policy_manager.cpp
empiredan: Yes, it's only used for test. I'll add `MOCK_TEST` macro to control it. Do not use `NDEBUG` since tests running on Github are built with `Release` rather than `Debug`.

acelyc111 (approved):

foreverneverer (approved):

PR #2261, opened 2025-06-16
perf(make_idempotent): improve performance of idempotent writes by reducing deserialization and removing the mapping table for atomic write handlers

empiredan:
https://github.com/apache/incubator-pegasus/issues/2197

Introduce `idempotent_writer` class to the primary replica to cache the idempotent
single-update requests generated during the `make_idempotent` phase as well as
the original atomic write RPC request from the client to improve performance, since：
1. both deserializations for single-update requests and the original atomic write RPC
request are reduced;
2. use `std::variant` to hold the applying function and the RPC for each atomic write
instead of the mapping table for its handler.


src/rpc/rpc_holder.h
```diff
@@ -224,9 +224,13 @@ class rpc_holder
         dsn_rpc_forward(dsn_request(), addr);
     }
 
-    inline void set_auto_reply(bool auto_reply) { _i->auto_reply = auto_reply; }
+    [[nodiscard]] inline bool is_auto_reply() const { return _i->auto_reply; }
 
-    inline void enable_auto_reply() { set_auto_reply(true); }
+    inline void set_auto_reply(bool auto_reply) const { _i->auto_reply = auto_reply; }
```
acelyc111, line 229: Can it be private?

acelyc111 (approved):

src/rpc/rpc_holder.h
empiredan: Yes, currently we've never called `set_auto_reply()` externally. So it could be made private.

foreverneverer (approved):

PR #2272, opened 2025-07-17
perf(network): support client RPC connection pool to each server address

empiredan:
# 1. Motivation

Currently, for each remote server (\<ip\>\:\<port\>), at most one client connection
is established, and all RPCs share this single connection. This approach is still
acceptable when network latency is low, but it becomes problematic when the
latency is high (e.g., cross-country or intercontinental communication where latency
can exceed 100 ms). Since all RPCs share one connection, high network latency can
cause one RPC to time out and fail, which in turn may lead to subsequent RPC failures.
This is especially critical for cases like duplication, where copying a checkpoint may
involve transferring large amounts of data — high latency can make duplication almost
impossible.

To address this, we introduce a client connection pool for each remote server (\<ip\>\:\<port\>).
This allows better utilization of network bandwidth when there are many RPCs (each
potentially involving large data transfers), reduces the risk of frequent RPC timeouts,
and provides significant improvements in scenarios with high network latency.

# 2. API

## 2.1 Get Session

The most important interface of the connection pool is `get_rpc_session()`. External
callers only need to use this interface to obtain a session (i.e., a connection) from the
pool, without worrying about how the connection is created, connected, or reused.

## 2.2 Callbacks

When a session is successfully established (connection creation is generally asynchronous
and will notify via a callback upon completion) or disconnected (I/O is generally asynchronous,
and if an error is detected in a completion callback, the connection should be closed),
the pool’s `on_connected()` and `on_disconnected()` interfaces must be called to notify
the pool of these events.

## 2.3 Close

The pool supports closing all sessions by `close()`. If a closed connection is still involved in
I/O, it will trigger the disconnect callback, which in turn triggers the on_disconnected call.

# 3. Implementation

The pool size—i.e., the maximum number of sessions it can hold—is configurable via
`conn_pool_max_size`.

When an external caller requests a connection from the pool, if the pool has not yet reached
its limit, a new connection will be created and returned. Otherwise, an existing connection
will be selected and returned.

Each connection may have multiple ongoing RPCs (or none at all). The selection strategy
for existing connections is to pick the one with the fewest in-flight RPCs.

# 4. Changes

Compared to before, the main change is moving from a single client connection to a client
connection pool. This change is almost seamless—code that previously fetched a single
connection now simply fetches one from the pool. Similarly, where the previous single
connection had its own callbacks (for connection established and disconnected) and close
logic, the corresponding pool interfaces should now be used.

Additionally, this PR includes optimizations and refactoring for much of the connection-oriented
networking and RPC session (connection) code—without changing their behavior. Two new
configurations, `socket_send_buffer_bytes` and `socket_recv_buffer_bytes`, have been
added to control the send/receive buffer size for each socket.

Following configurations are newly added:
```diff
[network]
+ conn_pool_max_size = 4
+ socket_send_buffer_bytes = 16777216
+ socket_recv_buffer_bytes = 16777216
```

### DIFF
--- a/src/rpc/asio_net_provider.cpp
+++ b/src/rpc/asio_net_provider.cpp
@@ -127,6 +127,9 @@ error_code asio_network_provider::start(rpc_channel channel, int port, bool clie
         return ERR_SERVICE_ALREADY_RUNNING;
     }
 
+    // The start() function is designed to support multiple invocations (see what have been
+    // done in AsioNetProvider for NetProviderTest), so it is necessary to ensure that `_workers`
+    // does not exceed the limit defined by io_service_worker_count.
     for (auto i = _workers.size(); _workers.size() < FLAGS_io_service_worker_count; ++i) {
         _workers.push_back(std::make_shared<std::thread>([this, i]() {
             task::set_tls_dsn_context(node(), nullptr);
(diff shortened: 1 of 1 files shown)


empiredan:
> Could you please describe the changes and design simply?

@acelyc111 Sorry for the missing information about this PR, which now have been added to the description.

acelyc111 (approved):

PR #2279, opened 2025-08-05
build(deps): bump both Thrift compiler and org.apache.thrift:libthrift from 0.11.0 to 0.13.0 in /java-client

empiredan:


acelyc111:
@empiredan Have you check the compatibility? use old version client to access the newer version of servers.

empiredan:
On the other hand, I also reviewed the differences between the code generated by Thrift 0.11.0 and 0.13.0 and found only two changes: one is the introduction of the `org.apache.thrift.annotation.Nullable` annotation (see https://github.com/apache/thrift/pull/1522), and the other is the renaming of an internal method from `handleRuntimeExceptions` to `rethrowUnhandledExceptions` (see https://github.com/apache/thrift/pull/1544). Neither of these changes affects the Java client code that depends on the generated Thrift code. Therefore, the compatibility tests are mainly focused on the Thrift library itself.

acelyc111 (approved):

Samunroyu (approved):

PR #2285, opened 2025-08-13
feat: Introduced Prometheus and Falcon for monitoring and added multiple metric collection sites

nanorth:
##### Tests 

- Unit test
- Integration test
- No code

##### Code changes

- Has persistent data change

##### Side effects

- Increased code complexity

##### Related changes


go-client/config/config.go
```diff
@@ -17,9 +17,15 @@
...
 // Config is the configuration of pegasus client.
 type Config struct {
-	MetaServers []string `json:"meta_servers"`
+	MetaServers      []string          `json:"meta_servers"`
+	PerfCounterTags  map[string]string `json:"perf_counter_tags"`
+	EnableFalcon     bool              `json:"enable_falcon"`
+	EnablePrometheus bool              `json:"enable_prometheus"`
+	FalconServer     string            `json:"falcon_server"`
+	FalconInterval   int               `json:"falcon_interval_secs"`
+	PrometheusPort   int               `json:"prometheus_port"`
```
lengyuexuexuan, line 30: Is a default config necessary?

go-client/pegasus/client.go
```diff
@@ -59,17 +63,25 @@ func NewClient(cfg Config) Client {
...
 	if err != nil {
 		return nil, err
 	}
 
+	if cfg.EnablePrometheus || cfg.EnableFalcon {
+		metrics.InitMetrics(prometheus.DefaultRegisterer, cfg)
+		if cfg.EnableFalcon {
+			go metrics.GetFalconReporter(cfg.FalconServer, cfg.FalconInterval).Start()
+		}
+	}
```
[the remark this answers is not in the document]
nanorth: The Prometheus register will be initialized if either Prometheus or Falcon is enabled. When `cfg.EnablePrometheus` is false and `cfg.EnableFalcon` is true, the Prometheus register is still initialized so that the Falcon reporter can collect metrics from it and report. `cfg.EnablePrometheus` only ensures Prometheus to exposing its port. 

```
func InitMetrics(registry prometheus.Registerer, cfg config.Config) {
	initRegistry = registry
	perfCounterMap = cfg.PerfCounterTags
	if cfg.EnablePrometheus {
		startServerOnce.Do(func() {
			port := 9090
			if cfg.PrometheusPort > 0 {
				port = cfg.PrometheusPort
			}
			go func() {
				http.Handle("/metrics", promhttp.Handler())
				addr := fmt.Sprintf(":%d", port)
				pegalog.GetLogger().Print("Starting Prometheus metrics server on", addr)
				if err := http.ListenAndServe(addr, nil); err != nil {
					pegalog.GetLogger().Fatal("Failed to start Prometheus metrics server:", err)
				}
			}()
		})
	}
}
```

go-client/config/config.go
nanorth: Since `MetaServers` is a required configuration, and most users are likely to initialize it from a configuration file, the relevant default values have already been set in the JSON example, the sample configuration file may serve as a default configuration.

go-client/pegasus/table_connector.go
```diff
@@ -715,8 +719,35 @@ func (p *pegasusTableConnector) Incr(ctx context.Context, hashKey []byte, sortKe
...
+			}
+
+			// The metaIP is added to the metric name because some users may use multiple client instances
+			// within a single process to access tables with the same name in different availability zones.
+			// Including the metaIP helps to distinguish monitoring metrics for tables with the same name.
+			// The reason for not putting metaIP into labels (Prometheus) or tags (Falcon) is that labels/tags
+			// are designed to be unique and constant for a single process.
+			firstMetaIP := strings.ReplaceAll(p.meta.GetMetaIPAddrs()[0], ".", "_")
+			pm.MarkMeter(fmt.Sprintf("pegasus_client_%s_%s_%s_total_%s", p.tableName, optype.String(), status, firstMetaIP), 1)
+			elapsed := time.Since(start).Nanoseconds()
```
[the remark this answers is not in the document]
nanorth: Yes, agreed. Code updated

go-client/session/replica_session.go
```diff
@@ -188,6 +188,8 @@ type ReplicaManager struct {
 	creator NodeSessionCreator
 
 	unresponsiveHandler UnresponsiveHandler
+
+	enablePerfCounter bool
```
acelyc111, line 192: Maybe `Metrics` is more general than `PerfCounter`?

go-client/session/meta_call.go
```diff
@@ -141,7 +141,7 @@ func (c *metaCall) issueSingleMeta(ctx context.Context, curLeader int) bool {
 			c.lock.Lock()
 			c.metaIPAddrs = append(c.metaIPAddrs, addr)
 			c.metas = append(c.metas, &metaSession{
-				NodeSession: newNodeSession(addr, NodeTypeMeta),
+				NodeSession: newNodeSession(addr, NodeTypeMeta, false),
```
acelyc111, line 144: The bool type true/false parameter is meaningless, how about using a `const` to define enable or disable metrics?

go-client/pegasus/table_connector.go
```diff
@@ -715,8 +719,29 @@ func (p *pegasusTableConnector) Incr(ctx context.Context, hashKey []byte, sortKe
...
+			status := "success"
+			if errResult != nil {
+				if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+					status = "timeout"
+				} else {
+					status = "fail"
+				}
+			}
+			meta := strings.Join(p.meta.GetMetaIPAddrs(), ",")
+			pm.MarkMeter(fmt.Sprintf("pegasus_client_%s_%s_%s_total", p.tableName, optype.String(), status), 1, map[string]string{"meta": meta})
```
acelyc111, line 737: How about add more labels rather than name components? the labels in Prometheus is more flexible, we can filter or aggregation metrics by labels.

acelyc111 (approved):
@nanorth Thanks for the contribution!

Issue #2297, opened 2025-09-19
Bug(scripts): Running pegasus_manual_compact.sh on a very small table will hang indefinitely.


limowang:
1. What did you do?
sh {script_cmd} --config {shell_config_file_path} -a {table_name} --bottommost_level_compaction {bottommost_level_compaction} 

2. What did you expect to see?
It can finish normally after a period of time.
3. What did you see instead?
The operation is stuck and never finishes.

PR #2298, opened 2025-09-19
fix(scripts): Fix the issue where running pegasus_manual_compact.sh on small tables sometimes hangs

limowang:
[#2297](https://github.com/apache/incubator-pegasus/issues/2297)

### DIFF
--- a/admin_tools/pegasus_manual_compact.sh
+++ b/admin_tools/pegasus_manual_compact.sh
@@ -132,7 +132,7 @@ function wait_manual_compact()
     total_replica_count=$3
 
     query_cmd="remote_command -t replica-server replica.query-compact ${app_id}"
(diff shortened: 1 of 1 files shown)


acelyc111 (approved):

acelyc111:
> This pull request makes a minor adjustment to the `wait_manual_compact()` function in `admin_tools/pegasus_manual_compact.sh`, specifically changing the timestamp format used when displaying the earliest finish time for a compact operation.
> 
> * Changed the output format of `earliest_finish_time_ms` to exclude milliseconds when displaying the compact progress timestamp.

Sumaried by Copilot
image

empiredan:
The root cause is still comparing timestamps as strings, right? Would it be safer to convert the strings to integers for comparison during the later refactoring?

empiredan (approved):

PR #2300, opened 2025-09-24
fix(shell): Fixed  the shell couldn't connect to the onebox cluster after onebox launch in a Docker environment

Samunroyu:
The shell couldn't connect to onebox cluster because a configuration parameter was missing.And the shell can't resolve the right hostname.

image
image

### DIFF
--- a/src/test/function_test/config.ini
+++ b/src/test/function_test/config.ini
@@ -51,6 +51,7 @@ random_seed = 0
 [network]
 ; how many network threads for network library(used by asio)
 io_service_worker_count = 4
+primary_interface = @LOCAL_HOSTNAME@
 
 ; specification for each thread pool
 [threadpool..default]


src/test/function_test/config.ini
```diff
@@ -51,6 +51,7 @@ random_seed = 0
 [network]
 ; how many network threads for network library(used by asio)
 io_service_worker_count = 4
+primary_interface = @LOCAL_HOSTNAME@
```
acelyc111, line 54: `primary_interface` is considered as a network interface name, e.g. `eth0`, `lo`, why assign it a hostname?

Samunroyu:
> From the printed logs, it looks like the program successfully detected the `eth0` network interface and its corresponding IP address `10.206.0.16`, but failed to resolve its corresponding hostname.
> 
> Would adding a mapping in `/etc/hosts` fix this issue?

i checked carefully. The `/etc/hosts` do have the mapping of address and hostname.
However i started docker in '--network=host' mode, so the `eth0` return the Host Machine ip address, so the lookup_hostname failed.

PR #2302, opened 2025-09-25
fix(zookeeper): fix potential dangling path pointer issue when calling ZooKeeper APIs

empiredan:
Fix https://github.com/apache/incubator-pegasus/issues/2307.

The heap-use-after-free error found in the ASan test was caused by a dangling
pointer issue: the `path` pointer was freed in the thread for completion callback
while it was still used in the ZooKeeper C client API.

To fix this problem, `std::shared_ptr` is introduced to manage the `path`, ensuring
that the memory is automatically released only when all references to it have been
destroyed.

### DIFF
--- a/src/zookeeper/zookeeper_session.cpp
+++ b/src/zookeeper/zookeeper_session.cpp
@@ -402,7 +402,12 @@ void zookeeper_session::visit(zoo_opcontext *ctx)
     // TODO(clang-tidy): the read ops from zookeeper might get the staled data, need to fix
     int ec = ZOK;
     zoo_input &input = ctx->_input;
-    const auto path = input._path;
+
+    // A local variable is needed here to hold a reference to `_path` to prevent it from
+    // being freed in the thread executing the completion callback, which could otherwise
+    // lead to a dangling pointer issue.
+    const std::shared_ptr<std::string> path(input._path);
+
     switch (ctx->_optype) {
     case ZOO_CREATE:
         ec = zoo_acreate(_handle,
--- a/src/meta/meta_state_service.h
+++ b/src/meta/meta_state_service.h
@@ -120,7 +120,7 @@ public:
                                  const blob &value,
                                  dsn::task_tracker *tracker) = 0;
 
-    // The same as create_node(), except that the value is empty and the tracker is nullptr.
+    // The same as create_node(), except that `value` is empty and `tracker` is nullptr.
     task_ptr
     create_empty_node(const std::string &node, task_code cb_code, const err_callback &cb_create)
     {


src/meta/test/meta_state/meta_state_service.cpp
```diff
@@ -61,87 +62,112 @@ void provider_basic_test(const service_creator_func &service_creator,
                          const service_deleter_func &service_deleter)
 {
     // environment
-    auto service = service_creator();
+    auto *service = service_creator();
 
     // bondary check
-    service->node_exist("/", META_STATE_SERVICE_SIMPLE_TEST_CALLBACK, expect_ok)->wait();
```
acelyc111: Is it possible to keep use a default parameter?

src/zookeeper/lock_struct.cpp
```diff
@@ -637,20 +644,23 @@ void lock_struct::create_locknode()
...
+        enqueue([_this, path]() { lock_struct::after_create_locknode(_this, ZOK, path); }, _this);
     };
 
     zookeeper_session::zoo_input &input = op->_input;
-    input._path = _lock_dir + "/" + distributed_lock_service_zookeeper::LOCK_NODE_PREFIX;
+    input._path = std::make_shared<std::string>(
+        fmt::format("{}/{}", _lock_dir, distributed_lock_service_zookeeper::LOCK_NODE_PREFIX));
     input._value.assign(_myself._node_value.c_str(), 0, _myself._node_value.length());
-    input._flags = ZOO_EPHEMERAL | ZOO_SEQUENCE;
+    input._flags = ZOO_EPHEMERAL_SEQUENTIAL;
```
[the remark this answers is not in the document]
empiredan: For the following reasons, we must use `ZOO_EPHEMERAL_SEQUENTIAL` instead of `ZOO_EPHEMERAL | ZOO_SEQUENCE`:

* We rely on ZooKeeper C client **3.9.4**, which defines `ZOO_EPHEMERAL_SEQUENTIAL` as a replacement for `ZOO_EPHEMERAL | ZOO_SEQUENCE`.
* `ZOO_SEQUENCE` has already been **deprecated** — if it were to be used, it should be replaced by `ZOO_PERSISTENT_SEQUENTIAL`.
* `ZOO_EPHEMERAL | ZOO_SEQUENCE` cannot pass the linter check; **clang-tidy** reports the following warning:
  `warning: use of a signed integer operand with a binary bitwise operator [hicpp-signed-bitwise] input._flags = ZOO_EPHEMERAL | ZOO_SEQUENCE;`

src/meta/test/meta_state/meta_state_service.cpp
empiredan: `node_exist` here is a **virtual function** that implements a pure virtual function declared in `meta_state_service`.
However, having **default arguments** in a virtual function violates the linter rule — `clang-tidy` check would fail due to the following error: `warning: default arguments on virtual or override methods are prohibited [google-default-arguments]`.

src/zookeeper/zookeeper_session.cpp
```diff
@@ -398,58 +399,64 @@ void zookeeper_session::visit(zoo_opcontext *ctx)
...
-    // TODO: the read ops from zookeeper might get the staled data, need to fix
+    // TODO(clang-tidy): the read ops from zookeeper might get the staled data, need to fix
     int ec = ZOK;
     zoo_input &input = ctx->_input;
-    const char *path = input._path.c_str();
+
+    // A local variable is needed here to hold a reference to `_path` to prevent it from
+    // being freed in the thread executing the completion callback, which could otherwise
+    // lead to a dangling pointer issue.
+    const std::shared_ptr<std::string> path(input._path);
```
acelyc111: Is this the key changes of this PR?

acelyc111 (approved):

src/zookeeper/zookeeper_session.cpp
empiredan: Yes !

src/meta/test/meta_state/meta_state_service.cpp
empiredan: Indeed. At first, I didn’t plan to modify it — I thought it would be fine to keep the default argument.
However, since the virtual function in the `meta_state_service` class isn’t called in many places and can be easily updated, I decided to follow `clang-tidy`’s suggestion and make the change.

acelyc111 (approved):

PR #2303, opened 2025-10-15
fix(Python-client):prevent logger override and migrate config to YAML

WJSGDBZ:

By default, `logging.config.fileConfig` disables existing loggers if disable_existing_loggers is not set explicitly, causing previously configured loggers to be lost.
https://github.com/apache/incubator-pegasus/blob/fdbd6ebb460b97d39c027837217c1a41de1b3671/python-client/pypegasus/pgclient.py#L47-L48

We initially fixed this by setting `disable_existing_loggers=False`. However, in Python 3.7 and earlier, fileConfig still requires a [logger_root] section, which may override the existing root logger and alter logging behavior.

To avoid this, we now use `logging.config.dictConfig`, which supports disable_existing_loggers and does not require a root logger configuration, offering safer and more flexible logging setup.

1. Changed the configuration format from .conf to YAML.
2. Replaced logging.config.fileConfig with logging.config.dictConfig for log initialization.

##### Tests 

1. We created a custom logger app_logger before initializing the Pegasus logging system:
2. Then, import Pegasus
```
app_logger = logging.getLogger("app")
app_logger.setLevel(logging.INFO)
console_handler = logging.StreamHandler()
console_handler.setLevel(logging.INFO)
console_formatter = logging.Formatter("[APP] %(levelname)s: %(message)s")
console_handler.setFormatter(console_formatter)
app_logger.addHandler(console_handler)

from pypegasus.pgclient import Pegasus

if __name__ == "__main__":
    app_logger.info("This is an app log message")
```
In the original code, the app_logger configuration was lost after Pegasus initialized its logging via fileConfig. After switching to dictConfig, the app_logger remains intact and logs successfully, showing that the fix preserves existing loggers as intended.

### DIFF
--- a/python-client/pypegasus/logger.yaml
+++ b/python-client/pypegasus/logger.yaml
@@ -41,3 +41,4 @@ loggers:
     level: INFO
     handlers: [pegasus_file_handler]
     propagate: false
+    


acelyc111 (approved):
@WJSGDBZ Thanks for the contribution!

PR #2317, opened 2025-11-10
fix(cluster_balance_policy): clear partitions in migration info before outputting result into it

empiredan:


Samunroyu:
Why do we need to clear partitions in migration info ?Is the old partitions info will appear on the outputting result if we dont clear partition?

acelyc111 (approved):

PR #2341, opened 2025-12-23
fix(test): fix the failed ASan test for gutil due to absl::node_hash_map

empiredan:
https://github.com/apache/incubator-pegasus/issues/2344.

The internal types used by `absl::node_hash_map` have different layouts when
compiled with ASan versus non-ASan builds. As a third-party dependency, Abseil
is always built in non-ASan mode. However, when Pegasus is built with ASan enabled,
Abseil headers are also compiled under ASan, which leads to a mismatch between the
header-compiled structures and the Abseil library itself, causing ASAN tests to fail.

Possible solutions include:
1. When third-party dependencies change, also build those third-party libraries with
ASan enabled for ASan builds.
2. When building third-party images, additionally produce a separate ASan-based
image, which can be used by ASan CI workflows when third-party dependencies
themselves have not changed.

### DIFF
--- a/scala-client/build.sbt
+++ b/scala-client/build.sbt
@@ -34,9 +34,7 @@ scalafmtOnCompile := true
 //custom repository
 resolvers ++= Seq(
   //"Remote Maven Repository" at "http://your-url/",
-  "Local Maven Repository" at "file://" + Path.userHome.absolutePath + "/.m2/repository",
-  Resolver.mavenCentral,
-  "Sonatype OSS" at "https://oss.sonatype.org/content/repositories/releases"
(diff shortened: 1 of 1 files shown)


Samunroyu (commented):
So this issue is because we didn't distinguish between Asan and non-Asan modes when compiling Abseil, rather than being a problem with Abseil itself?

empiredan:
> So this issue is because we didn't distinguish between Asan and non-Asan modes when compiling Abseil, rather than being a problem with Abseil itself?

Yes, you can say that. It’s not that Abseil itself has a bug, although having different class memory layouts between ASAN and non-ASAN builds is not a recommended practice.

Samunroyu (approved):

foreverneverer (approved):

Issue #2211, opened 2025-03-14
bug: use "dup remove" command stop duplication, cluster may coredump sometimes


ninsmiracle:
## Bug Report
1. What did you do?
Use "dup remove" command stop a duplication, and all the nodes of the cluster coredump.

one node has this coredump info, and the coredump information of other nodes is very confusing.
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/home/work/app/pegasus/c3srv-xxx/replica/package/bin/pegasus_server con'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  SLL_Next (t=0x0) at src/linked_list.h:45
45      src/linked_list.h: No such file or directory.
(gdb) bt
#0  SLL_Next (t=0x0) at src/linked_list.h:45
#1  SLL_PopRange (end=<synthetic pointer>, start=<synthetic pointer>, N=32, head=0x2085a00) at src/linked_list.h:88
#2  PopRange (end=<synthetic pointer>, start=<synthetic pointer>, N=32, this=0x2085a00) at src/thread_cache.h:238
#3  tcmalloc::ThreadCache::ReleaseToCentralCache (this=this@entry=0x20859c0, src=src@entry=0x2085a00, cl=cl@entry=2, N=89) at src/thread_cache.cc:201
#4  0x00007fd342586677 in tcmalloc::ThreadCache::Scavenge (this=0x20859c0) at src/thread_cache.cc:224
#5  0x00007fd3410c363e in std::string::reserve(unsigned long) () from /home/work/app/pegasus/c3srv-feedprofile/replica/package/bin/libstdc++.so.6
#6  0x00007fd3410c3abf in std::string::append(std::string const&) () from /home/work/app/pegasus/c3srv-feedprofile/replica/package/bin/libstdc++.so.6
#7  0x000000000066ef2f in std::basic_string<char, std::char_traits<char>, std::allocator<char> > std::operator+<char, std::char_traits<char>, std::allocator<char> >(std::basic_string<char, std::char_traits<char>, std::allocator<char> >&&, std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
#8  0x000000000080f129 in rocksdb::ParseFileName(std::string const&, unsigned long*, rocksdb::Slice const&, rocksdb::FileType*, rocksdb::WalFileType*) ()
[...]

PR #2368, opened 2026-02-12
fix(duplication): prevent potential crash when removing or pausing duplication

lengyuexuexuan:
#2211 

1. Background   
    During duplication of a table, if the commands `dup remove/pause` are executed or `a balance operation` is performed at the same time, there is a chance that a node may core dump with signal `ID 11`. The core dump locations vary, but they all have one thing in common: they occur during memory allocation or deallocation.
2. Analysis 
    Based on extensive testing, the following conclusions can be drawn:
    a. The issue only reproduces when there is write traffic. The difference between having and not having write traffic is: It adds the ship and load_private_log tasks.  
    b. The core dump occurs during the execution of `cancel_all()`.
    c. The issue occurs with low probability (approximately 1 in 100).
   
    Through analysis using ASAN (AddressSanitizer): 
[dup_remove_asan.txt](https://github.com/user-attachments/files/19956043/dup_remove_asan.txt)
    Based on ASAN analysis, the following conclusions can be drawn:
    a. The memory corruption occurs during the ship process. The mutations obtained from replaying the plog are passed to ship, leading to the issue.
    b. `_load_mutations` is captured by a lambda expression and then passed to a `std::function`. Since `std::move` is used, the lifetime of `_load_mutations` is tied to that of the `std::function`.
    c. The `cancel_all()` function is executed in the default thread pool. At this point, the following function is called. When the `std::function` is set to nullptr, it will release the memory it manages.https://github.com/apache/incubator-pegasus/blob/e64faa7d03f836e5212874bbbfb8f900341897d7/src/task/task.h#L341
    d. However, each task executes `exec_internal()` in its own thread pool, and eventually calls `release_ref()`, which results in delete this. https://github.com/apache/incubator-pegasus/blob/e64faa7d03f836e5212874bbbfb8f900341897d7/src/task/task.cpp#L224

3. Conclusion  
    **1. Both `task.cancel()` and `task.exec_internal()` destruct the std::function member inside the task object. These two operations are executed in different threads, and there is no mechanism in place to prevent race conditions between them. As a result, it is possible for both threads to attempt to destruct the same std::function, which can lead to a double deletion of the memory associated with _load_mutations. This ultimately causes memory corruption.
    2. `_duplications` is accessed without proper synchronization in certain functions under multi-threaded scenarios, potentially causing race conditions.**

4. Solution  
   1.  Lock the `_cb` callback to ensure that only one thread executes its destructor.
   2. Add locking to functions that access `_duplications` without synchronization to prevent concurrent access conflicts.
  
##### Tests 
- Manual test 
**The changes have been production-validated at Xiaomi, running stably on more than 30 clusters for over six months, confirming that they resolve the concurrency issues described above.**

### DIFF
--- a/src/replica/duplication/replica_duplicator_manager.cpp
+++ b/src/replica/duplication/replica_duplicator_manager.cpp
@@ -196,7 +196,7 @@ void replica_duplicator_manager::update_confirmed_decree_if_secondary(decree con
 void replica_duplicator_manager::METRIC_FUNC_NAME_SET(dup_pending_mutations)()
 {
     zauto_lock l(_lock);
-    int64_t total = 0;
+    uint64_t total = 0;
     for (const auto &dup : _duplications) {
         total += dup.second->get_pending_mutations_count();
     }


empiredan:
Hi @lengyuexuexuan Thank you for your contribution!

Please modify the code according to the suggestions provided by Clang Tidy and IWYU.

lengyuexuexuan:
> Hi @lengyuexuexuan Thank you for your contribution!
> 
> Please modify the code according to the suggestions provided by Clang Tidy and IWYU.

done

PR #2374, opened 2026-02-13
fix: check if db exists in update_app_envs_before_open_db when starting a replica

nanorth:
Handle the `manual_compact.once.trigger_time` parameter in app_envs provided by duplication. When a backup cluster creates an app via duplication, incomplete replicas may trigger manual compaction if the source app has this parameter configured. However, these replicas have not yet opened RocksDB successfully, which can lead to coredumps.


src/server/pegasus_server_impl.cpp
```diff
@@ -3572,4 +3579,4 @@ dsn::replication::manual_compaction_status::type pegasus_server_impl::query_comp
     return _manual_compact_svc.query_compact_status();
 }
 
-} // namespace pegasus::server
+} // namespace pegasus::server
```
empiredan: The warning here appears to be caused by a missing newline at the end of the file. Please restore the trailing newline.

Issue #2379, opened 2026-03-09
feat: support Duplication to send non-idempotent write operations


ruojieranyishen:
## Feature Request

**Is your feature request related to a problem? Please describe:**
For some tables with duplication enabled, blocking non-idempotent writes is not strictly required in production.

Duplication does not support non-idempotent write operations (INCR, CHECK_AND_SET, CHECK_AND_MUTATE)
In `mutation_batch::add_mutation_if_valid`, all non-idempotent writes are skipped:
```
if (!task_spec::get(update.code)->rpc_request_is_write_idempotent) {
    continue;
}
```
In replica::need_reject_non_idempotent, non-idempotent writes are rejected:
```
return !spec->rpc_request_is_write_idempotent;
```

**Describe the feature you'd like:**
Support Duplication to send non-idempotent write operations (INCR, CHECK_AND_SET, CHECK_AND_MUTATE).

PR #2380, opened 2026-03-10
feat(duplication): allow non-idempotent writes during duplication

ruojieranyishen:
https://github.com/apache/incubator-pegasus/issues/2379.

This PR introduces a new configuration option `duplication_unsafe_allow_non_idempotent` to enable non-idempotent write operations (incr, check_and_set, check_and_mutate) during cluster duplication. Previously, these operations were rejected when duplication was enabled to ensure data consistency. However, production environments require flexibility to allow such writes with proper monitoring.

The change adds:
- New config flag `duplication_unsafe_allow_non_idempotent`
- Support for handling non-idempotent RPCs in mutation duplicator
- New metrics: `dup_unsafe_received_non_idempotent_duplicate_request` and
  `dup_retry_non_idempotent_duplicate_request`

### DIFF
--- a/src/server/pegasus_mutation_duplicator.cpp
+++ b/src/server/pegasus_mutation_duplicator.cpp
@@ -270,6 +270,12 @@ void pegasus_mutation_duplicator::log_non_idempotent_rpc_retry_if_need(duplicate
 
         METRIC_VAR_INCREMENT(dup_retry_non_idempotent_duplicate_request);
 
+        // Randomly log 1% of the non-idempotent duplicate RPC retries to avoid
+        // excessive log volume under high QPS.
+        if (dsn::rand::next_double01() > 0.01) {
+            continue;
+        }
+
         dsn::message_ex *write = dsn::from_blob_to_received_msg(entry.task_code, entry.raw_message);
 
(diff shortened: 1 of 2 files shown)


src/server/pegasus_mutation_duplicator.cpp
```diff
@@ -234,6 +255,48 @@ void pegasus_mutation_duplicator::on_duplicate_reply(uint64_t hash,
     }
 }
 
+void pegasus_mutation_duplicator::log_non_idempotent_rpc_retry_if_need(duplicate_rpc &rpc)
+{
+    if (!FLAGS_duplication_unsafe_allow_non_idempotent) {
+        return;
+    }
+
+    for (const auto &entry : rpc.request().entries) {
```
lengyuexuexuan, line 264: If we log every time, the log volume could become very high when the QPS of non-idempotent operations is significant. Could we switch to probabilistic  logging instead? 
Like https://github.com/apache/incubator-pegasus/blob/89f17c7722af064d326c0dbbdb147f4d1b207ee7/src/server/pegasus_mutation_duplicator.cpp#L202-L206

lengyuexuexuan:
Review done.

ruojieranyishen:
@empiredan please review this pr, thinks

Issue #2386, opened 2026-03-16
bug: Misleading error "The request maybe too large" when secondary replica disk is abnormal


limowang:
Description

The java-client logs output "The request maybe too large!".  
After checking the `handleReplicaException(Request request, DefaultPromise promise, client_operator op, Table table, int timeout)` function in `PegasusTable.java`, it is found that this message is only logged when `ERR_INVALID_DATA` is received.

Then, looking at the server-side code in `replica_2pc.cpp`, specifically the `on_client_write(dsn::message_ex *request, bool ignore_throttling)` function, it is observed that when the server returns `ERR_INVALID_DATA` to the client, it also logs the following message:

"client from {} write request body size exceed threshold, request = [{}], request_body_size = {}, FLAGS_max_allowed_write_size = {}, it will be rejected!"

However, when I run the command:

grep -r "request_body_size" log_directory_path

in the log directory, there is no output.

PR #2387, opened 2026-03-17
fix(replica): fix incorrect error code when secondary replica disk status is abnormal

limowang:
Fix [#2386](https://github.com/apache/incubator-pegasus/issues/2386)

 When the primary replica’s disk status is normal but the secondary replica encounters disk issues (IO_ERROR or SPACE_INSUFFICIENT), the primary correctly rejects write requests. However, the Java client logs a misleading error message: “The request maybe too large!” instead of the actual disk error, which makes troubleshooting more difficult.

  Root Cause

  Bug 1: When the primary replica’s disk status is normal but the secondary disk is abnormal, C++ returns ERR_OK.

  Location: src/replica/replica_2pc.cpp:184-188

  if (FLAGS_reject_write_when_disk_insufficient &&
      (_dir_node->status != disk_status::NORMAL || _primary_states.secondary_disk_abnormal())) {
      response_client_write(request, disk_status_to_error_code(_dir_node->status));
      // BUG: When secondary is abnormal but primary is NORMAL, this returns ERR_OK!
      return;
  }

  Bug 2: Java maps empty response to ERR_INVALID_DATA

  Location: java-client/src/main/java/org/apache/pegasus/rpc/async/ThriftFrameDecoder.java:70-81

  When C++ returns ERR_OK with no response body (because the write request was rejected), the Java client's recv_data() throws a TException, which is then incorrectly mapped to ERR_INVALID_DATA, resulting in misleading log output.

Solution:
1. Modified `replica_2pc.cpp` to return more accurate error codes:
   - When the primary node’s disk has issues, it returns the corresponding disk status code.
   - When the primary node’s disk is normal but the secondary replica’s disk is abnormal (`ERR_DISK_IO_ERROR` or `ERR_DISK_INSUFFICIENT`), it no longer incorrectly returns `ERR_OK`, but instead returns the corresponding disk error code of the secondary replica node.

2. Updated the Java client to handle `ERR_DISK_IO_ERROR` error codes.

### DIFF
--- a/collector/config.yml
+++ b/collector/config.yml
@@ -15,8 +15,16 @@
 # specific language governing permissions and limitations
 # under the License.
 
+# logging configurations.
+log:
+  filename: collector.log
+  max_file_size_mb: 64
+  retention_days: 3
+  max_file_number: 8
+  level: info
+
 # the cluster that this collector is binding
-cluster_name : "onebox"
+cluster_name: "onebox"
 
 # the meta server addresses of the cluster.
 meta_servers:
@@ -25,33 +33,34 @@ meta_servers:
   - 127.0.0.1:34603
 
 # local server port
-port : 34101
+port: 34101
 
 metrics:
   # use falcon as monitoring system.
-  sink : falcon
-  report_interval : 10s
+  sink: falcon
+  report_interval: 10s
 
 prometheus:
   # the exposed port for prometheus exposer
-  exposer_port : 1111
+  exposer_port: 1111
 
 falcon_agent:
   # the host IP of falcon agent
-  host : "127.0.0.1"
-  port : 1988
-  http_path : "/v1/push"
+  host: "127.0.0.1"
+  port: 1988
+  http_path: "/v1/push"
 
 availability_detect:
-  table_name : test
-  partition_count : 16
-  max_replica_count : 3
+  table_name: test
+  partition_count: 16
+  max_replica_count: 3
 
 hotspot:
-  rpc_timeout : 5s
-  partition_detect_interval : 30s
-  pull_metrics_timeout : 5s
-  sample_metrics_interval : 10s
-  max_sample_size : 128
+  retention_period: 24h
+  rpc_timeout: 5s
+  partition_detect_interval: 30s
+  pull_metrics_timeout: 5s
+  sample_metrics_interval: 10s
+  max_sample_size: 128
   hotspot_partition_min_score: 3
   hotspot_partition_min_qps: 100
--- a/collector/go.mod
+++ b/collector/go.mod
@@ -20,7 +20,7 @@ module github.com/apache/incubator-pegasus/collector
 go 1.18
 
 require (
-	github.com/apache/incubator-pegasus/go-client v0.0.0-20260121121155-96868ed93b2a
+	github.com/apache/incubator-pegasus/go-client v0.0.0-20260211095029-022854b0259f
 	github.com/gammazero/deque v1.0.0
 	github.com/kataras/iris/v12 v12.2.0
 	github.com/prometheus/client_golang v1.18.0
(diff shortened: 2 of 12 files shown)


src/replica/replica_2pc.cpp
```diff
@@ -183,7 +183,13 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
 
     if (FLAGS_reject_write_when_disk_insufficient &&
         (_dir_node->status != disk_status::NORMAL || _primary_states.secondary_disk_abnormal())) {
-        response_client_write(request, disk_status_to_error_code(_dir_node->status));
+        if (_dir_node->status != disk_status::NORMAL) {
+            // Primary replica disk is abnormal, return the corresponding error code
+            response_client_write(request, disk_status_to_error_code(_dir_node->status));
+        } else {
+            // Secondary replica disk is abnormal but primary is OK
+            response_client_write(request, ERR_REPLICATION_FAILURE);
```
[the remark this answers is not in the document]
empiredan: I tend to prefer returning `ERR_DISK_INSUFFICIENT` or `ERR_DISK_IO_ERROR` for disk issues on secondary replicas as well. The reasons are as follows:

1. Whether primary or secondary, they are essentially replicas of the data. From a high-level perspective, any disk issue on a replica results in a failure to write data.
2. A two-dimensional classification is introduced if disk issues on primary and secondary replicas are represented by different error codes, which will bloat the error code system.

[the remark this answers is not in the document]
limowang: Your suggestion is very reasonable, and I will make the changes accordingly.

src/replica/replica_2pc.cpp
```diff
@@ -183,7 +183,18 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
 
     if (FLAGS_reject_write_when_disk_insufficient &&
         (_dir_node->status != disk_status::NORMAL || _primary_states.secondary_disk_abnormal())) {
```
empiredan:
```suggestion
    if (FLAGS_reject_write_when_disk_insufficient) {
```

src/replica/replica_2pc.cpp
```diff
@@ -183,7 +183,18 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
...
+            // Primary replica disk is abnormal, return the corresponding error code
+            response_client_write(request, disk_status_to_error_code(_dir_node->status));
+        } else {
+            // Secondary replica disk is abnormal but primary is OK
+            for (const auto &kv : _primary_states.secondary_disk_status) {
+                if (kv.second != disk_status::NORMAL) {
+                    response_client_write(request, disk_status_to_error_code(kv.second));
+                    break;
+                }
+            }
```
empiredan:
```suggestion
            for (const auto &[addr, secondary_status] : _primary_states.secondary_disk_status) {
                if (secondary_status != disk_status::NORMAL) {
                    LOG_INFO("partition[{}] secondary[{}] disk space is {}",
                                        _primary_states.pc.pid,
                                        addr,
                                        enum_to_string(secondary_status));                    
                    response_client_write(request, disk_status_to_error_code(secondary_status));
                    return;
                }
            }
```

src/replica/replica_2pc.cpp
```diff
@@ -183,7 +183,18 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
...
+        }
         return;
```
empiredan:
```suggestion
```

src/replica/replica_2pc.cpp
```diff
@@ -183,7 +183,18 @@ void replica::on_client_write(dsn::message_ex *request, bool ignore_throttling)
 
     if (FLAGS_reject_write_when_disk_insufficient &&
         (_dir_node->status != disk_status::NORMAL || _primary_states.secondary_disk_abnormal())) {
-        response_client_write(request, disk_status_to_error_code(_dir_node->status));
+        if (_dir_node->status != disk_status::NORMAL) {
+            // Primary replica disk is abnormal, return the corresponding error code
+            response_client_write(request, disk_status_to_error_code(_dir_node->status));
```
empiredan, line 187:
```suggestion
            response_client_write(request, disk_status_to_error_code(_dir_node->status));
            return;
```

empiredan (approved):

PR #2401, opened 2026-05-26
ci(github): fix branch glob to actually match release branches

acelyc111:
## Summary

The branch filter `'v[0-9]+.*'` used in 15 workflow files **never matches any release branch**, because GitHub Actions branch filters use glob syntax (not regex):

- `+` is a **literal character**, not a quantifier
- `[0-9]` matches **exactly one** digit, not "one or more"

As a result, PRs targeting release branches like `v2.5` silently skip every workflow that has this filter, while only those without `branches:` filters (e.g., the Golang/Labeler ones) actually run.

## Evidence

**PR #2394** targets `v2.5` and modifies `src/replica/duplication/**` (C++ code):

- ✅ Triggered: `Golang Lint and Unit Test - admin-cli/pegic`, `Module Labeler` (no `branches` filter)
- ❌ Did NOT trigger: `Cpp CI`, `Lint and Test - admin-cli` (cli proper), `Lint and Test - go-client`, etc.

Confirmed historically:
```
$ gh run list --repo apache/incubator-pegasus \
    --workflow lint_and_test_cpp.yaml --branch v2.5 --limit 5
(empty — never run)
```

`Cpp CI` has **never run on the v2.5 branch** since the workflow was added.

## Fix

Replace `'v[0-9]+.*'` with `'v[0-9]*'`, which is valid GitHub Actions glob syntax and matches `v2.5`, `v2.5.1`, `v10.20`, etc.

Applied uniformly to all 15 affected workflows (mechanical sed replacement, one-line change per file).

## Verification

- [x] All 18 workflow YAMLs parse successfully (`yaml.safe_load`)
- [x] Diff is minimal: 15 files changed, +15/-15
- [ ] After merge: backport to active release branches (`v2.5`, etc.) so existing release-branch PRs start running full CI

## Notes

This fix on `master` does not change behavior for `master`-targeted PRs (they already match the first `- master` line). The benefit materializes once the change is backported to release branches, after which release-branch PRs will trigger the full CI matrix as originally intended.

### DIFF
--- a/.github/workflows/build-push-env-docker.yml
+++ b/.github/workflows/build-push-env-docker.yml
@@ -21,7 +21,7 @@ on:
   push:
     branches:
       - master
-      - 'v[0-9]+.*' # release branch
+      - 'v[0-9]*' # release branch
       - ci-test # testing branch for github action
       - '*dev' # developing branch
     paths:
--- a/.github/workflows/lint_and_test_admin-cli.yml
+++ b/.github/workflows/lint_and_test_admin-cli.yml
@@ -25,7 +25,7 @@ on:
     types: [ synchronize, opened, reopened ]
     branches:
       - master
-      - 'v[0-9]+.*' # release branch
+      - 'v[0-9]*' # release branch
       - ci-test # testing branch for github action
       - '*dev'
     paths:
--- a/.github/workflows/lint_and_test_admin-cli_always_pass.yml
+++ b/.github/workflows/lint_and_test_admin-cli_always_pass.yml
@@ -25,7 +25,7 @@ on:
     types: [ synchronize, opened, reopened ]
     branches:
       - master
-      - 'v[0-9]+.*' # release branch
+      - 'v[0-9]*' # release branch
       - ci-test # testing branch for github action
       - '*dev'
     paths-ignore:
--- a/.github/workflows/lint_and_test_collector.yml
+++ b/.github/workflows/lint_and_test_collector.yml
@@ -25,7 +25,7 @@ on:
     types: [ synchronize, reopened, opened ]
     branches:
       - master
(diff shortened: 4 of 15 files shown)


acelyc111:
## Closing — original diagnosis was wrong

After deeper investigation, the premise of this PR is incorrect:

**1. `v[0-9]+.*` actually does match `v2.5`.** GitHub Actions branch filters tolerate this glob shape and successfully match release branches like `v2.5`. PR #2394's `Cpp CI` workflow **was triggered** (run https://github.com/apache/incubator-pegasus/actions/runs/24079456724), it just doesn't show up in the PR Checks UI because of (2) below.

**2. The reason `Cpp CI` doesn't appear in PR #2394's Checks UI:** the workflow run ended in `conclusion: startup_failure` with an empty `jobs` array. When a run fails before any job is created, GitHub doesn't emit any `check_run`, so the PR's Checks UI has nothing to display.

**3. The actual root cause of `startup_failure`:** the **ASF GitHub Actions allow-list policy** (effective 2025-08-01, see [INFRA-27084](https://issues.apache.org/jira/browse/INFRA-27084)). On v2.5, workflows still reference unpinned third-party actions like `dorny/paths-filter@v2`, `actions/checkout@v3`, etc., which are blocked by the org-level policy. The error message reads:

> `The action dorny/paths-filter@v2 is not allowed in apache/incubator-pegasus because all actions must be ... or match one of the patterns: ...`

This is exactly what #2398 fixed on master (pinning every action to a SHA from the ASF allow-list, plus restructuring some workflows to use local composite actions).

## Correct fix

I'll open a new PR **targeting v2.5** that backports #2398-style SHA pinning of all third-party actions. Sorry for the noise.

